PR #42, opened 2021-11-24
Adding Event Listener for uncaught Promise rejections and updating webvitals tests to skip if browser is Safari or Firefox

direddyz:
1. The JS error plugin only supports synchronous errors it should also support asynchronous errors such as promise rejections.

This change causes the JS error plugin to caught unhandled rejections and dispatch them.

2.Firefox and Safari currently fail several web vitals browser integration tests. This is because Firefox and Safari do not natively support web vitals.To make CI workflow run without having this error we need to skip these tests if the browser is Safari or Firefox.

This change will make WebVitalsPlugin.test.ts to skip if browser is safari or firefox.

Note:changes in CI workflow is made in dinesh-rum-web branch as creating a pull request with a change in github actions file is stopping a pull request.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,9 +2,9 @@ name: AWS RUM Web Client Continuous Build
 
 on:
     push:
-        branches: [dinesh-add-workflow]
+        branches: [main]
     pull_request:
-        branches: [dinesh-add-workflow]
+        branches: [main]
 
 jobs:
     unit-test:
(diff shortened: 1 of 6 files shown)


qhanam:
Thanks @direddyz -- could you split this into two PRs? One for the rejected promise change and one for the firefox/safari integ tests.

In general it is good practice to submit unrelated changes as separate PRs. This makes it easier to review; there are fewer lines of diff to read, the discussion can focus on a single change, and one change won't block the other change from being merged.

PR #48, opened 2021-11-29
Fix broken links to markdown docs inside repo

ranman:


qhanam:
Ah! Sorry we noticed this as well and merged a patch -- https://github.com/aws-observability/aws-rum-web/pull/49

Thanks for seeing this so quickly and getting the patch to us though! Please do let us know if you spot anything else.

PR #54, opened 2021-11-30
Add webvitals tests to skip if browser is Safari or Firefox

direddyz:
Firefox and Safari currently fail several web vitals browser integration tests. This is because Firefox and Safari do not natively support web vitals.To make CI workflow run without having this error we need to skip these tests if the browser is Safari or Firefox.

1.This change will make WebVitalsPlugin.test.ts to skip if browser is safari or firefox.
2.This change will make CI workflow to run integ tests in firefox and edge

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,6 +71,9 @@ jobs:
             - name: Run Integ Tests
               run: npm run integ:local:${{ matrix.browser }}:headless
 
-            - name: Run IE Integ Tests
+            - name: Run IE,Edge,Firefox Integ Tests
               if: matrix.os == 'windows-latest'
-              run: npm run integ:local:ie
+              run: |
+                  npm run integ:local:ie
+                  npm run integ:local:edge
+                  npm run integ:local:firefox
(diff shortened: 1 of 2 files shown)


src/plugins/event-plugins/__integ__/WebVitalsPlugin.test.ts
```diff
@@ -37,32 +37,35 @@ const removeUnwantedEvents = (json: any) => {
 test('WebVitalEvent records lcp and cls events', async (t: TestController) => {
     // If we click too soon, the client/event collector plugin will not be loaded and will not record the click.
     // This could be a symptom of an issue with RUM web client load speed, or prioritization of script execution.
-    await t.wait(300);
+    let browser = t.browser.name;
+    if (browser != 'Safari' && browser != 'Firefox') {
```
qhanam, line 41: nit: consider adding a return statement here to avoid nesting the test code below

.github/workflows/ci.yaml
```diff
@@ -71,6 +71,9 @@ jobs:
             - name: Run Integ Tests
               run: npm run integ:local:${{ matrix.browser }}:headless
```
qhanam, line 72: We can get rid of `browser` from the test matrix, since we're not using it.

Running the integ tests on each browser sequentially within a job is a bit more efficient anyway.

.github/workflows/ci.yaml
```diff
@@ -71,6 +71,9 @@ jobs:
...
 
-            - name: Run IE Integ Tests
+            - name: Run IE,Edge,Firefox Integ Tests
               if: matrix.os == 'windows-latest'
-              run: npm run integ:local:ie
+              run: |
+                  npm run integ:local:ie
+                  npm run integ:local:edge
+                  npm run integ:local:firefox
```
qhanam, line 79: Firefox can go under the `Run Integ Tests` step. The [Ubuntu image](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md) has Firefox installed.

Edge also runs on Linux, but it's not installed by default, so I'm ok leaving it as Windows-only.

PR #57, opened 2021-11-30
Add webvitals tests to skip if browser is Safari or Firefox

direddyz:
Firefox and Safari currently fail several web vitals browser integration tests. This is because Firefox and Safari do not natively support web vitals.To make CI workflow run without having this error we need to skip these tests if the browser is Safari or Firefox.

1.This change will make WebVitalsPlugin.test.ts to skip if browser is safari or firefox.
2.This change will make CI workflow to run integ tests in firefox and edge.

Manual Testing is done to check whether webvital test is skipped if the browser is safari or firefox.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,6 +71,9 @@ jobs:
             - name: Run Integ Tests
               run: npm run integ:local:${{ matrix.browser }}:headless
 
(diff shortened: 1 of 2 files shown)


qhanam:
Duplicate of: https://github.com/aws-observability/aws-rum-web/pull/54

qhanam (approved):

PR #62, opened 2021-12-07
Read cdn release secrets from environment.

qhanam:
The release workflow is currently not protected by required reviewers or deployment branches.

This change configures the deployment workflow to use the [cdn-prod-release](https://github.com/aws-observability/aws-rum-web/settings/environments) environment. This environment is protected with required reviewers and is restricted to the main branch.

### Notes

- Bumps Node.js to v16

### Testing
- I ran a [successful release](https://github.com/qhanam/aws-rum-web/actions/runs/1547181121) from my fork to the [gamma stack](https://client.rum-gamma.us-west-2.amazonaws.com/1.0.3-test/cwr.js).

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -11,6 +11,7 @@ jobs:
     publish_to_cdn:
         name: Publish New Release
         runs-on: ubuntu-latest
+        environment: cdn-prod-release
         permissions: write-all
         steps:
             - name: Checkout AWS RUM Web Client Repository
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yaml
```diff
@@ -19,11 +20,11 @@ jobs:
             - name: Setup Node
               uses: actions/setup-node@v1
               with:
-                  node-version: 14
+                  node-version: 16
 
             - name: Fetch AWS Credentials
               run: |
-                  export AWS_ROLE_ARN=${{ secrets.PROD_US_EAST_1_ROLE }}
```
anuraaga, line 26: configure-aws-credentials got native support for this flow so will be good to migrate to it at some point

anuraaga (approved):

.github/workflows/cd.yaml
qhanam: I tried configure-aws-credentials when I first set this up, but I ran into issues (i.e., the deployment was failing and difficult to debug -- see https://github.com/aws-observability/aws-rum-web/pull/4).

Maybe its worth revisiting now that it's a few months more mature, but this `curl` method works fine so I'm not sure I want to put effort into it.

PR #67, opened 2021-12-09
Add npm prepublish script.

qhanam:
The `npm publish` command  does not currently run a build before publishing the package to npm. This is problematic because it may result in an incorrect version being published to NPM.

This change adds the `release` script as a `prepublishOnly` action, so that the package is re-built before being published.

This change also replaces the `.npmignore` (exclude) list with a `files` (include) list. Using an include list [is safer than using an exclude list](https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d), so that confidential files are inadvertently published.


package.json
```diff
@@ -19,6 +25,7 @@
         "clean": "rm -rf ./build && rm -rf ./node_modules & rm package-lock.json",
         "server": "webpack-dev-server --config webpack/webpack.dev.js",
         "release": "npm run test && npm run build",
```
anuraaga, line 27: Looks like these scripts could use a cleanup scrub. One awkwardness I see is that this builds both a dev and release bundle, even though I'd expect `release` to only build the latter. I'm not sure I've ever seen `build:dev` before as dev is usually starting up a dev server, not building anything.

anuraaga (approved):

package.json
qhanam: Agreed -- scripts are a little janky. I've refactored as follows:
- `server`: runs webpack dev server
- `test`: builds webpack dev bundle and runs unit tests
- `release`: builds webpack prod bundle and builds npm package files

anuraaga (approved):

PR #75, opened 2021-12-23
Fix: Uncaught TypeError: Cannot read the properties of undefined (reading 'record')

madhuhc:
Adding a check when `getSession()` returns undefined to avoid JS error and breaking the page.

---

### DIFF
--- a/src/plugins/event-plugins/FetchPlugin.ts
+++ b/src/plugins/event-plugins/FetchPlugin.ts
@@ -85,7 +85,7 @@ export class FetchPlugin extends MonkeyPatched implements Plugin {
     };
 
     private isSessionRecorded = () => {
-        return this.context.getSession().record;
+        return this.context.getSession()?.record || false;
     };
 
     private beginTrace = (
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "aws-rum-web",
-    "version": "1.0.3",
+    "version": "1.0.4",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "aws-rum-web",
-            "version": "1.0.3",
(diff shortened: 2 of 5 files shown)


src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -85,7 +85,7 @@ export class FetchPlugin extends MonkeyPatched implements Plugin {
     };
 
     private isSessionRecorded = () => {
-        return this.context.getSession().record;
+        return this.context.getSession()?.record || false;
```
qhanam, line 88: Please add the following test to `FetchPlugin.test.ts` to verify the new behavior and prevent regressions:

```
    test('when getSession returns undefined then the plugin does not record a trace', async () => {
        const getSession: jest.MockedFunction<GetSession> = jest.fn(
            () => undefined
        );
        const config: PartialHttpPluginConfig = {
            logicalServiceName: 'sample.rum.aws.amazon.com',
            urlsToInclude: [/aws\.amazon\.com/]
        };
        const xRayOnContext: PluginContext = {
            applicationId: 'b',
            applicationVersion: '1.0',
            config: { ...DEFAULT_CONFIG, ...{ enableXRay: true } },
            record,
            recordPageView,
            getSession
        };

        const plugin: FetchPlugin = new FetchPlugin(config);
        plugin.load(xRayOnContext);

        // Run
        await fetch('https://aws.amazon.com');
        plugin.disable();

        // Assert
        expect(mockFetch).toHaveBeenCalledTimes(1);
        expect(record).not.toHaveBeenCalled();
    });
```

qhanam (changes requested):
Thanks for raising this!

Looks good to me, but please add the two unit tests I've proposed to verify the new behavior and prevent regressions.

qhanam (approved):

PR #76, opened 2022-01-11
fix: add Cache-Control max-age to CDN files

### LINKED ISSUE #73: Cache cwr.js file from CDN
I'm currently trying out the CloudWatch RUM product. One thing I noticed today is that the https://client.rum.us-east-1.amazonaws.com/1.0.2/cwr.js file is served without any Cache-Control headers:

```
 curl -v https://client.rum.us-east-1.amazonaws.com/1.0.2/cwr.js
*   Trying 65.8.233.80:443...
* Connected to client.rum.us-east-1.amazonaws.com (65.8.233.80) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=client.rum.us-east-1.amazonaws.com
*  start date: Nov 12 00:00:00 2021 GMT
*  expire date: Dec 11 23:59:59 2022 GMT
*  subjectAltName: host "client.rum.us-east-1.amazonaws.com" matched cert's "client.rum.us-east-1.amazonaws.com"
*  issuer: C=US; O=Amazon; OU=Server CA 1B; CN=Amazon
*  SSL certificate verify ok.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x7f98b6811000)
> GET /1.0.2/cwr.js HTTP/2
> Host: client.rum.us-east-1.amazonaws.com
> user-agent: curl/7.77.0
> accept: */*
>
* Connection state changed (MAX_CONCURRENT_STREAMS == 128)!
< HTTP/2 200
< content-type: application/javascript
< content-length: 114311
< last-modified: Tue, 23 Nov 2021 00:51:47 GMT
< x-amz-version-id: mQ3VEzOswAhdeQEPxFrH_CGxedkxk4zn
< accept-ranges: bytes
< server: AmazonS3
< date: Sat, 18 Dec 2021 12:39:43 GMT
< etag: "0dfa77fbb42f338ac9955897e9f55641"
< vary: Accept-Encoding
< x-cache: Hit from cloudfront
< via: 1.1 30ea845097208edbc19305c535a5be99.cloudfront.net (CloudFront)
< x-amz-cf-pop: DEN52-C1
< x-amz-cf-id: Vo5ZG0zegHb7QUxU2Iu4PwiIXRBZv3_ibZd5Ack_ZENrsWNgB2kHJQ==
< age: 38372
```

This (rightfully) triggers warnings from Lighthouse:

Screen Shot 2021-12-18 at 4 20 24 PM

As you can see from the screenshot, Google Analytics caches for 2 hours but is still flagged. Since the URL for the CloudWatch asset has a version embedded (1.0.2), it seems like this should be safe to cache for at least 24 hours, if not more.

qhanam:
Partially resolves #73 .

This change adds a Cache-Control header to new web client versions deployed to CDN. Specific versions (e.g., `1.0.4`) have a max age of **one week**. Auto-update versions (e.g., `1.0.x` and `1.x`) have a max age of **two hours**, so that impact is limited in case a change needs be be rolled back.

Note that browsers typically cache files by default (using a heuristic to estimate TTL), so this change is non-critical. 

We still need to add Cache-Control headers to the existing CDN files to fully resolve #73. This will be done by directly modifying the header info of the existing files.

### Testing

I ran the deployment workflow in the gamma environment and verified the header `cache-control: max-age=604800` is present in `1.0.4/cwr.js`, and `cache-control: max-age=7200` is present in `1.0.x/cwr.js` and `1.x/cwr.js`. The Lighthouse recommendation no longer appeared for `1.0.4/cwr.js`.

---

### DIFF
--- a/.github/scripts/deploy.sh
+++ b/.github/scripts/deploy.sh
@@ -3,15 +3,15 @@
 version=$(npm pkg get version | sed 's/"//g')
 bucket=$1
 
-aws s3api put-object --bucket $bucket --key "content/$version/cwr.js" --body build/assets/cwr.js
-aws s3api put-object --bucket $bucket --key "content/$version/LICENSE-THIRD-PARTY" --body LICENSE-THIRD-PARTY
+aws s3api put-object --bucket $bucket --key "content/$version/cwr.js" --body build/assets/cwr.js --cache-control max-age=604800
+aws s3api put-object --bucket $bucket --key "content/$version/LICENSE-THIRD-PARTY" --body LICENSE-THIRD-PARTY --cache-control max-age=604800
 
 if [[ $version =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
 then
     minorUpdate=$(echo $version | sed -En "s/^([0-9]+\.)[0-9]+\.[0-9]+/\1x/p")
-    aws s3api put-object --bucket $bucket --key "content/$minorUpdate/cwr.js" --body build/assets/cwr.js
-    aws s3api put-object --bucket $bucket --key "content/$minorUpdate/LICENSE-THIRD-PARTY" --body LICENSE-THIRD-PARTY
+    aws s3api put-object --bucket $bucket --key "content/$minorUpdate/cwr.js" --body build/assets/cwr.js --cache-control max-age=7200
+    aws s3api put-object --bucket $bucket --key "content/$minorUpdate/LICENSE-THIRD-PARTY" --body LICENSE-THIRD-PARTY --cache-control max-age=7200
     patchUpdate=$(echo $version | sed -En "s/^([0-9]+\.[0-9]+\.)[0-9]+/\1x/p")
-    aws s3api put-object --bucket $bucket --key "content/$patchUpdate/cwr.js" --body build/assets/cwr.js
-    aws s3api put-object --bucket $bucket --key "content/$patchUpdate/LICENSE-THIRD-PARTY" --body LICENSE-THIRD-PARTY
+    aws s3api put-object --bucket $bucket --key "content/$patchUpdate/cwr.js" --body build/assets/cwr.js --cache-control max-age=7200
+    aws s3api put-object --bucket $bucket --key "content/$patchUpdate/LICENSE-THIRD-PARTY" --body LICENSE-THIRD-PARTY --cache-control max-age=7200
 fi


.github/scripts/deploy.sh
```diff
@@ -3,15 +3,15 @@
 version=$(npm pkg get version | sed 's/"//g')
 bucket=$1
 
-aws s3api put-object --bucket $bucket --key "content/$version/cwr.js" --body build/assets/cwr.js
-aws s3api put-object --bucket $bucket --key "content/$version/LICENSE-THIRD-PARTY" --body LICENSE-THIRD-PARTY
+aws s3api put-object --bucket $bucket --key "content/$version/cwr.js" --body build/assets/cwr.js --cache-control max-age=604800
```
anuraaga, line 6: Does it make sense to set these even longer, like 1 year?

anuraaga (approved):

.github/scripts/deploy.sh
qhanam: I'm not sure. It's about risk -- what is the likelihood that a specific version will need to be modified?

In theory a version should never change once deployed, but my preference is still to reduce risk, at least for now, by limiting a potential impact to one week instead of one year.

PR #81, opened 2022-02-03
fix: capture PerformanceNavigationTiming events even when window.load fires before plugin loads

adebayor123:
Modified NavigationPlugin to capture PerformanceNavigationEvents when the plugin loads after the page does.
- Removed PerformanceObserver from eventListener.
- Added two new unit test cases and modified existing one case with respect to the new implementation change.
- Created a new app called delayed_page.html that loads the script after the page finishes loading.
- Created a new integ test case using the delayed_page app to verify that PerformanceNavigationEvents are captured.

To test the change, do the following:
- Create a temporary app (delayed_page.html) in your local repository without any other implementation changes
- Run server and either use Chrome debugger or open up local host, compare the output between index.html and delayed_page.html.
- index.html should have the navigation events, while the delayed_page.html does not.
- This indicates that with the original implementation, delayed_page properly simulates a web page loading prior to plugin
- Once this is verified, pull in the NavigationPlugin changes, and rerun the server/debugger to see that now PerformanceNavigationEvents are captured in the delayed_page app.
---

### DIFF
--- a/src/plugins/event-plugins/NavigationPlugin.ts
+++ b/src/plugins/event-plugins/NavigationPlugin.ts
@@ -22,10 +22,20 @@ export class NavigationPlugin implements Plugin {
         this.enabled = true;
     }
 
+    /**
+     * If the client is initialized after the window has fired a load event,
+     * invoke the callback method directly to trigger the record event.
+     * Otherwise, keep the original implementation to add callback method to eventListener.
+     * However, if the page finishes loading right before adding addEventListener, we still cannot provide data
+     */
     load(context: PluginContext): void {
         this.recordEvent = context.record;
         if (this.enabled) {
-            window.addEventListener(LOAD, this.eventListener);
+            if (this.hasTheWindowLoadEventFired()) {
+                this.eventListener();
+            } else {
+                window.addEventListener(LOAD, this.eventListener);
+            }
         }
     }
 
@@ -47,6 +57,24 @@ export class NavigationPlugin implements Plugin {
         }
     }
 
+    /**
+     * Use the loadEventEnd field from window.performance to check if the website
+     * has loaded already.
+     * @returns boolean
+     */
+    hasTheWindowLoadEventFired() {
+        if (
+            window.performance &&
+            window.performance.getEntriesByType(NAVIGATION).length
+        ) {
+            const navData = window.performance.getEntriesByType(
+                NAVIGATION
+            )[0] as PerformanceNavigationTiming;
+            return Boolean(navData.loadEventEnd);
+        }
+        return false;
+    }
+
     getPluginId(): string {
         return this.pluginId;
     }
@@ -57,23 +85,18 @@ export class NavigationPlugin implements Plugin {
      *
      * If browser provides support, use Navigation Timing Level 2 specification -
      * https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming
+     *
+     * Only the current document resource is included in the performance timeline;
+     * there is only one PerformanceNavigationTiming object in the performance timeline.
+     * https://www.w3.org/TR/navigation-timing-2/
      */
     eventListener = () => {
         if (performance.getEntriesByType(NAVIGATION).length === 0) {
             this.performanceNavigationEventHandlerTimingLevel1();
         } else {
-            const navigationObserver = new PerformanceObserver((list) => {
-                list.getEntries().forEach((event) => {
-                    if (event.entryType === NAVIGATION) {
-                        this.performanceNavigationEventHandlerTimingLevel2(
-                            event
-                        );
-                    }
-                });
-            });
-            navigationObserver.observe({
-                entryTypes: [NAVIGATION]
-            });
+            this.performanceNavigationEventHandlerTimingLevel2(
+                performance.getEntriesByType(NAVIGATION)[0]
+            );
         }
     };
 
--- a/src/plugins/event-plugins/__integ__/NavigationPlugin.test.ts
+++ b/src/plugins/event-plugins/__integ__/NavigationPlugin.test.ts
@@ -1,0 +1,150 @@
+import {
+    STATUS_202,
+    DISPATCH_COMMAND,
+    COMMAND,
+    PAYLOAD,
+    SUBMIT,
+    REQUEST_BODY,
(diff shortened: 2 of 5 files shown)


src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -47,6 +57,24 @@ export class NavigationPlugin implements Plugin {
...
+    /**
+     * Use the loadEventEnd field from window.performance to check if the website
+     * has loaded already.
+     * @returns boolean
+     */
+    hasTheWindowLoadEventFired() {
+        if (
+            window.performance &&
+            window.performance.getEntriesByType(NAVIGATION).length
+        ) {
```
qhanam, line 69: Does this guarantee that `window.load` has fired? 

Or should we look for a specific event that signals the page has finished loading -- i.e., [`PerformanceTiming.loadEventEnd`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming/loadEventEnd) or [`PerformanceNavigationTiming.loadEventEnd`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/loadEventEnd)

src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
```diff
@@ -92,13 +93,43 @@ describe('NavigationPlugin tests', () => {
...
 
         // Assert
         expect(record).toHaveBeenCalledTimes(0);
     });
+
+    test('when page has not loaded then we do not fire until load is completed', async () => {
+        // Setting up new mocked window that has not loaded
+        (window as any).performance = performanceEventNotLoaded.performance();
+        (window as any).PerformanceObserver =
+            performanceEvent.PerformanceObserver;
```
qhanam, line 112: Do we need to mock `PerformanceObserver` here? It was removed from `NavigationPlugin`

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -62,18 +90,9 @@ export class NavigationPlugin implements Plugin {
...
-                            event
-                        );
-                    }
-                });
-            });
-            navigationObserver.observe({
-                entryTypes: [NAVIGATION]
-            });
+            this.performanceNavigationEventHandlerTimingLevel2(
+                performance.getEntriesByType(NAVIGATION)[0]
```
qhanam, line 94: Is it guaranteed that there is only one navigation timing entry?

src/plugins/event-plugins/__integ__/NavigationPlugin.test.ts
```diff
@@ -0,0 +1,152 @@
...
+const COMPRESSION_RATIO = 'compressionRatio';
+const FILE_TYPE = 'fileType';
+const SAFARI = 'Safari';
+const FIREFOX = 'Firefox';
+
+fixture('NagivationEvent Plugin').page(
+    'http://localhost:8080/delayed_page.html'
+);
+
+test('Delayed plugin loading still records load events', async (t: TestController) => {
```
qhanam, line 55: nit: "when plugin loads after window.load then navigation timing events are recorded"

src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
```diff
@@ -92,13 +93,43 @@ describe('NavigationPlugin tests', () => {
...
+        plugin.load(context);
+        // assert that the plugin did not fire
+        expect(record).toHaveBeenCalledTimes(0);
+
+        // now that the page has loaded, we should fire
+        window.dispatchEvent(new Event('load'));
+        expect(record).toHaveBeenCalledTimes(1);
+    });
+
+    test('when page loads before then we still fire when loading plugin', async () => {
```
qhanam, line 125: nit: 'when window.load fires before plugin loads then navigation timing is recorded'

src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
```diff
@@ -92,13 +93,43 @@ describe('NavigationPlugin tests', () => {
...
+        plugin.load(context);
+
         window.dispatchEvent(new Event('load'));
         plugin.disable();
 
         // Assert
         expect(record).toHaveBeenCalledTimes(0);
     });
+
+    test('when page has not loaded then we do not fire until load is completed', async () => {
```
qhanam, line 108: nit: 'when window.load fires after plugin loads then navigation timing is recorded'

qhanam:
Please modify the commit message to use the following conventions:
1. Use [conventional commit format](https://www.conventionalcommits.org/)
    i.e.,\<type\>[optional scope]: \<description\>
2. Use [the imperative mood in the subject line](https://cbea.ms/git-commit/#imperative)
i.e., a commit message subject line should be able to complete the sentence _"If applied, this commit will __commit message subject line__"_

src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
adebayor123: Agreed, no need. Will reflect on next revision.

src/plugins/event-plugins/NavigationPlugin.ts
adebayor123: Good catch, I will change it to iterate instead.

src/plugins/event-plugins/NavigationPlugin.ts
adebayor123: Discussed offline; The following line to the if statement uses PerformanceNavigationTiming.loadEventEnd

src/plugins/event-plugins/NavigationPlugin.ts
adebayor123: Actually, just found this: https://www.w3.org/TR/navigation-timing-2/ 
"Only the [current document](https://www.w3.org/TR/navigation-timing-2/#dfn-current-document) resource is included in the performance timeline; there is only one [PerformanceNavigationTiming](https://www.w3.org/TR/navigation-timing-2/#dom-performancenavigationtiming) object in the performance timeline."

qhanam (approved):

PR #82, opened 2022-02-04
feat: support unique session cookie names

qhanam:
Multiple AppMonitors cannot currently be used on the same page. Because the session cookie name `cwr_s` is not unique to an AppMonitor, multiple web clients running on the same page would attempt to read and store session info to the same cookie, which would produce incorrect data.

This change adds a configuration option to add the AppMonitor ID as a postfix to the session cookie `cwr_s`. When used, the session cookie name will have the format `cwr_s_[AppMonitor Id]`, and multiple web clients can run on the same page.

---


docs/cdn_installation.md
```diff
@@ -90,6 +90,7 @@ The snippet below shows several configuration options with the body of the snipp
 | path | String | `/` | See https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_where_cookies_are_sent |
 | sameSite | Boolean | `true` | See https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_where_cookies_are_sent |
 | secure | Boolean | `true` | See https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_where_cookies_are_sent |
+| unique | Boolean | `false` | When this field is `false`, the session cookie name is `cwr_s`. When this field is true `true`, the session cookie name is `cwr_s_[AppMonitor Id]`.<br/><br/>Set this field to `true` when multiple AppMonitors will monitor the same page. For example, this might be the case if one AppMonitor is used for logged-in users, and a second AppMonitor is used for guest users.  |
```
anuraaga, line 93: Is there any downside to always using the suffixed behavior?

anuraaga (approved):

adebayor123 (approved):

docs/cdn_installation.md
qhanam: My reasoning for preferring `cwr_s` from a technical perspective is that the name takes up fewer bytes. My reasoning for preferring `cwr_s` from a non-technical perspective is that (1) it promotes setting the cookie at the appropriate domain and path, and (2) it makes it easier to document the names of the cookies for consent/transparency purposes.

PR #87, opened 2022-02-11
feat: identify DOM events using CSS locator

limhjgrace:
Users currently configure the DOM event plugin to listen to events either with (1) the ID of the DOM element or (2) the actual JavaScript object. However, a customer may want to add a tag to certain elements which should be recorded. There is currently no way to listen for a certain tag.

This feature allows customers to specify listeners for DOM elements using CSS locators.

---

### DIFF
--- a/src/event-schemas/dom-event.json
+++ b/src/event-schemas/dom-event.json
@@ -16,8 +16,12 @@
         "elementId": {
             "type": "string",
             "description": "DOM element ID."
+        },
+        "cssLocator": {
+            "type": "string",
+            "description": "CSS Locator string."
         }
     },
     "additionalProperties": false,
-    "required": ["version", "event", "elementId"]
+    "required": ["version", "event", "elementId", "cssLocator"]
 }
--- a/src/event-schemas/dom-event.json
+++ b/src/event-schemas/dom-event.json
@@ -23,5 +23,5 @@
         }
     },
     "additionalProperties": false,
-    "required": ["version", "event", "elementId", "cssLocator"]
+    "required": ["version", "event", "elementId"]
 }
--- a/src/plugins/event-plugins/DomEventPlugin.ts
+++ b/src/plugins/event-plugins/DomEventPlugin.ts
@@ -16,6 +16,11 @@ export type TargetDomEvent = {
     elementId?: string;
 
     /**
+     * DOM element map to identify one element attribute and its expected value
+     */
+    cssLocator?: string;
+
+    /**
      * DOM element
      */
     element?: HTMLElement;
@@ -84,12 +89,13 @@ export class DomEventPlugin implements Plugin {
         );
     }
 
-    private getEventListener(): EventListener {
+    private getEventListener(domEvent?: TargetDomEvent): EventListener {
         return (event: Event): void => {
             const eventData: DomEvent = {
                 version: '1.0.0',
                 event: event.type,
-                elementId: this.getElementId(event)
+                elementId: this.getElementId(event),
+                cssLocator: this.getElementCSSLocator(event, domEvent)
             };
             if (this.recordEvent) {
                 this.recordEvent(DOM_EVENT_TYPE, eventData);
@@ -113,15 +119,33 @@ export class DomEventPlugin implements Plugin {
         return '';
     }
 
+    private getElementCSSLocator(event: Event, domEvent: TargetDomEvent) {
+        if (!event.target) {
+            return '';
+        }
+
+        if (event.target instanceof Element && domEvent.cssLocator) {
+            return domEvent.cssLocator;
+        }
+
+        if (event.target instanceof Node) {
+            return event.target.nodeName;
+        }
+    }
     private addEventHandler(domEvent: TargetDomEvent): void {
         const eventType = domEvent.event;
-        const eventListener = this.getEventListener();
+        const eventListener = this.getEventListener(domEvent);
         this.eventListenerMap.set(domEvent, eventListener);
 
         if (domEvent.elementId) {
             document
                 .getElementById(domEvent.elementId)
                 ?.addEventListener(eventType, eventListener);
+        } else if (domEvent.cssLocator) {
+            const elementList = document.querySelectorAll(domEvent.cssLocator);
+            for (let i = 0; i < elementList.length; i++) {
+                elementList[i].addEventListener(eventType, eventListener);
+            }
         } else if (domEvent.element) {
             domEvent.element.addEventListener(eventType, eventListener);
         }
--- a/src/plugins/event-plugins/DomEventPlugin.ts
+++ b/src/plugins/event-plugins/DomEventPlugin.ts
@@ -41,13 +41,18 @@ const defaultConfig: DomEventPluginConfig = {
 export class DomEventPlugin implements Plugin {
     private recordEvent: RecordEvent | undefined;
     private pluginId: string;
-    private eventListenerMap: Map<TargetDomEvent, EventListener>;
+    private eventListenerMap: Map<TargetDomEvent, Map<string, EventListener>>;
+    private identifierToEventListenerMap: Map<string, EventListener>;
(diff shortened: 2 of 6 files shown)


src/plugins/event-plugins/DomEventPlugin.ts
```diff
@@ -84,12 +89,13 @@ export class DomEventPlugin implements Plugin {
...
 
-    private getEventListener(): EventListener {
+    private getEventListener(domEvent?: TargetDomEvent): EventListener {
         return (event: Event): void => {
             const eventData: DomEvent = {
                 version: '1.0.0',
                 event: event.type,
-                elementId: this.getElementId(event)
+                elementId: this.getElementId(event),
+                cssLocator: this.getElementCSSLocator(event, domEvent)
```
qhanam, line 98: Since the schema specifies that `cssLocator` is type `string`, we omit the `cssLocator` field to the event object if `cssLocator` is undefined.

src/event-schemas/dom-event.json
```diff
@@ -16,8 +16,12 @@
...
             "description": "DOM element ID."
+        },
+        "cssLocator": {
+            "type": "string",
+            "description": "CSS Locator string."
         }
     },
     "additionalProperties": false,
-    "required": ["version", "event", "elementId"]
+    "required": ["version", "event", "elementId", "cssLocator"]
```
qhanam, line 26: We probably want `cssLocator` field to be an optional field. We always provide `elementId`, but we do not always provide `cssLocator`.

src/plugins/event-plugins/DomEventPlugin.ts
```diff
@@ -113,15 +119,33 @@ export class DomEventPlugin implements Plugin {
...
+        }
+
+        if (event.target instanceof Element && domEvent.cssLocator) {
+            return domEvent.cssLocator;
+        }
+
+        if (event.target instanceof Node) {
+            return event.target.nodeName;
+        }
+    }
```
qhanam, line 134: Unlike the `elementId`, we can consider the CSS locator optional. In this case we can return a CSS locator if one exists, or `undefined` otherwise.
```javascript
private getElementCSSLocator(domEvent: TargetDomEvent) {
    if (domEvent.cssLocator) {
        return domEvent.cssLocator;
    }
    return undefined;
}
```

src/plugins/event-plugins/DomEventPlugin.ts
```diff
@@ -113,15 +119,33 @@ export class DomEventPlugin implements Plugin {
...
         const eventType = domEvent.event;
-        const eventListener = this.getEventListener();
+        const eventListener = this.getEventListener(domEvent);
         this.eventListenerMap.set(domEvent, eventListener);
 
         if (domEvent.elementId) {
             document
                 .getElementById(domEvent.elementId)
                 ?.addEventListener(eventType, eventListener);
+        } else if (domEvent.cssLocator) {
```
qhanam, line 144: If there are multiple locators in `domEvent`, which locator should be used? I.e., right now it is `elementID -> cssLocator -> element`. We could also throw an error, but I think it's better to have a sane default here instead.

I think we should add a unit test that defines which locator is used in this case.

src/event-schemas/dom-event.json
limhjgrace: Removed `cssLocator` from being a required property.

src/plugins/event-plugins/DomEventPlugin.ts
limhjgrace: Refactored so that `cssLocator` field is only added when defined.

src/plugins/event-plugins/DomEventPlugin.ts
limhjgrace: Refactored as suggested.

src/plugins/event-plugins/DomEventPlugin.ts
limhjgrace: As discussed, we'll add listeners for all fields that are passed in.

src/plugins/event-plugins/DomEventPlugin.ts
```diff
@@ -36,13 +41,18 @@ const defaultConfig: DomEventPluginConfig = {
 export class DomEventPlugin implements Plugin {
     private recordEvent: RecordEvent | undefined;
     private pluginId: string;
-    private eventListenerMap: Map<TargetDomEvent, EventListener>;
+    private eventListenerMap: Map<TargetDomEvent, Map<string, EventListener>>;
```
qhanam, line 44: **suggestion:** The problem we have here is that the TargetDomEvent can potentially contain three selectors: ID, CSS locator or object. Having a map within a map makes things a little hard to follow. Consider a different strategy:

1. Map a `TargetDomEvent` to an array of `EventListener`. There will be either one or two items in the array. This will remove a lot of the complexity, because we don't need to think about the 'type' of event listener.

2. Only use one of the three selectors. This adds the least complexity. We can document that only one of selectors should be given, and the order in which the selectors will be preferred.

src/plugins/event-plugins/DomEventPlugin.ts
```diff
@@ -84,13 +94,16 @@ export class DomEventPlugin implements Plugin {
         );
     }
 
-    private getEventListener(): EventListener {
+    private getEventListener(domEvent?: TargetDomEvent): EventListener {
```
qhanam, line 97: **nit:** We can make this a bit more readable by changing the `domEvent` parameter to `cssLocator`

src/plugins/event-plugins/DomEventPlugin.ts
```diff
@@ -113,32 +126,82 @@ export class DomEventPlugin implements Plugin {
...
+        // add event listener to element identified by the element ID if no CSS locator provided or
+        // the element is not one of the elements identified by the CSS locator
+        if (domEvent.elementId && !isElementIdentifiedByLocator) {
             document
                 .getElementById(domEvent.elementId)
-                ?.addEventListener(eventType, eventListener);
+                ?.addEventListener(eventType, eventListenerID);
+
+            // safe default
         } else if (domEvent.element) {
```
qhanam, line 165: **issue:** If we are adding event listeners to **all** selectors (i.e., ID, object and CSS) in a given `TargetDomEvent`, this should be an `if` statement rather than an `else if` statement.

src/plugins/event-plugins/DomEventPlugin.ts
limhjgrace: Refactored

src/plugins/event-plugins/DomEventPlugin.ts
```diff
@@ -84,13 +94,16 @@ export class DomEventPlugin implements Plugin {
...
-    private getEventListener(): EventListener {
+    private getEventListener(domEvent?: TargetDomEvent): EventListener {
         return (event: Event): void => {
             const eventData: DomEvent = {
                 version: '1.0.0',
                 event: event.type,
                 elementId: this.getElementId(event)
             };
+            if (domEvent?.cssLocator) {
+                eventData.cssLocator = this.getElementCSSLocator(domEvent);
```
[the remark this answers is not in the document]
limhjgrace: Did you mean:
if (cssLocator !== undefined) {
    eventData.cssLocator = cssLocator;

src/plugins/event-plugins/DomEventPlugin.ts
limhjgrace: Refactored

src/plugins/event-plugins/DomEventPlugin.ts
limhjgrace: Will go with option 2. Preference will be: CSS Selector > Element ID > Element. CSS Selector will be preferred if provided because it point to the most elements while element ID points to only one element.

qhanam (approved):

PR #93, opened 2022-02-18
fix: Set x-amzn-trace-id header directly in request headers

qhanam:
When `fetch` is called with a `Request` object as the first argument, the web client overrides the custom http headers in the `Request` object. This is because the web client adds the trace id header to a new `init` argument, which causes existing custom http headers to be overwritten.

This change modifies the behavior of the `fetch` monkey patch such that when `fetch` is called with a `Request` object, the `X-Amzn-Trace-Id` header is added directly to the `Request.header` object, instead of to a new `init` object.

---

### DIFF
--- a/src/plugins/event-plugins/FetchPlugin.ts
+++ b/src/plugins/event-plugins/FetchPlugin.ts
@@ -10,12 +10,13 @@ import {
     defaultConfig,
     epochTime,
     createXRayTraceEvent,
-    addAmznTraceIdHeader,
+    addAmznTraceIdHeaderToInit,
     HttpPluginConfig,
     createXRayTraceEventHttp,
     isUrlAllowed,
     createXRaySubsegment,
-    requestInfoToHostname
+    requestInfoToHostname,
+    addAmznTraceIdHeaderToHeaders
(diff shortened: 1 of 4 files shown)


src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -106,19 +103,37 @@ export class FetchPlugin extends MonkeyPatched implements Plugin {
...
+                (input as Request).headers,
                 xRayTraceEvent.trace_id,
                 xRayTraceEvent.subsegments[0].id
             );
         }
 
-        return xRayTraceEvent;
+        if (!init) {
+            init = {};
+            [].push.call(argsArray, init);
```
anuraaga, line 129: Or maybe `(argsArray as any[]).push(init)`

anuraaga (approved):

src/plugins/event-plugins/FetchPlugin.ts
qhanam: Agree this would be more readable, however TypeScript doesn't like this :-)

PR #104, opened 2022-03-01
feat: Dynamically update DOM event listeners

limhjgrace:
Currently: 
1. The DOM event plugin doesn't dynamically update event handlers when new content is loaded. 
2. The DOM events that the plugin listens for are set at compile time and cannot be added at runtime.

This feature allows customers to:
1. Detect DOM changes and automatically register DOM event listeners when new elements are added.
2. Add DOM events to the RUM web client during runtime. 

---


limhjgrace:
Git pull brought in unnecessary previous commits. Will raise PR from another branch.

PR #108, opened 2022-03-02
feat: Capture SPA virtual page load timing (changes for manual API call to recordPageViews)

adebayor123:
- Changes to distinguish `initial_load` vs `route_change` when invoked manually
- When the customer uses manual API instead of automatic page views, the initial page is not captured at all. As a result, assuming all recordPageView is for `route_change` may be misleading.
- Input command would be `cwr('recordPageView', ['/page_view_two', 'route_change'])`

---

### DIFF
--- a/src/sessions/__integ__/PageManager.test.ts
+++ b/src/sessions/__integ__/PageManager.test.ts
@@ -1,3 +1,4 @@
+import { Assertion } from 'chai';
 import { Selector } from 'testcafe';
 import { REQUEST_BODY } from '../../test-utils/integ-test-utils';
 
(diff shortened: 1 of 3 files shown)


src/sessions/__integ__/PageManager.test.ts
```diff
@@ -21,7 +22,7 @@ test('PageViewEventPlugin records landing page view event', async (t: TestContro
     // If we click too soon, the client/event collector plugin will not be loaded and will not record the click.
     // This could be a symptom of an issue with RUM web client load speed, or prioritization of script execution.
     await t
-        .wait(300)
+        .wait(3000)
```
adebayor123, line 25: This needs to be fixed before merging

PR #113, opened 2022-03-03
feat: Capture SPA virtual page load timing

adebayor123:
**Context**:

Currently, RUM Web Client does not provide a mechanism to calculate the virtual page load timing in single page applications, as the `navigation-event` is only sent when the window's `load` event is fired.

**Changes**:

1. Detect route change as a starting point to measure virtual page load.
2. Create a timing mechanism within VirtualPageLoadTimer using Timers.
3. Intercept outgoing AJAX requests in the form of `Fetch` and `XMLHttpRequest` and use it as a datapoint to calculate the current page's load time.
4. Observe DOM mutations to calculate the current page's load time.

**Test**:
1. Modified existing unit tests based on parameter changes for maintenance
2. No existing unit test's behaviors were changed
3. Added new unit test cases for the new changes
4. Added new integ test cases for VirtualPageLoadTimer
5. Verified behavior on Pre-Prod environment

---


adebayor123:
TODO: Simplify SPA test app
TODO: ongoingActivity/requestCache type
TODO: Documentation update

qhanam:
Please add a description to this PR. The description should contain at least (1) what the problem is, and (2) how this change solves the problem.

src/plugins/Plugin.ts
```diff
@@ -13,6 +18,11 @@ export type PluginContext = {
     record: RecordEvent;
     recordPageView: RecordPageView;
     getSession: GetSession;
+    getCurrentUrl: GetCurrentUrl;
+    getCurrentPage: GetCurrentPage;
+    getRequestCache: GetRequestCache;
+    incrementFetch: IncrementFetch;
+    decrementFetch: DecrementFetch;
```
qhanam, line 25: issue: I think once we move `TrackerPlugin` to the core logic (i.e., in the same folder as `PageManager`), these can be removed.

src/plugins/event-plugins/PageViewPlugin.ts
```diff
@@ -1,17 +1,30 @@
...
  *
  * When a session is initialized, the PageManager records the landing page. When
  * subsequent pages are viewed, this plugin updates the page.
+ *
+ * When AutoPageView is enabled, this plugin does the following:
+ * (1) Records initial page as INITIAL_LOAD during load()
+ * (2) Any subsequent route changes via Browser History API is intercepted to:
+ *      - Create a new Page object with pageInvokeType as ROUTE_CHANGE
+ *
+ * In order for accurate tracking in SPA, recommend using PATH_AND_HASH for pageId.
```
qhanam, line 27: question: What is important to communicate in this user documentation? 

> In order for accurate tracking in SPA, recommend using PATH_AND_HASH for pageId.

Why is PATH_AND_HASH needed for accurate tracking? What happens if it is not used? We probably don't want to recommend a format here, since the correct format depends on the application.

src/plugins/event-plugins/TrackerPlugin.ts
```diff
@@ -0,0 +1,209 @@
...
+ * Page object's ongoingActivity.
+ *
+ * Once the request is completed, it will mark the time of completion to calculate the current
+ * Page object's latestEndTime (if in ongoingActivity and page has not loaded),
+ * or simply removed from requestCache (if page has loaded).
+ *
+ * For Fetch, we use the input url combined with the init? parameter.
+ * Intuition behind this is that requests that have the same URL will most likely be
+ * POST requests, that have the init? parameter populated. As a result, combining init? with
+ * input url will ensure trackerPlugin can distinguish different fetch requests with same URLs.
```
qhanam, line 41: suggestion: Consider the audience when creating 'javadoc' documentation vs code comments. A javadoc (or in this case jsdoc) reader is likely interested in the what the class provides (timing virtual page loads), while a code comment reader is likely interested in implementation details. 

Also see the [Google Java Style Guide: The Summary Fragment](https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment).

src/plugins/event-plugins/PageViewPlugin.ts
```diff
@@ -74,27 +91,41 @@ export class PageViewPlugin extends MonkeyPatched implements Plugin {
...
+    private recordPageView = (invokeType: string) => {
+        this.context.recordPageView(this.createIdForCurrentPage(), invokeType);
     };
 
+    /**
+     * Creates the pageId for the current page.
+     * Output depends on the Partial Configuration.
+     * PATH_AND_HASH recommended for SPA.
+     * @returns pageId
+     */
```
qhanam, line 128: nit: I'm not sure this comment adds value. Consider removing or refactoring.

src/plugins/event-plugins/PageViewPlugin.ts
```diff
@@ -26,13 +39,17 @@ export class PageViewPlugin extends MonkeyPatched implements Plugin {
...
     }
 
     public getPluginId(): string {
         return this.pluginId;
     }
 
+    /**
+     * Monkey patches pushState and replaceState APIs to check whether
+     * route change has occurred.
+     */
```
qhanam, line 52: nit: I'm not sure this comment adds value. Consider removing or refactoring.

src/orchestration/Orchestration.ts
```diff
@@ -71,6 +84,7 @@ export type PartialConfig = {
     sessionEventLimit?: number;
     sessionLengthSeconds?: number;
     sessionSampleRate?: number;
+    spaTimeoutLimit?: number;
```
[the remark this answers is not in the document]
adebayor123: This config value was referring to the hard timeout constant. Changed the name to `spaHardTimeoutLimit`

src/orchestration/Orchestration.ts
```diff
@@ -58,6 +70,7 @@ export type PartialConfig = {
     disableAutoPageView?: boolean;
     dispatchInterval?: number;
     enableRumClient?: boolean;
+    enableSpaTracking?: boolean;
```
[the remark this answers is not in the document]
adebayor123: Refactored the current structure to remove enabling/disabling option for single page application page timing.
Now, pages created via route change will always have page load timing as a functionality.

src/plugins/event-plugins/PageViewPlugin.ts
```diff
@@ -74,27 +91,41 @@ export class PageViewPlugin extends MonkeyPatched implements Plugin {
...
         };
     };
 
+    /**
+     * Checks the current browser's URL against the current Page object's id.
+     * If they are different, indicate route change has occurred, start Page creation.
+     */
+    private recordRouteChange = () => {
+        const newUrl = location.pathname + location.hash;
+        if (this.context.getCurrentUrl() !== newUrl) {
```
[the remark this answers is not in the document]
adebayor123: Agreed, now that the feature will not rely on `PATH_AND_HASH` to distinguish unique pages, it will use the existing `recordPageView` method in PageManager to decide whether to create a new page or not.

src/plugins/event-plugins/TrackerPlugin.ts
```diff
@@ -0,0 +1,209 @@
...
+ */
+export class TrackerPlugin extends MonkeyPatched implements Plugin {
```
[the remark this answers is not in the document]
adebayor123: Great suggestion. After syncing offline, came to a conclusion that when a page loads initially, the type of pages that are created are either `resumed page` or `landing page`. As a result, 
(1) we no longer require the extra paramenter of `INVOKE_TYPE` in any of the methods 
(2) we treat single page app timing as a core behavior under `next page` 
(3) pull XMLHttpRequest and Fetch tracking logic into PageManager, as it is a core behavior of a PageManager

src/sessions/PageManager.ts
```diff
@@ -49,36 +66,84 @@ export class PageManager {
...
 
     public getPage(): Page | undefined {
         return this.page;
     }
 
+    public getCurrentUrl(): string {
+        return this.page.pageId;
+    }
+
+    public getTimeoutValue(): number {
```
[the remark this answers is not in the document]
adebayor123: Originally intended to be used in unit tests to verify the private attributes. Removed and will use other ways to check private attributes in unit tests.

src/sessions/PageManager.ts
```diff
@@ -49,36 +66,84 @@ export class PageManager {
...
+        return this.TIMEOUT_TIME;
+    }
+
+    public getFetchCounter(): number {
```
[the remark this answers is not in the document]
adebayor123: Originally intended to be used in unit tests to verify the private attributes. Removed and will use other ways to check private attributes in unit tests.

src/sessions/PageManager.ts
```diff
@@ -49,36 +66,84 @@ export class PageManager {
...
+        return this.fetchCounter;
+    }
+
     public getAttributes(): Attributes | undefined {
         return this.attributes;
     }
 
+    public getIntervalId() {
```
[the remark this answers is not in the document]
adebayor123: Originally intended to be used in unit tests to verify the private attributes. Removed and will use other ways to check private attributes in unit tests.

src/sessions/PageManager.ts
```diff
@@ -49,36 +66,84 @@ export class PageManager {
...
+        return this.periodicCheckId;
+    }
+
+    public getTimeoutId() {
```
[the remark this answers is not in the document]
adebayor123: Originally intended to be used in unit tests to verify the private attributes. Removed and will use other ways to check private attributes in unit tests.

src/sessions/PageManager.ts
```diff
@@ -49,36 +66,84 @@ export class PageManager {
...
+
+    public decrementFetchCounter = () => {
+        if (this.getFetchCounter() > 0) {
+            this.fetchCounter -= 1;
+        }
+    };
+
+    public recordPageView(
+        pageId: string,
+        invokeType = PAGE_INVOKE_TYPE.INITIAL_LOAD
```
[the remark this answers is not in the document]
adebayor123: Agreed, after some thought, pages that are "virtual pages" technically can only be a "next page" and pages created by initial loads will be either landing or resumed page.
Will reflect the change in next iteration.

src/sessions/PageManager.ts
```diff
@@ -49,36 +66,84 @@ export class PageManager {
...
+    ) {
         if (this.useCookies()) {
             this.recordInteraction = true;
         }
 
         if (!this.page && this.resumed) {
-            this.createResumedPage(pageId);
+            this.createPage(pageId, invokeType, PAGE_TYPE.RESUME);
```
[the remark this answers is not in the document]
adebayor123: Reverted the change. Will reflect in next revision.

src/plugins/event-plugins/TrackerPlugin.ts
adebayor123: Thanks for the suggestion, duly noted for future iterations!

app/spa.html
```diff
@@ -0,0 +1,187 @@
...
+    <head>
+        <title>RUM Integ Test for SPA</title>
+        <script src="./loader_spa.js"></script>
+
+        <link
+            rel="icon"
+            type="image/png"
+            href="https://awsmedia.s3.amazonaws.com/favicon.ico"
+        />
+        <link href="assets/css/style.css" rel="stylesheet" />
```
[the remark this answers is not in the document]
adebayor123: Removed CSS + external JS dependency

app/spa.html
```diff
@@ -0,0 +1,187 @@
...
+            <label for="command">Command : </label>
+            <input type="text" id="command" /><br /><br />
+            <label for="payload">Payload : </label>
+            <textarea id="payload"></textarea><br /><br />
+            <input
+                type="button"
+                id="submit"
+                value="Submit"
+                onclick="onSubmitCommand()"
+            />
```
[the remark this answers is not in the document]
adebayor123: Thanks for the suggestion! Will reflect in next revision.

app/spa.html
```diff
@@ -0,0 +1,187 @@
...
+                    <h3>STRAWBERRY</h3>
+                    <img src="" />
+                    <p></p>
+                    <span id="close" class="close">×</span>
+                </div>
+            </div>
+        </div>
+        <script src="//ajax.googleapis.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
+        <script src="//cdnjs.cloudflare.com/ajax/libs/handlebars.js/2.0.0/handlebars.min.js"></script>
+        <script src="assets/js/script.js"></script>
```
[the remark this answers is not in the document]
adebayor123: Updated the test application.

src/orchestration/Orchestration.ts
```diff
@@ -36,6 +36,12 @@ interface TelemetriesFunctor {
 
 type Telemetry = string | (string | object)[];
 
+export enum PAGE_TYPE {
```
adebayor123, line 39: Not being used.

adebayor123:
Debugging FetchPlugin integ test

app/spa.html
```diff
@@ -0,0 +1,199 @@
...
+                );
+                for (let i = 0; i < 40; i++) {
+                    setTimeout(domMutation, 99);
+                }
+                for (let i = 0; i < 30; i++) {
+                    createXMLRequest();
+                }
+            }
+
+            function defaultState() {
```
adebayor123, line 54: question: should we test using the back/forward, etc?

docs/cdn_commands.md
```diff
@@ -17,7 +17,7 @@ Commands may be sent to the web client after the snippet has executed. In the fo
             disableAutoPageView: true
         }
     );
-    cwr('recordPageView', window.location.hash);
+    cwr('recordPageView', '[window.location.hash, initial_load]');
```
adebayor123, line 20: This should be reverted

src/event-cache/EventCache.ts
```diff
@@ -2,13 +2,12 @@ import { Session, SessionManager } from '../sessions/SessionManager';
 import { v4 } from 'uuid';
 import { MetaData } from '../events/meta-data';
 import { Config } from '../orchestration/Orchestration';
-import { PageManager } from '../sessions/PageManager';
+import { PageManager, Page } from '../sessions/PageManager';
```
adebayor123, line 5: Need to remove.

src/plugins/Plugin.ts
```diff
@@ -1,9 +1,9 @@
+import { Page } from '../sessions/PageManager';
```
adebayor123, line 1: Need to remove.

src/sessions/__integ__/PageManager.test.ts
```diff
@@ -21,7 +21,7 @@ test('PageViewEventPlugin records landing page view event', async (t: TestContro
     // If we click too soon, the client/event collector plugin will not be loaded and will not record the click.
     // This could be a symptom of an issue with RUM web client load speed, or prioritization of script execution.
     await t
-        .wait(300)
+        .wait(3000)
```
adebayor123, line 24: Revert to 300

src/sessions/__tests__/PageManager.test.ts
```diff
@@ -11,6 +14,48 @@ Object.defineProperty(document, 'referrer', {
 });
 Object.defineProperty(document, 'title', { value: 'Amazon AWS Console' });
 
+const mockFetch = jest.fn(
```
adebayor123, line 17: Could add this to test-util and re-use by import in fetchPlugin and pageManager test files.

src/sessions/PageManager.ts
```diff
@@ -28,13 +36,30 @@ export type Attributes = {
  *
  * The interaction level is the order of a page in the sequence of pages sorted by the time they were viewed.
  */
-export class PageManager {
+export class PageManager extends MonkeyPatched {
```
qhanam, line 39: suggestion: Create a helper class to handle timing virtual page loads. This way we have separation of concerns by separating the page-interaction state from the page load timing state (increases maintainability).

src/sessions/PageManager.ts
```diff
@@ -44,11 +69,24 @@ export class PageManager {
...
+        this.periodicCheckerId = undefined;
+        this.activityTimeoutCheckerId = undefined;
+        this.domMutationObserver = new MutationObserver(
+            this.domMutationCallback
+        );
+        this.requestBuffer = new Set();
+        this.PERIODIC_CHECKER_INTERVAL = 100;
+        this.ACTIVITY_TIMEOUT_LIMIT = this.config.spaActivityTimeoutLimit
+            ? this.config.spaActivityTimeoutLimit
+            : 1000;
```
qhanam, line 88: suggestion: Instead of using this member variable, [set a default value](https://github.com/aws-observability/aws-rum-web/blob/main/src/orchestration/Orchestration.ts#L104) for `spaActivityTimeoutLimit` in `Orchestration`.

Same for `PERIODIC_CHECKER_INTERVAL` if we add a config setting for that as well.

src/sessions/PageManager.ts
```diff
@@ -156,10 +339,68 @@ export class PageManager {
...
     }
+
+    /**
+     * Checks whether the virtual page is still being loaded.
+     * If completed:
+     * (1) Clear the timers
+     * (2) Record data using NavigationEvent
+     * (3) Indicate current page has finished loading
+     */
+    private periodicCheckerCallback = () => {
```
qhanam, line 360: nit: Rename as verb of verb phrase (e.g., `checkLoadStatus`)

src/sessions/PageManager.ts
```diff
@@ -156,10 +339,68 @@ export class PageManager {
...
+        clearInterval(this.periodicCheckerId);
+        this.periodicCheckerId = undefined;
+        this.activityTimeoutCheckerId = undefined;
+        this.domMutationObserver.disconnect();
+        this.page.isLoaded = true;
+    };
+
+    /** Resets periodic check timer and updates page's latestEndTime. */
+    private domMutationCallback = () => {
+        this.page.latestEndTime = Math.max(this.page.latestEndTime, Date.now());
```
qhanam, line 383: question: Under what conditions would `this.page.latestEndTime` be greater than `Date.now()`?

src/sessions/PageManager.ts
```diff
@@ -156,10 +339,68 @@ export class PageManager {
...
+            this.domMutationObserver.disconnect();
+            this.recordVirtualPageNavigationEvent();
+            this.periodicCheckerId = undefined;
+            this.activityTimeoutCheckerId = undefined;
+            this.page.isLoaded = true;
+        }
+    };
+
+    /** Clears timers and disconnects observer to stop page timing. */
+    private activityTimeoutCallback = () => {
```
qhanam, line 373: nit: Rename as verb of verb phrase (e.g., `timeout`)

src/orchestration/Orchestration.ts
```diff
@@ -159,6 +160,7 @@ export type Config = {
     sessionEventLimit: number;
     sessionLengthSeconds: number;
     sessionSampleRate: number;
+    spaActivityTimeoutLimit?: number;
```
qhanam, line 163: issue: `SPA activity` is not common terminology. Let's think of something clearer. "Route change" is common terminology amongst front-end frameworks. Consider `routeChangeTimeout`.

Similarly, I recommend replacing other instances of the term `activity` in the code and documentation.

src/sessions/PageManager.ts
```diff
@@ -156,10 +339,68 @@ export class PageManager {
...
+    private domMutationCallback = () => {
+        this.page.latestEndTime = Math.max(this.page.latestEndTime, Date.now());
+        clearInterval(this.periodicCheckerId);
+        this.periodicCheckerId = setInterval(
+            this.periodicCheckerCallback,
+            this.PERIODIC_CHECKER_INTERVAL
+        );
+    };
+
+    private recordVirtualPageNavigationEvent() {
```
qhanam, line 391: nit: `recordRouteChangeNavigationEvent` (for consistency)

src/sessions/PageManager.ts
```diff
@@ -91,31 +162,143 @@ export class PageManager {
...
+                input: RequestInfo,
+                init?: RequestInit
+            ): Promise<Response> {
+                self.fetchCounter += 1;
+                return self.fetch(original, this, arguments, input, init);
+            };
+        };
+    };
+
+    private updateFetchCounter() {
```
qhanam, line 259: nit: `decrementFetchCounter`

src/sessions/PageManager.ts
```diff
@@ -91,31 +162,143 @@ export class PageManager {
...
+        const page = this.page;
+        // prevent NPE
+        if (page && !page.isLoaded) {
+            page.latestEndTime = Math.max(page.latestEndTime, Date.now());
+        }
+        if (this.fetchCounter > 0) {
```
qhanam, line 265: question: In what situation would fetchCounter be `0`?

src/sessions/PageManager.ts
```diff
@@ -1,14 +1,22 @@
...
 export const PAGE_VIEW_TYPE = 'com.amazon.rum.page_view_event';
 
 export type Page = {
     pageId: string;
     interaction: number;
     parentPageId?: string;
     start: number;
+    ongoingActivity: Set<XMLHttpRequest>;
+    latestEndTime: number;
+    isLoaded?: boolean;
```
qhanam, line 19: issue: These probably shouldn't be stored inside `Page` objects. `Page` objects are serialized and stored/restored by cookies inside `SessionManager`.

src/sessions/PageManager.ts
adebayor123: Sounds good, will reflect in next revision.

src/sessions/PageManager.ts
adebayor123: Will reflect in next revision.

src/sessions/PageManager.ts
adebayor123: Ideally, fetchCounter should never be 0, since for every fetch, whether successful or erroneous, is decrementing. This was mainly for just a safety check just in case the counter goes lower than 0. What do you think?

src/sessions/PageManager.ts
adebayor123: Makes sense, the pageManager/helper class can hold these and the functionality can be maintained. Will reflect in next revision.

src/sessions/PageManager.ts
qhanam: To go below zero, either (1) the counter fails to increment, or (2) `updateFetchCounter` gets executed multiple times for the same `fetch` call. I don't see a path where either of these could happen.

src/sessions/PageManager.ts
```diff
@@ -156,10 +339,68 @@ export class PageManager {
...
+        }
+    };
+
+    /** Clears timers and disconnects observer to stop page timing. */
+    private activityTimeoutCallback = () => {
+        clearInterval(this.periodicCheckerId);
+        this.periodicCheckerId = undefined;
+        this.activityTimeoutCheckerId = undefined;
+        this.domMutationObserver.disconnect();
+        this.page.isLoaded = true;
```
qhanam, line 378: question: Do we need to reset `fetchCounter` and `ongoingActivity` as well? Do we have a test that does a timeout followed by a completed virtual page load?

src/sessions/PageManager.ts
adebayor123: Added both to config, to give more flexibility to the customer's application.

src/sessions/PageManager.ts
adebayor123: Changed to `checkLoadStatus`

src/sessions/PageManager.ts
adebayor123: Will reflect in next revision.

src/sessions/PageManager.ts
adebayor123: Will reflect in next revision.

src/sessions/PageManager.ts
```diff
@@ -91,31 +162,143 @@ export class PageManager {
...
+    ): Promise<Response> => {
+        const self = this;
+        return original
+            .apply(thisArg, argsArray)
+            .then((response: Response) => {
+                self.updateFetchCounter();
+            })
+            .catch((error) => {
+                self.updateFetchCounter();
+            });
```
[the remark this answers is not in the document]
adebayor123: Will reflect in next revision.

src/sessions/PageManager.ts
adebayor123: Synced offline; fetchCounter should not go below zero as long as we do not reset the fetch counter after timeout is declared. This implies that if there's a faulty page load, it will affect subsequent virtual page loads until the fetch/xhr request are completed. If this becomes an issue, we will revise in the future.

src/orchestration/Orchestration.ts
```diff
@@ -215,6 +221,8 @@ export class Orchestration {
 
         if (this.config.enableRumClient) {
             this.enable();
+            // Enable pageManager's monkey patch after xhrPlugin and fetchPlugin are initialized
```
adebayor123, line 224: Need to revise this comment

src/orchestration/Orchestration.ts
```diff
@@ -215,6 +221,8 @@ export class Orchestration {
...
+            this.eventCache.enablePageManagerMonkeyPatch();
```
adebayor123, line 225: Update the name

src/orchestration/__tests__/Orchestration.test.ts
```diff
@@ -19,11 +19,13 @@ jest.mock('../../dispatch/Dispatch', () => ({
...
 const enableEventCache = jest.fn();
 const disableEventCache = jest.fn();
+const enableMonkeyPatch = jest.fn();
 
 jest.mock('../../event-cache/EventCache', () => ({
     EventCache: jest.fn().mockImplementation(() => ({
         enable: enableEventCache,
-        disable: disableEventCache
+        disable: disableEventCache,
+        enablePageManagerMonkeyPatch: enableMonkeyPatch
```
adebayor123, line 28: Update name

src/sessions/PageManager.ts
```diff
@@ -43,12 +45,23 @@ export class PageManager {
      */
     private recordInteraction: boolean;
 
+    /** Helper class to provide virtual page load timing functionality */
```
adebayor123, line 48: Remove

src/sessions/VirtualPageLoadTimer.ts
```diff
@@ -0,0 +1,245 @@
...
+        );
+        this.timeoutCheckerId = setTimeout(
+            this.declareTimeout,
+            this.config.routeChangeTimeout
+        );
+        this.domMutationObserver.observe(document, {
+            subtree: true,
+            childList: true,
+            attributes: true,
+            characterData: true
```
qhanam, line 79: question: Would it be sufficient to observe added/deleted nodes? I.e., and ignore attributes and character data?

src/sessions/VirtualPageLoadTimer.ts
```diff
@@ -0,0 +1,245 @@
...
+        const page = this.pageManager.getPage();
+        this.latestEndTime = Date.now();
+        clearInterval(this.periodicCheckerId);
+        this.periodicCheckerId = setInterval(
+            this.checkLoadStatus,
+            this.config.routeChangeLoadCheckInterval
+        );
+    };
+
+    private moveItemsFromBuffer = (item: any) => {
```
qhanam, line 226:
```suggestion
private moveItemsFromBuffer = (item: XMLHttpRequest) => {
```

src/sessions/VirtualPageLoadTimer.ts
```diff
@@ -0,0 +1,245 @@
...
+        this.ongoingRequests.add(item);
+    };
+
+    private recordRouteChangeNavigationEvent(page: Page) {
+        const virtualPageNavigationEvent: NavigationEvent = {
+            version: '1.0.0',
+            initiatorType: 'route_change',
+            navigationType: 'navigate',
+            startTime: page.start,
```
qhanam, line 235: issue: `page.start` is the time when `PageManager.recordPageView` is called. However, in the design we defined the start virtual page load duration as the most recent DOM click event.

src/orchestration/Orchestration.ts
```diff
@@ -215,6 +221,8 @@ export class Orchestration {
 
         if (this.config.enableRumClient) {
             this.enable();
+            // Enable monkey patch to interecept AJAX requests for virtual page load timing
+            this.eventCache.enableVirtualPageLoadTimerPatch();
```
qhanam, line 225: suggestion: Since this is always enabled, remove this sequence of calls and instead enable the monkey patch in the `VirtualPageLoadTimer` constructor.

src/sessions/VirtualPageLoadTimer.ts
```diff
@@ -0,0 +1,245 @@
...
+        argsArray: IArguments,
+        input: RequestInfo,
+        init?: RequestInit
+    ): Promise<Response> => {
+        const self = this;
+        return original
+            .apply(thisArg, argsArray)
+            .catch((error) => {
+                // Silently swallowing error
+            })
```
qhanam, line 157: issue: If we catch the error without re-throwing it, the error will not be visible to the application. This changes the application's behavior.

adebayor123: Ack, will reflect in next revision. Adjusted unit test cases accordingly.

src/sessions/VirtualPageLoadTimer.ts
adebayor123: For first iteration, we will remove unnecessary attributes and only observe added/deleted nodes.

package.json
```diff
@@ -99,6 +99,7 @@
         "@aws-sdk/signature-v4": "^3.36.0",
         "@aws-sdk/util-hex-encoding": "^3.36.0",
         "@babel/runtime": "^7.16.0",
+        "chromedriver": "^99.0.0",
```
qhanam, line 102: nit: Remove -- different Chrome versions require different drivers

src/sessions/VirtualPageLoadTimer.ts
```diff
@@ -0,0 +1,268 @@
...
+            this.updateLatestInteractionTime
+        );
+        document.body.addEventListener(
+            'keydown',
+            this.updateLatestInteractionTime
+        );
+        document.body.addEventListener(
+            'mouseleave',
+            this.updateLatestInteractionTime
+        );
```
qhanam, line 70: question: What is the reason for including this? Is there a risk that this could overwrite the actual interaction time?

src/sessions/VirtualPageLoadTimer.ts
```diff
@@ -0,0 +1,268 @@
...
+        this.latestEndTime = 0;
+        this.latestInteractionTime = 0;
+
+        this.config = config;
+        this.pageManager = pageManager;
+        this.record = record;
+        this.enable();
+
+        // Start tracking the timestamps
+        document.body.addEventListener(
```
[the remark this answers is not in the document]
adebayor123: In Firefox, it has to be `document.body` instead of `document`

app/spa.html
```diff
@@ -0,0 +1,265 @@
...
+                );
+                beginMutation(200);
+            }
+
+            function createXMLRequest() {
+                var xhr = new XMLHttpRequest((async = true));
+                var url =
+                    'https://dataplane.rum.us-west-2.amazonaws.com/appmonitors/abc123/';
+                xhr.open('GET', url, true);
+                xhr.send();
```
qhanam, line 143: issue: Could we access a local resources here instead of dataplane?

adebayor123: Agreed, I will replace it with the existing example endpoint we are using in other test applications.

src/sessions/__tests__/VirtualPageLoadTimer.test.ts
```diff
@@ -0,0 +1,567 @@
...
+        jest.resetModules();
+    });
+
+    test('when next page is created then items in requestBuffer should be added to ongoingRequests', async () => {
+        // Init
+        const config: Config = {
+            ...DEFAULT_CONFIG,
+            allowCookies: true
+        };
+        const pageManager: PageManager = new PageManager(config, record);
```
qhanam, line 52: issue: We should be instantiating VirtualPageLoadTimer directly in this test suite (i.e., we should not instantiate PageManager), since we are testing VirtualPageLoadTimer. 

There is also only one function we need to test -- `startTiming`

src/sessions/__tests__/VirtualPageLoadTimer.test.ts
```diff
@@ -0,0 +1,567 @@
...
+            helper['requestBuffer'].add(xhr);
+        }
+
+        // Create next page
+        pageManager.recordPageView('/console/home');
+
+        // Assert
+        expect(helper['ongoingRequests'].size).toEqual(3);
+        expect(helper['requestBuffer'].size).toEqual(0);
+        expect(helper['isPageLoaded']).toEqual(false);
```
qhanam, line 70: issue: Accessing private member variables is an anti-pattern (although *sometimes* necessary). This is because it is better to verify a property of the expected output rather than a property of the internal state.

In this case, we are verifying that the set of ongoing requests includes the requests from the requestBuffer. What we want is to verify that requests that started before the recordPageViewEvent will be included in the route change time.

src/sessions/VirtualPageLoadTimer.ts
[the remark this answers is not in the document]
adebayor123: Verify document.body is necessary for other event types in FF, Chrome, and IE

src/sessions/VirtualPageLoadTimer.ts
adebayor123: Synced offline, remove `eventListener` on `mouseleave`

src/sessions/VirtualPageLoadTimer.ts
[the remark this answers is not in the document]
adebayor123: Going with document.addEventListener as we no longer track `mouseleave` event that was incompatible in FireFox

PR #120, opened 2022-03-15
fix: Add version property to http-event schema and convert PromiseRejection.reason into string

adebayor123:
Problem
- When HTTP Event with an error is created, it uses `js-error-event` to populate the error field. `js-error-event` contains the field `version`, which is not included `http-event`'s `error` field. Due to this inconsistency that is also reflected on the dataplane's validation, any HTTP event with an error field is rejected. 
- When JSErrorPlugin handles `unhandledrejection` event, it adds `PromiseRejectionEvent.reason` as the `message` field. However, if the `PromiseRejectionEvent.reason` is an empty object rather than a string message, it is returned in the form of `{}`, which fails dataplane's validation for JsErrorEvents.

Fix
- This PR contains changes to `JsErrorPlugin` and its unit test cases to handle situations where the `PromiseRejectionEvent.reason` is an empty object, and also changes the event schema for `http-event`.

---

### DIFF
--- a/src/plugins/event-plugins/JsErrorPlugin.ts
+++ b/src/plugins/event-plugins/JsErrorPlugin.ts
@@ -73,7 +73,10 @@ export class JsErrorPlugin implements Plugin {
         const errorEvent: JSErrorEvent = {
             version: '1.0.0',
             type: event.type,
-            message: event.reason
+            message:
+                typeof event.reason === 'string'
+                    ? event.reason
+                    : JSON.stringify(event.reason)
         };
         this.recordEvent(JS_ERROR_EVENT_TYPE, errorEvent);
     };
--- a/src/event-schemas/http-event.json
+++ b/src/event-schemas/http-event.json
@@ -30,6 +30,10 @@
         "error": {
             "type": "object",
             "properties": {
+                "version": {
+                    "type": "string",
+                    "description": "JSErrorEvent schema version."
+                },
                 "type": {
                     "type": "string",
                     "description": "Error type (eg., TypeError, ParseError, etc)."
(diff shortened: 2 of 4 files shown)


qhanam:
Does this change fix two separate bugs? Would it make sense to file separate PRs for them?

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -73,7 +73,10 @@ export class JsErrorPlugin implements Plugin {
         const errorEvent: JSErrorEvent = {
             version: '1.0.0',
             type: event.type,
-            message: event.reason
+            message:
+                typeof event.reason === 'string'
+                    ? event.reason
+                    : JSON.stringify(event.reason)
         };
         this.recordEvent(JS_ERROR_EVENT_TYPE, errorEvent);
```
qhanam, line 81: In the case where `event.reason` is an object, it likely inherits from `Error`. In this case, we should extract type, message, file, line number, stack trace, etc.

This can be accomplished with the proposed change.
```suggestion
        this.record(event.reason);
```

qhanam: Actually, the above omits `event.type`, might need to think through a little more.

adebayor123:
> Does this change fix two separate bugs? Would it make sense to file separate PRs for them?

Agreed, will split the PRs into two, one for `http-event` schema change, other for `JsErrorPlugin` fix

src/plugins/event-plugins/JsErrorPlugin.ts
adebayor123: Will investigate a bit further for this bug fix and raise a separate PR once ready.

PR #123, opened 2022-03-17
fix: Parse unhandledrejection error objects

adebayor123:
Problem:
The current implementation to handle `unhandledrejection` events simply assumes that the `PromiseRejectEvent.reason` is a string and populates it under `js-error-event.message` as a string. 
As a result, if the `PromiseRejectEvent.reason` is actually an error object, it does not properly stringify the error object, and instead populates the `js-error-event.message` field with an empty object, `{}`.
This results in the `JsErrorEvent` being rejected by the dataplane, as the `js-error-event.message` field is expected to be a string.

Fix:
This PR addresses the issue by doing two things:
1. Reuse `eventHandler` in `JsErrorPlugin` in `promiseRejectHandler` by extracting the `event.type` and `event.reason` and casting into an `ErrorEvent`.
2. Reuse parsing methods in `js-error-util` to create a `JsErrorEvent` that properly reflects the `unhandledrejection` event.

---

### DIFF
--- a/src/plugins/event-plugins/JsErrorPlugin.ts
+++ b/src/plugins/event-plugins/JsErrorPlugin.ts
@@ -1,10 +1,6 @@
-import { JSErrorEvent } from '../../events/js-error-event';
 import { RecordEvent, Plugin, PluginContext } from '../Plugin';
 import { JS_ERROR_EVENT_TYPE } from '../utils/constant';
-import {
-    errorEventToJsErrorEvent,
-    buildBaseJsErrorEvent
-} from '../utils/js-error-utils';
+import { errorEventToJsErrorEvent } from '../utils/js-error-utils';
 
 export const JS_ERROR_EVENT_PLUGIN_ID = 'com.amazonaws.rum.js-error';
 
--- a/src/plugins/event-plugins/JsErrorPlugin.ts
+++ b/src/plugins/event-plugins/JsErrorPlugin.ts
@@ -69,7 +69,6 @@ export class JsErrorPlugin implements Plugin {
     };
 
     private promiseRejectEventHandler = (event: PromiseRejectionEvent) => {
-        // TODO: Improve current behavior to correctly indicate error comes from PromiseRejectionEvent
         this.eventHandler({
             type: event.type,
             error: event.reason
--- a/src/plugins/event-plugins/JsErrorPlugin.ts
+++ b/src/plugins/event-plugins/JsErrorPlugin.ts
@@ -1,7 +1,10 @@
 import { JSErrorEvent } from '../../events/js-error-event';
 import { RecordEvent, Plugin, PluginContext } from '../Plugin';
 import { JS_ERROR_EVENT_TYPE } from '../utils/constant';
-import { errorEventToJsErrorEvent } from '../utils/js-error-utils';
+import {
+    errorEventToJsErrorEvent,
+    buildBaseJsErrorEvent
+} from '../utils/js-error-utils';
 
 export const JS_ERROR_EVENT_PLUGIN_ID = 'com.amazonaws.rum.js-error';
 
@@ -70,12 +73,10 @@ export class JsErrorPlugin implements Plugin {
     };
 
     private promiseRejectEventHandler = (event: PromiseRejectionEvent) => {
-        const errorEvent: JSErrorEvent = {
-            version: '1.0.0',
+        this.eventHandler({
             type: event.type,
-            message: event.reason
-        };
-        this.recordEvent(JS_ERROR_EVENT_TYPE, errorEvent);
+            error: event.reason
+        } as ErrorEvent);
     };
 
     private addEventHandler(): void {
--- a/src/plugins/event-plugins/JsErrorPlugin.ts
+++ b/src/plugins/event-plugins/JsErrorPlugin.ts
@@ -69,6 +69,7 @@ export class JsErrorPlugin implements Plugin {
     };
 
     private promiseRejectEventHandler = (event: PromiseRejectionEvent) => {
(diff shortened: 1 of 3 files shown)


src/plugins/event-plugins/__tests__/JsErrorPlugin.test.ts
```diff
@@ -422,4 +422,102 @@ describe('JsErrorPlugin tests', () => {
...
+        );
+        plugin.disable();
+
+        // Assert
+        expect(record).toHaveBeenCalledTimes(1);
+        expect(record.mock.calls[0][0]).toEqual(JS_ERROR_EVENT_TYPE);
+        expect(record.mock.calls[0][1]).toMatchObject(
+            expect.objectContaining({
+                version: '1.0.0',
+                type: 'unhandledrejection: TypeError',
```
adebayor123, line 517: I think we should indicate it's from an `unhandledrejection` by appending to the string instead of having to wrap the `JsErrorEvent`. Let me know what you think!

src/plugins/utils/js-error-utils.ts
```diff
@@ -48,7 +48,10 @@ const appendErrorPrimitiveDetails = (
     rumEvent: JSErrorEvent,
     error: any
 ): void => {
-    rumEvent.type = error.toString();
+    // Keep unhandledrejection as type as it will write to rumEvent.message
+    if (rumEvent.type !== 'unhandledrejection') {
+        rumEvent.type = error.toString();
+    }
```
qhanam, line 54: issue: We want the behavior of promise rejection events to be the same as the behavior of error events. This is so that monitoring dashboards can display errors from both sources in the same way.

src/plugins/utils/js-error-utils.ts
```diff
@@ -60,7 +63,9 @@ const appendErrorObjectDetails = (
     // error may extend Error here, but it is not guaranteed (i.e., it could
     // be any object)
     if (error.name) {
-        rumEvent.type = error.name;
+        rumEvent.type === 'unhandledrejection'
+            ? (rumEvent.type += ': ' + error.name)
+            : (rumEvent.type = error.name);
```
qhanam, line 68: issue: Same here. We want to only display the error name if it is available, because this is the current behavior.

src/plugins/event-plugins/__tests__/JsErrorPlugin.test.ts
qhanam: I believe you're thinking that it would be useful to record the fact that this error was caught as a promise rejection event rather than an error event.

Making this distinction could be useful for debugging. However, we must apply conventions consistently across error types. The current behavior for uncaught errors does not prefix the error type with how the error was caught.

We can (and maybe should) change this behavior, but this will require some planning. E.g., do we add another field, do we add prefixes like `Uncaught TypeError`, `Uncaught (in promise) TypeError`, `Caught TypeError`, etc.

src/plugins/utils/js-error-utils.ts
adebayor123: Agreed, will reflect on next revision.

src/plugins/utils/js-error-utils.ts
adebayor123: Agreed, will reflect on next revision.

src/plugins/event-plugins/__tests__/JsErrorPlugin.test.ts
adebayor123: I do agree with your point that we should be consistent with the current conventions.
In the next revision, I will add a comment in the code mentioning we should make this improvement in the future and also raise an action item.

src/plugins/utils/js-error-utils.ts
adebayor123: The reason for this change is due to the unit test case `'when an unhandled rejection is thrown then the plugin records the name, message'` in `JsErrorPlugin.test.ts`.

If the error's reason is simply a string, without this change the output `JsErrorEvent` becomes `{"message": "Something went wrong!", "type": "Something went wrong!", "version": "1.0.0"}`.

Even though we set `type: event.type` and `error: event.reason` in `promiseRejectEventHandler`, when we build the `JsErrorEvent` while the `event.reason` is a string, it goes through `appendErrorPrimitiveDetails` in `js-error-utils` that simply sets both the `rumEvent.type` and `rumEvent.message` with `error.toString()`.

As a result, even though the existing, expected behavior in the case where the output `JsErrorEvent` is `{"message": "Something went wrong!", "type": "unhandledrejection", "version": "1.0.0"}`, it becomes `{"message": "Something went wrong!", "type": "Something went wrong!", "version": "1.0.0"}` as mentioned above.

adebayor123: It seems that the existing convention for a `PromiseRejectEvent` when its `reason` is a string is to keep the output `JsErrorEvent.type` as `'unhandledrejection'` and update `JsErrorEvent.message` as `error.toString()`

I will raise a new revision maintaining this section so we can discuss again.

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -70,12 +69,11 @@ export class JsErrorPlugin implements Plugin {
     };
 
     private promiseRejectEventHandler = (event: PromiseRejectionEvent) => {
-        const errorEvent: JSErrorEvent = {
-            version: '1.0.0',
+        // TODO: Improve current behavior to correctly indicate error comes from PromiseRejectionEvent
```
qhanam, line 72: nit: Remove TODOs -- use issues to track future work

qhanam (approved):

PR #124, opened 2022-03-17
feat: Capture SPA route change load timing

adebayor123:
**Context**:

Currently, RUM Web Client does not provide a mechanism to calculate the route change load timing in single page applications, as the `navigation-event` is only sent when the window's `load` event is fired.

**Changes**:

1. Detect when route changes occur, and infer the starting point of the route change based on the user's latest activity.
2. Create a timing mechanism within VirtualPageLoadTimer using Timers.
3. Intercept outgoing AJAX requests in the form of `Fetch` and `XMLHttpRequest` and use it as a datapoint to calculate the current page's load time.
4. Observe DOM mutations to calculate the current page's load time.

---


qhanam (approved):
Reviewed in https://github.com/aws-observability/aws-rum-web/pull/113

PR #129, opened 2022-03-29
fix: Dynamic DOM event handlers configurable

limhjgrace:
Currently, dynamic DOM event handlers are enabled by default, which may impact customers' application due to MutationObserver. This fix will set dynamic DOM event handlers (MutationObserver) to not be enabled, by default, and will let customers enable it as a configuration, or to toggle it on/off. 

---


src/orchestration/Orchestration.ts
```diff
@@ -135,6 +137,7 @@ export type CookieAttributes = {
 
 export type Config = {
     allowCookies: boolean;
+    allowDynamicDomEventListeners: boolean;
```
qhanam, line 140: issue: The design convention for plugins is that have their own configurations. Consider adding this configuration option to `DomEventPluginConfig` instead of the global configuration.

src/plugins/event-plugins/DomEventPlugin.ts
```diff
@@ -31,10 +31,12 @@ export type PartialDomEventPluginConfig = {
 };
 
 export type DomEventPluginConfig = {
+    allowDynamicDomEventListeners?: boolean;
```
qhanam, line 34: issue: Let's consider some alternatives for the config option name:
- `enableDynamicUpdates`
- `enableMutationObserver`

src/orchestration/Orchestration.ts
```diff
@@ -282,6 +285,13 @@ export class Orchestration {
         this.config.allowCookies = allow;
     }
 
+    /**
+     * @param allow when {@code false}, the RUM web client will not listen to dynamically added DOM targets.
+     */
+    public allowDynamicDomEventListeners(allow: boolean) {
```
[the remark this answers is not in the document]
limhjgrace: As discussed offline, current use case is that owners will want dynamic event listeners on or off the entire time. We can add this later if it's common use case. Will omit for now.

src/orchestration/Orchestration.ts
limhjgrace: Will refactor to add to `DomEventPluginConfig`

src/plugins/event-plugins/DomEventPlugin.ts
limhjgrace: Will refactor to use `enableMutationObserver`

qhanam (approved):

PR #133, opened 2022-04-13
update: fix repeat code in package.json and remove old hook module

maniator:
### Notes

- Add an `.nvmrc` so that when a new person creates a PR, the version of NPM is correct and the package-lock doesn't change versions
- Updates to `package.json`
   - utilize `npm-run-all` and its `run-s` command to run certain scripts in serial (can switch to `run-p` if needed)
      - can see docs here: https://github.com/mysticatea/npm-run-all/blob/master/docs/run-s.md
   - remove very ancient module `pre-commit` in favor of a standard `husky`/`lint-staged` combination

There probably can be more optimizations that can be made as well to the build process and the `package.json` that can be followed up in future updates

---


qhanam:
Hi @maniator thanks for the contribution!

I think adding nvmrc and npm-run-all are good changes. 

It is not clear to me why we would want to replace `pre-commit`.

My preference is to run `prettier` as a check. (1) the CI pipeline should fail when the code does not conform, and (2) some files should be added to .prettierrc, rather that being auto-formatted.

Could you resolve these and the failing CI? Thanks!

maniator:
> Hi @maniator thanks for the contribution!
> 
> I think adding nvmrc and npm-run-all are good changes.
> 
> It is not clear to me why we would want to replace `pre-commit`.
> 
> My preference is to run `prettier` as a check. (1) the CI pipeline should fail when the code does not conform, and (2) some files should be added to .prettierrc, rather that being auto-formatted.
> 
> Could you resolve these and the failing CI? Thanks!

I will resolve rebase issues for sure :-) 

in terms of `pre-commit` -- it is a library that has not been updated in a long time and has quite a few issues that husky/lint-staged fix

I am not sure what you are referring to in terms of the prettier issues.

qhanam:
Please add the disclaimer found in the PR template to your PR description.

package.json
```diff
@@ -108,10 +115,10 @@
         "uuid": "^8.3.2",
         "web-vitals": "^1.1.1"
     },
-    "pre-commit": [
-        "lint",
-        "prettier"
-    ],
+    "lint-staged": {
+        "*.{js,ts}": [ "npm run lint"],
+        "*.{js,html,md,yml}": [ "npm run prettier:fix"]
```
qhanam, line 120: 1. Our preference is that prettier check during pre-commit, rather than format during pre-commit. 
2. Please include typescript and json in this list.

maniator: For the first one, I can update, but why not format on pre-commit?
Wouldn't that make it easier for people to make updates?

Also for number 2 -- will add, was a copy/paste from docs.

maniator:
> Please add the disclaimer found in the PR template to your PR description.

Updated.

maniator:
@qhanam 

What is the reason for not having auto-fix in the pre-commit stage?

maniator:
@qhanam any ideas why unit tests are failing in github actions, but passing fine locally?

https://github.com/aws-observability/aws-rum-web/runs/6326630073?check_suite_focus=true

maniator:
@qhanam I am going to break this into multiple PRs (as it might find the CI issue)

PR #137, opened 2022-04-22
test: Integrate Playwright and define initial cases to verify endpoint success and capture eventIds

adebayor123:
Problem:
- Currently smoke tests are performed manually, and in order to gain assurance on newly released web client's correct behavior, automated smoke test is required to (1) verify web client recording events and delivering to data plane successfully (2) verify the complete ingestion of recorded events.

This PR:
- As mentioned above, there are two levels of verification. This PR contains the initial set of test cases & test framework foundation to record events and verifying successful delivery to the data plane. Subsequent commits will include test case to conduct the second verification.
- Endpoint, monitorId, url are replaced with placeholder values to mimic accessing GitHub secrets.
- All the test cases in this PR do three things: (1) trigger an event and record (2) verify successful delivery to data plane by only collecting successful network requests made to the data plane (3) save the unique eventIds in an array to use for final ingestion verification.
- Only a single browser session is used and the test cases execute in the defined order; this allows the test to mimic human interaction with a web page and prevent creation of duplicate events, such as `session_start_event`.

Note:
- Current smoke test application only emits two core web vital metrics; in future iterations, need to update the test application so cumulative layout shift metric is emitted as well.

---


playwright.config.ts
```diff
@@ -0,0 +1,22 @@
+// playwright.config.js
+// @ts-check
+const { devices } = require('@playwright/test');
+
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
```
qhanam, line 5: nit: remove unused

src/__smoke-test__/smoke-test.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+        const expectedPage = '/smoke.html';
+        const requestBody = [];
+
+        // Collect successful requests
+        page.on('response', (response) =>
+            filterSuccessfulResponse(response, requestBody)
+        );
+
+        // Wait until fetch is complete
+        await page.waitForTimeout(10000);
```
qhanam, line 85: Can we use waitFor (event to happen) instead of sleep? Sleep create nondeterminism (flakiness) and increases the length of time the test runs.

src/__smoke-test__/smoke-test.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+    PERFORMANCE_NAVIGATION_EVENT_TYPE,
+    PERFORMANCE_RESOURCE_EVENT_TYPE,
+    PAGE_VIEW_EVENT_TYPE,
+    SESSION_START_EVENT_TYPE
+} from '../plugins/utils/constant';
+
+// Secrete constant values retrieved from GitHub
+const endpoint = '${process.env.ENDPOINT}';
+const monitorId = '${process.env.MONITOR_ID}';
+const url = '${process.env.SMOKE_URL}';
```
qhanam, line 16: suggestion: Make these command line arguments instead of environment variables, since they are inputs to the test and do not need to be shared across processes.

package.json
```diff
@@ -47,11 +47,13 @@
         "postinteg:local:nightwatch:chrome": "kill $(lsof -t -i:8080)",
         "preinteg:local:nightwatch:firefox": "http-server ./build/dev -s &",
         "integ:local:nightwatch:firefox": "nightwatch -e firefox",
-        "postinteg:local:nightwatch:firefox": "kill $(lsof -t -i:8080)"
+        "postinteg:local:nightwatch:firefox": "kill $(lsof -t -i:8080)",
+        "smoke": "npx playwright test"
```
qhanam, line 51: question: Under what situations will this target be executed? Specifically, I'm thinking about how we will execute these tests locally. Should we add an additional target which spins up a server locally, as we do for the other tests here?

src/__smoke-test__/smoke-test.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+    const request = response.request();
+    if (
+        request &&
+        response.status() === 200 &&
+        request.method() === 'POST' &&
+        response.url() === endpoint + monitorId + '/' &&
+        request.postData()
+    ) {
+        requestBody.push(request.postData());
+    }
```
qhanam, line 50: suggestion: Array functions like `filter` and `map` may make this a bit cleaner.

src/__smoke-test__/smoke-test.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+        eventList.push(parsedPayload[i].id);
+    }
+}
+
+function getEventsByType(requestBody, eventType) {
+    return requestBody.filter((e) => e.type === eventType);
+}
+
+function filterSuccessfulResponse(response, requestBody) {
+    // filter out irrelevant, unsuccessful network request/response
```
qhanam, line 40: question: What are irrelevant and unsuccessful network requests?

src/__smoke-test__/smoke-test.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+
+test.describe('Smoke test fixtures', () => {
+    let page;
+    const eventIds = [];
+
+    // Use single browser context to prevent multiple sessions per test
+    test.beforeAll(async ({ browser }) => {
+        const context = await browser.newContext();
+        page = await context.newPage();
+        await page.goto(url);
```
qhanam, line 64: issue: Tests should be independent of each other

src/__smoke-test__/smoke-test.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+    });
+
+    test.afterAll(async ({ browser }) => {
+        browser.close;
+    });
+
+    test('test initial events are emitted', async () => {
```
qhanam, line 71: issue: We should standardize the way we name tests. We already use the "when/then" format in unit tests, so I suggest we adopt this for our integ tests as well. For example: _"When PutRumEvents is called then the response code is 200"_

src/__smoke-test__/smoke-test.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+        const pageId = JSON.parse(pageView[0].metadata).pageId;
+        const navigationPageId = JSON.parse(pageView[0].metadata).pageId;
+        expect(pageId).toEqual(expectedPage);
+        expect(navigationPageId).toEqual(expectedPage);
+
+        // Add all event ids to eventIds array
+        addEventsToList(parsedPayload, eventIds);
+    });
+
+    test('test FID and LCP events are emitted', async () => {
```
qhanam, line 119: question: What value do these tests add above the integration tests we currently have?

If we’re just checking that we get a 200 from DP, then I think one test is sufficient to cover it.

Later, when we verify that events were ingested, we will want to split up the tests in this manner.

src/__smoke-test__/smoke-test.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+        // Parse metadata to verify pageId
+        for (let i = 0; i < pageView.length; i++) {
+            const pageId = JSON.parse(pageView[i].metadata).pageId;
+            const navigationPageId = JSON.parse(pageView[i].metadata).pageId;
+            expect(pageId).toEqual(expectedPages[i]);
+            expect(navigationPageId).toEqual(pageId);
+        }
+
+        // Add all event ids to eventIds array
+        addEventsToList(parsedPayload, eventIds);
```
qhanam, line 201: issue: What behavior does this test specify? And what property does it need to verify. 

I.e., if we name the test _"When x then y"_, then what is _y_?

In general, we want tests to specify a single behavior.

src/__smoke-test__/smoke-test.spec.ts
```diff
@@ -0,0 +1,446 @@
...
+    FID_EVENT_TYPE,
+    PERFORMANCE_NAVIGATION_EVENT_TYPE,
+    PERFORMANCE_RESOURCE_EVENT_TYPE,
+    PAGE_VIEW_EVENT_TYPE,
+    SESSION_START_EVENT_TYPE
+} from '../plugins/utils/constant';
+
+// Secrete constant values retrieved from GitHub
+const endpoint = '${process.env.ENDPOINT}';
+const monitorId = '${process.env.MONITOR_ID}';
```
qhanam, line 15: question: How was the app monitor created? Are we assuming that it already exists?

package.json
adebayor123: Yes, for local development purpose we do need that option; will add in the next revision.

src/__smoke-test__/smoke-test.spec.ts
adebayor123: As mentioned before, for the scope of smoke tests, we do not need to create new application monitors for every run, as we can filter the events retrieved from CP by a timestamp.

src/__smoke-test__/smoke-test.spec.ts
adebayor123: Irrelevant: requests to sts, fetching resources, etc.
Unsuccessful: all data plane requests that are not 200

This allows us to capture only the data plane requests that were successful, as the smoke test (1) verifies events generated and sent to data plane successfully (2) verifies whether all the successfully sent events are ingested.

src/__smoke-test__/smoke-test.spec.ts
adebayor123: Good call, will reflect in next revision.

src/__smoke-test__/smoke-test.spec.ts
adebayor123: I agree that tests should be independent, but there are three reasons as to why I wanted to have the test cases use the same browser session:
1. unique browser session leads to browser opening and closing for all the test cases
2. certain events such as session_start_event will be emitted over and over
3. when we perform the tests manually, we do not open and close the page after clicking/testing one button; instead, we go through all the buttons under one session

src/__smoke-test__/smoke-test.spec.ts
adebayor123: Agreed, will reflect in next revision.

src/__smoke-test__/smoke-test.spec.ts
adebayor123: Didn't know waitForTimeout has those side effects. Just looked into the official documentation, I think we could use waitForRequest or waitForResponse to have the same effect. Will reflect in next revision.

src/__smoke-test__/smoke-test.spec.ts
adebayor123: Did some research to find if Playwright accepts custom parameters from the command line, and the short answer is Playwright does not support this option.

Several feature requests were raised to support this, but the Playwright team stated that using env variables is the preferred option at the moment.

As a result, the proposed solution is the following:
1. define the npm command as `"cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT npx playwright test"` 
2. when we run `npm run smoke`, we pass in the parameters as `URL=test_url.com MONITOR=monitor_name ENDPOINT=endpoint  npm run smoke`

src/__smoke-test__/smoke-test.spec.ts
adebayor123: Response is an object, so array functions are not applicable. Removed two conditions, `request` and `request.postData()` as a `response` object will always have `request` object and a `request` object will always have `postData()`

playwright.local.config.ts
```diff
@@ -0,0 +1,26 @@
+// playwright.local.config.js
```
adebayor123, line 1: Based on https://playwright.dev/docs/test-advanced#launching-a-development-web-server-during-the-tests and https://playwright.dev/docs/test-cli, in order to have different configs (to allow local development), require two different config files.

src/__smoke-test__/dataplane-integ.spec.ts.ts
```diff
@@ -0,0 +1,83 @@
+import { test, expect } from '@playwright/test';
+import {
+    PERFORMANCE_NAVIGATION_EVENT_TYPE,
+    PERFORMANCE_RESOURCE_EVENT_TYPE,
+    PAGE_VIEW_EVENT_TYPE,
+    SESSION_START_EVENT_TYPE
+} from '../plugins/utils/constant';
+
+// Environment variables set through CLI command
```
adebayor123, line 9: Did some research to find if Playwright accepts custom parameters from the command line, and the short answer is Playwright does not support this option.

Several feature requests were raised to request this, but the Playwright team stated that using env variables is the preferred option at the moment.

As a result, the proposed solution is the following:

define the npm command as "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT npx playwright test"
when we run npm run smoke, we pass in the parameters as URL=test_url.com MONITOR=monitor_name ENDPOINT=endpoint npm run smoke

src/__smoke-test__/dataplane-integ.spec.ts
```diff
@@ -0,0 +1,83 @@
+import { test, expect } from '@playwright/test';
+import {
+    PERFORMANCE_NAVIGATION_EVENT_TYPE,
+    PERFORMANCE_RESOURCE_EVENT_TYPE,
+    PAGE_VIEW_EVENT_TYPE,
+    SESSION_START_EVENT_TYPE
+} from '../plugins/utils/constant';
+
+// Environment variables set through CLI command
```
adebayor123, line 9: Did some research to find if Playwright accepts custom parameters from the command line, and the short answer is Playwright does not support this option.

Several feature requests were raised to request this, but the Playwright team stated that using env variables is the preferred option at the moment.

As a result, the proposed solution is the following:

define the npm command as "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT npx playwright test"
when we run npm run smoke, we pass in the parameters as URL=test_url.com MONITOR=monitor_name ENDPOINT=endpoint npm run smoke

qhanam (approved):

PR #141, opened 2022-05-02
Add Code QL analysis

vasireddy99:
This PR  will add the CodeQL analysis, which helps to discover vulnerabilities across a codebase, with every change that is being pushed.

---

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,63 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
+        # Learn more:
+        # https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
+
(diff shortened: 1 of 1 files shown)


qhanam:
@vasireddy99 please provide context for this PR in the description.
1. What is the problem being solved?
2. How does this change solve the problem?

vasireddy99:
> @vasireddy99 please provide context for this PR in the description.
> 
> 1. What is the problem being solved?
> 2. How does this change solve the problem?

thanks @qhanam :-)

qhanam:
Hi @vasireddy99, this still isn't enough information for us to make a decision on whether or not to integrate this static analysis tool into our build process.

I see this is a tool [developed by GitHub](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql). However, it is not clear to me what its behavior is, and what benefit we get from integrating it.

For example, once the database has been created, what do we do with it? Does GitHub automatically run queries against it to identify known or emerging vulnerabilities? Do we need to run queries against it ourselves? Which queries would we run, and why?

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,63 @@
...
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
```
qhanam, line 11: issue: Please clean up this file by removing copied instructional comments such as these.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,63 @@
...
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        # queries: ./path/to/local/query, your-org/your-repo/queries@main
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v1
```
qhanam, line 49: issue: This is a TypeScript repo, I don't think this step applies.

Is there a reason why the NPM build would need to run? How does CodeQL handle JavaScript?

vasireddy99:
Fixed the lint check in the PR

qhanam (approved):

PR #144, opened 2022-05-04
chore: 🤖 remove duplicate ts configurations and instead extend

maniator:
tsconfigs can extend from each other 

Now if one is updated, the others also will be updated (if need be) based off changes from the root tsconfig

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ safaridriver.log
 tests_output
 app/smoke.html
 logs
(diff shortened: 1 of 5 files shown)


maniator:
Hmmm

I can fix the prettier issues

That's why it's better for it to be auto fixed on pre-commit for the most part...

qhanam (approved):
Looks good; thank you!

PR #145, opened 2022-05-05
fix: 🐛 fix issue that decrement fetch counter was called always

maniator:
Basically the finally call of fetch was not only being called on
finally.

---

### DIFF
--- a/src/sessions/VirtualPageLoadTimer.ts
+++ b/src/sessions/VirtualPageLoadTimer.ts
@@ -168,7 +168,7 @@ export class VirtualPageLoadTimer extends MonkeyPatched {
             .catch((error) => {
                 throw error;
             })
-            .finally(self.decrementFetchCounter());
+            .finally(self.decrementFetchCounter.bind(self));
     };
 
     /**
--- a/src/sessions/VirtualPageLoadTimer.ts
+++ b/src/sessions/VirtualPageLoadTimer.ts
@@ -162,13 +162,12 @@ export class VirtualPageLoadTimer extends MonkeyPatched {
         input: RequestInfo,
         init?: RequestInit
     ): Promise<Response> => {
-        const self = this;
         return original
             .apply(thisArg, argsArray)
             .catch((error) => {
                 throw error;
             })
-            .finally(self.decrementFetchCounter.bind(self));
+            .finally(this.decrementFetchCounter);
     };
 
     /**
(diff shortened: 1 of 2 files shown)


src/sessions/VirtualPageLoadTimer.ts
```diff
@@ -168,7 +168,7 @@ export class VirtualPageLoadTimer extends MonkeyPatched {
             .catch((error) => {
                 throw error;
             })
-            .finally(self.decrementFetchCounter());
+            .finally(self.decrementFetchCounter.bind(self));
```
maniator, line 171: basically we want to only call the `decrementFetchCounter` when `finally` is invoked and not when the wrapping function is called.

qhanam, line 171:
```suggestion
            .finally(() => self.decrementFetchCounter());
```

qhanam: Could you add a unit test that covers the behavior?

maniator: Same idea.

maniator: sure

qhanam: I'm trying to propose something [more idiomatic](https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions). However I think my prior suggestion is incorrect. We can make `decrementFetchCounter` an arrow function, and then we do not need to bind. I don't think we need `self` here either, since `fetch` is an arrow function.
```suggestion
            .finally(this.decrementFetchCounter);
```

```typescript
    private decrementFetchCounter = () => {
        if (!this.isPageLoaded) {
            this.latestEndTime = Date.now();
        }
        this.fetchCounter -= 1;
    };
```

maniator: Sounds good to me :-)

maniator: Done

maniator: Complete

qhanam (approved):
Nice, thank you!

maniator:
> Nice, thank you!

Happy to help :-)

adebayor123 (approved):

maniator:
@adebayor123 @qhanam 

Good to merge? (I don't have the ability)

PR #148, opened 2022-05-06
refactor: 💡 utilize configuration files for integ tests

maniator:
Utilize config files for the integration tests instead of having to
write a new npm script for each (aside for ie which is special)

Forked from work here: https://github.com/aws-observability/aws-rum-web/pull/133 

Can probably follow-up in the future with something like this: https://testcafe.io/documentation/402817/guides/continuous-integration/github-actions

---


.testcafe.browser.js
```diff
@@ -0,0 +1,7 @@
+const rootConfig = require('./.testcaferc.json');
+
+module.exports = {
+    ...rootConfig,
+    browsers: ['edge', 'firefox', 'chrome'],
+    concurrency: 1
```
adebayor123, line 6: What are your thoughts on increasing the concurrency value to 3 (edge, firefox, chrome) instead of 1? Would such parallelization work for CI?

package.json
```diff
@@ -32,13 +32,11 @@
...
-        "integ:local:firefox": "npm run build:webpack:dev && cross-env NODE_OPTIONS=--max-http-header-size=80000 testcafe firefox --app \"http-server ./build/dev -s\"",
-        "integ:local:firefox:headless": "npm run build:webpack:dev && cross-env NODE_OPTIONS=--max-http-header-size=80000 testcafe firefox:headless --app \"http-server ./build/dev -s\"",
-        "integ:local:safari": "npm run build:webpack:dev && cross-env NODE_OPTIONS=--max-http-header-size=80000 testcafe safari --app \"http-server ./build/dev -s\"",
-        "integ:local:edge": "npm run build:webpack:dev && cross-env NODE_OPTIONS=--max-http-header-size=80000 testcafe edge --app \"http-server ./build/dev -s\"",
-        "integ:local:ie": "npm run build:webpack:dev && cross-env NODE_OPTIONS=--max-http-header-size=80000 testcafe ie --app \"http-server ./build/dev -s\" src/__integ__/unsupportedBrowser.test.ts",
+        "preinteg:browser": "npm run build:webpack:dev",
+        "integ:browser": "testcafe --config-file=./.testcafe.browser.js --app \"http-server ./build/dev -s\"",
```
adebayor123, line 36: Is `NODE_OPTIONS=--max-http-header-size=80000` no longer required?

.testcafe.browser.js
maniator: The default is 3, and it was failing in Windows

package.json
maniator: It's an env variable added in the ci yaml file

No need to manipulate node options locally if not needed (otherwise can update local env)

maniator: that PR was forked from this one, but there is nothing preventing you from doing:

```npm run integ:browser -- chrome```

which will run tests in chrome :-)

Comment (maniator) [reply package.json:36]: We can add to README if youd like :-)

Comment (adebayor123) [reply package.json:36]: I think using configs is good practice, but we can still maintain individual commands for different browsers without having to use a new command:

```
"preinteg:browser": "npm run build:webpack:dev",
"preinteg:headless": "npm run build:webpack:dev",
"integ:browser": "testcafe --config-file=./.testcafe.browser.js --app \"http-server ./build/dev -s\"",
"integ:headless": "testcafe --config-file=./.testcafe.headless.js --app \"http-server ./build/dev -s\"",
"integ:local:chrome": "npm run integ:headless -- chrome",
"integ:local:firefox": "npm run integ:browser -- firefox",
"integ:local:firefox:headless": "npm run integ:headless -- firefox",
```

This still maintains the idea of using configs, and we reuse `preinteg` commands without having to add them into the existing commands for separate browsers. On the other hand, if we will be migrating into a single test framework in the near future, is it still worth to integrate change now?

What are your thoughts, qhanam@ and maniator@?

qhanam: If the `npm run integ:browser -- chrome` works that lgtm.

qhanam (approved):

package.json
maniator: agreed

adebayor123 (approved):

PR #153, opened 2022-05-11
test: Integrate automated smoke test into release workflow

adebayor123:
### Details
Currently as part of the release process, contributors in AWS RUM manually perform smoke tests to gain assurance on the newly-releasing web client version's behavior. This process has a couple of issues:
1. There are no predefined set of tests that contributors go through.
2. It is easy to miss out on regressions.

To mitigate this, this PR integrates automated smoke test as part of the release workflow.

### Changes
1. Embed `smoke.html` into repository, with all sensitive parameters anonymized and stored in GitHub secrets.
2. During the release workflow, the automated smoke test will run after a new version has been pushed to the CDN.
3. The smoke test will perform a set of defined test cases on the new version of the web client.
4. If the smoke test cases are successful, proceeds to NPM release. Otherwise, stops the release workflow.

### Testing
1. Successful Run: https://github.com/adebayor123/aws-rum-web/runs/6381491772?check_suite_focus=true
2. Unsuccessful Run: https://github.com/adebayor123/aws-rum-web/runs/6381538124?check_suite_focus=true
3. Final verification Run: https://github.com/adebayor123/aws-rum-web/runs/6381730545?check_suite_focus=true
### Setup
This is only applicable to AWS RUM members.

### Notes
When the test is run with 4 workers, all the ingested events were retrieved within a maximum of 2 minutes. However, when the tests are run with 6 workers, around 1-2 cases time out (5m) and the overall time to retrieve events is increased to 2-3 minutes.
AWS RUM team should investigate into this issue.

---


app/smoke.html
```diff
@@ -0,0 +1,221 @@
...
+                    x.q.push({ c: c, p: p });
+                };
+                z = document.createElement('script');
+                z.async = true;
+                z.src = s;
+                document.head.insertBefore(
+                    z,
+                    document.getElementsByTagName('script')[0]
+                );
+            })('cwr', $MONITOR_ID, '1.0.0', $REGION, $CDN, {
```
limhjgrace, line 29: Do we always want to use `1.0.0` for the web client version? Or should/can we do the smoke test against the most recent release?

playwright.config.ts
```diff
@@ -4,6 +4,8 @@ const { devices } = require('@playwright/test');
 
 const config = {
     forbidOnly: !!process.env.CI,
+    reporter: 'list',
+    workers: process.env.CI ? 6 : undefined,
```
limhjgrace, line 8: In the commit details, you said that using 6 workers causes 1-2 cases to time out. Do we want to set the number of workers to 6 then? Or should we set it to 4 workers until we find out why the overall time increases with more workers?

app/smoke.html
qhanam: This field is meant to store the application version. When we add a web client version field, it will be constant (i.e., not configurable).

.github/scripts/upload_smoke_test.sh
```diff
@@ -0,0 +1,2 @@
+bucket=$1
+aws s3api put-object --bucket $bucket --key "smoke.html" --body processed_smoke.html --content-type "text/html"
```
qhanam, line 2: question: Would it be difficult to version the smoke test file? E.g., `smoke-1.5.1.html`? This way we could (1) retain older versions for testing, and (2) it would be easy to verify which versions have been tested.

This isn't a blocker, but something to consider if it's easy to do.

qhanam (commented):
lgtm pending @limhjgrace question

playwright.config.ts
adebayor123: Very good point. I do think it's better to use 4 workers as (1) it completes the smoke test run faster (2) showing ingestion lag at this point doesn't provide any value for web client's E2E behavior.

I will be updating the PR with 4 workers instead.

.github/scripts/upload_smoke_test.sh
adebayor123: In order to do so, I think the changes need to happen on (1) CDK to edit the bucket with versioning characteristics (2) `upload_smoke_test.sh` and `cd.yaml`.

I think it's better to have the fix incorporated with this, since otherwise we will have to keep the enhancement in our backlog (which might get overlooked).

I will add the changes with the next revision.

playwright.config.ts
adebayor123: Changed the number of workers to 4

.github/scripts/upload_smoke_test.sh
adebayor123: Clarification: Bucket versioning is not needed - we expect the bucket to hold different versions of the smoke test, with the naming convention of `*.cloudfront.net/smoke-{version}.html`
Verified testing in gamma

adebayor123:
New Tests:
- local environment & no URL => Uses localhost
- local environment & custom URL with path => Uses custom URL
- local environment & custom URL without path => Uses `.com/smoke-{version}.html`
- Verified in gamma stage: https://github.com/adebayor123/aws-rum-web/runs/6481046195?check_suite_focus=true

PR #154, opened 2022-05-12
fix: Change pageIdFormat from enum to type for TypeScript compatibility

adebayor123:
### Details
When the `PageIdFormat` was defined as enum, TypeScript based applications were not able to compile due to two reasons:
1. when the string enum value was passed in, TypeScript does not allow string equivalent to enum types
2. when the enum value was passed in, application did not compile since the enum was not exported.

### This PR
To fix the issue, this PR changes `PAGE_ID_FORMAT` enum to `PageIdFormat` type, where the values are defined with strings. All the unit tests & implementations have been updated to use `PageIdFormat` type.

---


src/orchestration/Orchestration.ts
```diff
@@ -41,11 +41,7 @@ interface TelemetriesFunctor {
 
 type Telemetry = string | (string | object)[];
 
-export enum PAGE_ID_FORMAT {
-    PATH = 'PATH',
-    HASH = 'HASH',
-    PATH_AND_HASH = 'PATH_AND_HASH'
-}
```
qhanam, line 48: issue: This enum can still be used to reference the page ID formats so we don't have to write out the string literals. The issue is related to the type of `PartialConfig.pageIdFormat` and `Config.pageIdFormat`. So if you change the type of those values to `PageIdFormat` I think the problem is solved.

src/orchestration/Orchestration.ts
```diff
@@ -41,11 +41,7 @@ interface TelemetriesFunctor {
...
+export type pageIdFormat = 'PATH' | 'HASH' | 'PATH_AND_HASH';
```
qhanam, line 44: nit: Capitalize the type `PageIdFormat`

src/orchestration/Orchestration.ts
qhanam: observation: At some point we should export all the types used in `PartialConfig` from `index.ts`. This includes `PAGE_ID_FORMAT`.  This will allow the types to be used by the applications installing the web client.

We could do it in this PR, or create a separate PR for it later.

src/orchestration/Orchestration.ts
maniator: agreed

adebayor123: Fixed!

src/orchestration/Orchestration.ts
adebayor123: (1) Moved the enums in `Orchestration.ts` to `constants.ts` in the same directory
(2) Added 4 types (Telemetry, PageIdFormat, PartialCookieAttribute, and ClientBuilder) to `index.ts` to export.

src/orchestration/constants.ts
```diff
@@ -0,0 +1,12 @@
+export enum PAGE_ID_FORMAT {
```
adebayor123, line 1: qhanam@, do you think `src/orchestration/constants.ts` appropriate naming/place to hold the enums used in Orchestration and test cases?

src/index.ts
```diff
@@ -1,7 +1,11 @@
 export {
     PartialConfig as AwsRumConfig,
-    Orchestration as AwsRum
+    Orchestration as AwsRum,
+    PartialCookieAttributes as PartialCookieAttributes,
+    PageIdFormat as PageIdFormat,
+    Telemetry as Telemetry
```
qhanam, line 6: nit: We don't need to add an alias for names we aren't changing.

src/orchestration/constants.ts
qhanam: I think these should stay in Orchestration, as they are part of the configuration.

We can also export them from `index.ts` so they are available to consumers of the web client.

src/index.ts
```diff
@@ -1,7 +1,11 @@
...
 } from './orchestration/Orchestration';
```
maniator, line 7: can really even be 

```ts
export * from from './orchestration/Orchestration';
```

if we are not changing anything that is exported

src/index.ts
maniator, line 6:
```suggestion
    Orchestration as AwsRum,
    PartialCookieAttributes,
    PageIdFormat,
    Telemetry
```

No need for the `as ...` as the names are not changing

adebayor123: Makes sense!

src/index.ts
adebayor123: Unfortunately we are using alias for some exports :P

qhanam (approved):

PR #156, opened 2022-05-16
feat: Use data plane endpoint path prefix.

qhanam:
In cases where a proxy (e.g., using API Gateway) is used in between the web client and CloudWatch RUM, the proxy may contain a path prefix. For example, an API Gateway URL may contain a path prefix to disambiguate environments:
```
https://{restapi_id}.execute-api.{region}.amazonaws.com/{stage_name}/
```

This change allows the `endpoint` configuration option to contain a path. This path will be used as the prefix for sending data to CloudWatch RUM. RUM data sent to the API Gateway URL from the example above will then be:
```
https://{restapi_id}.execute-api.{region}.amazonaws.com/{stage_name}/appmonitors/application123
```

---


src/dispatch/DataPlaneClient.ts
```diff
@@ -65,20 +64,20 @@ export class DataPlaneClient {
     public sendFetch = async (
         putRumEventsRequest: PutRumEventsRequest
     ): Promise<{ response: HttpResponse }> => {
-        const host = getHost(this.config.endpoint);
         const serializedRequest: string = JSON.stringify(
             serializeRequest(putRumEventsRequest)
         );
+        const path = this.config.endpoint.pathname.replace(/\/$/, '');
```
maniator, line 70: consider getting these values from config on line 67 like so:

```ts
const { pathname, host, hostname, protocol } = this.config.endpoint
```

Makes the code following a bit cleaner

src/dispatch/DataPlaneClient.ts
```diff
@@ -94,20 +93,20 @@ export class DataPlaneClient {
     public sendBeacon = async (
         putRumEventsRequest: PutRumEventsRequest
     ): Promise<{ response: HttpResponse }> => {
-        const host = getHost(this.config.endpoint);
         const serializedRequest: string = JSON.stringify(
             serializeRequest(putRumEventsRequest)
         );
+        const path = this.config.endpoint.pathname.replace(/\/$/, '');
```
maniator, line 99: same as above comment

src/dispatch/__tests__/DataPlaneClient.test.ts
```diff
@@ -70,8 +70,8 @@ describe('DataPlaneClient tests', () => {
         await client.sendFetch(Utils.PUT_RUM_EVENTS_REQUEST);
 
         // Assert
-        // @ts-ignore
-        const signedRequest: HttpRequest = fetchHandler.mock.calls[0][0];
+        const signedRequest: HttpRequest = (fetchHandler.mock
```
maniator, line 73: can be:

```ts

const [[signedRequest: HttpRequest] = fetchHandler.mock.calls;

File: src/plugins/utils/http-utils.ts [hunk 4]
```diff
@@ -123,15 +122,20 @@ export const createXRaySubsegment = (
...
-        hostname = getHost((request as Request).url);
-    } else {
-        hostname = getHost(request.toString());
+    try {
+        if ((request as URL).hostname) {
+            return (request as URL).hostname;
+        } else if ((request as Request).url) {
+            return new URL((request as Request).url).hostname;
+        } else {
+            return new URL(request.toString()).hostname;
```
maniator, line 131: why the need for the `.toString`?

src/dispatch/__tests__/DataPlaneClient.test.ts
```diff
@@ -132,4 +132,107 @@ describe('DataPlaneClient tests', () => {
             'd37eb756444ebf6f785233714d6d942f2b20f69292fb09533f6b69556eb0ff2b'
         );
     });
+
+    test('when the endpoint contains a path then the fetch reqest url contains the path prefix', async () => {
+        // Init
+        const endpoint = new URL(
+            `${Utils.AWS_RUM_ENDPOINT.toString()}${'prod'}`
```
limhjgrace, line 139: nit: it looks like the endpoint URL object would look like `<first part of url>.comprod` - do we want a `/` in between the two place holders?

src/dispatch/DataPlaneClient.ts
adebayor123: Agreed

src/dispatch/__tests__/DataPlaneClient.test.ts
```diff
@@ -132,4 +132,107 @@ describe('DataPlaneClient tests', () => {
             'd37eb756444ebf6f785233714d6d942f2b20f69292fb09533f6b69556eb0ff2b'
         );
     });
+
+    test('when the endpoint contains a path then the fetch reqest url contains the path prefix', async () => {
```
adebayor123, line 136: nit: `reqest -> request`

src/dispatch/__tests__/DataPlaneClient.test.ts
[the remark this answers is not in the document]
adebayor123: I think `.toString()` is required to get the entire URL as string, with `prod` attached to it at the end.

src/dispatch/__tests__/DataPlaneClient.test.ts
```diff
@@ -132,4 +132,107 @@ describe('DataPlaneClient tests', () => {
...
+            credentials: Utils.createAwsCredentials()
+        });
+
+        // Run
+        await client.sendFetch(Utils.PUT_RUM_EVENTS_REQUEST);
+
+        // Assert
+        const signedRequest: HttpRequest = (fetchHandler.mock
+            .calls[0] as any)[0];
+        expect(signedRequest.hostname).toEqual(Utils.AWS_RUM_ENDPOINT.hostname);
```
adebayor123, line 155: question: Could we add examples for specific parameters, such as hostname, host, path, etc as comments?

src/dispatch/__tests__/DataPlaneClient.test.ts
```diff
@@ -132,4 +132,107 @@ describe('DataPlaneClient tests', () => {
...
+        expect(signedRequest.path).toEqual(
+            `${endpoint.pathname}/appmonitors/application123/`
+        );
+    });
+
+    test('when the endpoint path contains a trailing slash then the fetch reqest url drops the trailing slash', async () => {
```
adebayor123, line 161: nit: `reqest -> request`

src/dispatch/__tests__/DataPlaneClient.test.ts
adebayor123: I think the test is trying to capture that we using `URL.hostname` only extracts up to `<first part of url>.com` based on the rest of the test - `Utils.AWS_RUM_ENDPOINT`'s hostname is `rumservicelambda.us-west-2.amazonaws.com`, and we are checking even if it's `.comprod`, the `signedRequest`'s hostname does not include `prod`

src/dispatch/__tests__/DataPlaneClient.test.ts
```diff
@@ -132,4 +132,107 @@ describe('DataPlaneClient tests', () => {
...
+
+        // Run
+        await client.sendFetch(Utils.PUT_RUM_EVENTS_REQUEST);
+
+        // Assert
+        const signedRequest: HttpRequest = (fetchHandler.mock
+            .calls[0] as any)[0];
+        expect(signedRequest.hostname).toEqual(Utils.AWS_RUM_ENDPOINT.hostname);
+        expect(signedRequest.path).toEqual(
+            `${endpoint.pathname.replace(
```
adebayor123, line 182: question: Seems that we will use this similar regex pattern to parse the part of the `pathname` in the main code - Could we put that into a helper method and place it in util, then (1) reuse in test cases (2) create a unit test for it?

src/dispatch/__tests__/DataPlaneClient.test.ts
```diff
@@ -132,4 +132,107 @@ describe('DataPlaneClient tests', () => {
...
+            )}/appmonitors/application123/`
+        );
+    });
+
+    test('when the endpoint contains a path then the beacon reqest url contains the path prefix', async () => {
+        // Init
+        const endpoint = new URL(
+            `${Utils.AWS_RUM_ENDPOINT.toString()}${'prod'}`
+        );
+        const client: DataPlaneClient = new DataPlaneClient({
```
adebayor123, line 194: nit: For all the cases we are creating a `DataPlaneClient`, we can extract this into a method called `createDPClient` at the start of the file, and reuse in test cases - this would reduce the file length.

src/orchestration/Orchestration.ts
```diff
@@ -47,6 +46,8 @@ export enum PAGE_ID_FORMAT {
     PATH_AND_HASH = 'PATH_AND_HASH'
 }
 
+export type pageIdFormat = 'PATH' | 'HASH' | 'PATH_AND_HASH';
```
adebayor123, line 49: nit: `pageIdFormat -> PageIdFormat` :)

src/orchestration/Orchestration.ts
```diff
@@ -213,6 +233,13 @@ export class Orchestration {
 
         this.config.endpoint = this.getDataPlaneEndpoint(region, partialConfig);
 
+        // If the URL is not formatted correctly, a TypeError will be thrown.
```
adebayor123, line 236: question: Would it be worth adding a unit test case to verify URLs that is expected to fail do fail? (even though URL is not our code).

src/dispatch/__tests__/DataPlaneClient.test.ts
adebayor123: maniator@, does this automatically index for us?

src/dispatch/__tests__/DataPlaneClient.test.ts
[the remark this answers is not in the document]
adebayor123: On top, it adds a trailing slash to complete URL

adebayor123: `const test = new URL(
            `${endpoint.toString()}${'prod'}`)
console.log('test ', test)` leads to `"test ",  URL: "https://rumservicelambda.us-west-2.amazonaws.com/prod" `

src/dispatch/__tests__/DataPlaneClient.test.ts
maniator: this would get the first index value of the first index value of the array of calls.

Can test this out in dev tools by doing the following:

```js
const [ [ signedRequest ] ] = [['req', 12], [1,2,3]];
console.log(signedRequest);
```

image

src/dispatch/__tests__/DataPlaneClient.test.ts
maniator: related to https://github.com/aws-observability/aws-rum-web/pull/156#discussion_r874838260

src/plugins/utils/http-utils.ts
qhanam: TypeScript likes the type to be explicit.

src/dispatch/__tests__/DataPlaneClient.test.ts
adebayor123: Never knew that!

maniator: The more you know :-) 

Same as object destructuring here: https://github.com/aws-observability/aws-rum-web/pull/156#discussion_r873865794 but with an array :-)

src/plugins/utils/http-utils.ts
maniator: can do:

```ts
             return new URL(<String>request).hostname;
```

but I guess that is a bit hacky :-|

src/dispatch/__tests__/DataPlaneClient.test.ts
qhanam: The test is verifying that the trailing slash is dropped. Re-using the same code would defeat the purpose of the test, as it would not detect a regression in that function.

src/dispatch/__tests__/DataPlaneClient.test.ts
qhanam: What value would this add? The test is checking one property of the code.

src/dispatch/__tests__/DataPlaneClient.test.ts
qhanam: I'm not a big fan of destructuring. In many cases its additional layer of abstraction makes the code less clear (especially for those new to JS/TS). E.g., this would become:
```
const [[signedRequest]]: [[signedRequest: HttpRequest]] = fetchHandler.mock.calls as any;
```

src/dispatch/DataPlaneClient.ts
qhanam: I'm not a big fan of destructuring. In many cases its additional layer of abstraction makes the code less clear (especially for those new to JS/TS). E.g., is the value from the config, or from a local variable? I guess it saves some bits, but there is lower hanging fruit if we're worried about size.

src/dispatch/__tests__/DataPlaneClient.test.ts
qhanam: There is a tradeoff here. Verbosity gives us clarity into what the test's inputs are. While we could extract method for each instantiation, we would need to navigate to the instantiation to see the configuration for the test.

For test suites, file length is not a problem since each test is a self-contained specification of the program's behavior. (However, having long test methods IS a code smell.)

src/dispatch/__tests__/DataPlaneClient.test.ts
```diff
@@ -132,4 +132,107 @@ describe('DataPlaneClient tests', () => {
...
+        expect(signedRequest.hostname).toEqual(Utils.AWS_RUM_ENDPOINT.hostname);
+        expect(signedRequest.path).toEqual(
+            `${endpoint.pathname.replace(
+                /\/$/,
+                ''
+            )}/appmonitors/application123/`
+        );
+    });
+
+    test('when the endpoint contains a path then the beacon reqest url contains the path prefix', async () => {
```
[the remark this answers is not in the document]
qhanam: Fixed

src/plugins/utils/http-utils.ts
```diff
@@ -123,15 +122,20 @@ export const createXRaySubsegment = (
...
+        }
+    } catch (e) {
+        // The URL could not be parsed. This library's convention is to fail
+        // silently to limit the risk of impacting the application being
+        // monitored.  We will use the hosname of the current page instead.
```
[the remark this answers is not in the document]
qhanam: Fixed

src/dispatch/__tests__/DataPlaneClient.test.ts
qhanam: Fixed

src/dispatch/__tests__/DataPlaneClient.test.ts
qhanam: Fixed

src/orchestration/Orchestration.ts
qhanam: Fixed

src/dispatch/__tests__/DataPlaneClient.test.ts
adebayor123: Perhaps the better place might be in the dataplane client code... The idea was to show how we parse a URL into `hostname`, `host`, `path`, to help contributors understand how we create HTTP requests - just a thought

src/dispatch/__tests__/DataPlaneClient.test.ts
adebayor123: Good point

src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -134,10 +133,19 @@ export class ResourcePlugin implements Plugin {
...
+        try {
+            // Ignore monitoring beacons.
+            if (
+                new URL(entryData.name).host ===
+                this.context.config.endpointUrl.host
+            ) {
+                return;
+            }
+        } catch (e) {
+            // Fail-safe ignore
```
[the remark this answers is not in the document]
adebayor123: I don't think we can, because the web client resides in customer applications - one day if we have a way to collect error metrics of web client for ourselves, we could add it here

adebayor123 (approved):

src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -134,10 +133,19 @@ export class ResourcePlugin implements Plugin {
...
-            getHost(entryData.name) !== getHost(this.context.config.endpoint)
-        ) {
+        try {
+            // Ignore monitoring beacons.
+            if (
+                new URL(entryData.name).host ===
+                this.context.config.endpointUrl.host
+            ) {
+                return;
+            }
```
maniator, line 143: I don't see how this would ever throw an error (and need a try/catch block)

src/dispatch/DataPlaneClient.ts
maniator: In my mind it is easier to parse -- I agree w you on the array destructuring, but for objects, I still maintain that it makes sense.

src/dispatch/__tests__/DataPlaneClient.test.ts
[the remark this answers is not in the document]
qhanam: Fixed.

src/plugins/event-plugins/ResourcePlugin.ts
qhanam: Agree -- fixed

src/dispatch/__tests__/DataPlaneClient.test.ts
qhanam: Added a line to the user docs about the feature.

URL parsing is standardized/defined by W3C -- should be clear.

PR #158, opened 2022-05-17
chore: 🤖 set node version for nvm

maniator:
Set version of node used by nvm so that all devs of project stay on the same version (so that package-lock stays consistent etc)

---


adebayor123:
qhanam@ I think this is a good addition (remembering I had node version issues myself), what do you think?

PR #159, opened 2022-05-17
refactor: 💡 Update typings/interface for Plugins

maniator:
Make `InternalPlugin` an abstract class so that certain values can be set by default along with defining a constructor.

---


src/plugins/Plugin.ts
```diff
@@ -15,28 +16,34 @@ export type PluginContext = {
     getSession: GetSession;
 };
 
-export interface Plugin {
+export abstract class Plugin {
+    protected enabled: boolean = true;
+
+    constructor(
+        public readonly pluginName: string,
+        private readonly pluginId = `${RUM_AWS_PREFIX}.${pluginName}`
```
qhanam, line 24: issue: A third party may also implement a plugin.

src/plugins/Plugin.ts
```diff
@@ -15,28 +16,34 @@ export type PluginContext = {
     getSession: GetSession;
 };
 
-export interface Plugin {
+export abstract class Plugin {
+    protected enabled: boolean = true;
+
+    constructor(
+        public readonly pluginName: string,
```
qhanam, line 23: question: Do we need both `pluginName` and `pluginId`?

src/sessions/VirtualPageLoadTimer.ts
```diff
@@ -63,6 +64,10 @@ export class VirtualPageLoadTimer extends MonkeyPatched {
         document.addEventListener('keydown', this.updateLatestInteractionTime);
     }
 
+    load?(context: PluginContext): void {
+        throw new Error('Method not implemented.');
```
qhanam, line 68: nit: I'd prefer if this function did nothing rather than throwing an error to follow a fail-safe convention.

maniator: Yup -- my ide created this fn (as it said the abstract method was not implemented)

I can make it empty

src/plugins/Plugin.ts
maniator: Hmmmm maybe can make the prefix optional in the abstract?

Would that make sense? or make it so that it can be changed in the inherited class?

maniator: but also, the pluginId can be passed as the 2nd param to the constructor (then this default value wont be used :-) )

src/plugins/Plugin.ts
maniator: I think so, especially if we get to a point where ids could get a bit abstract

qhanam: Semantically, what is the difference between the two? When would we use one over the other?

It seems like a single ID is sufficient for `PluginManager` to uniquely identify the plugin, which I believe is the only place it is used.

maniator: well in one of my commits, I did rename `pluginId` to `awsRumId` or something like that -- I think (personally) that the aws id should not be the plugin identifier in the code, and we should have a cleaner "id" (here named `pluginName`) to identify the plugin in code.

src/plugins/Plugin.ts
qhanam: If we use an abstract class we would be enforcing implementation details, rather than simply defining behavior. This is ok for the built-in plugins, which we can change, but I don't think we should assume anything about how plugins will be implemented outside of this package.

For example, I may want to dynamically create my plugin's ID.

maniator: and you still can


```ts

class BananaPlugin extends Plugin {
    constructor() {
        super('banana', someIdCreationMethod());
        // ...
    }

   // ...
}
```

Which will create a `BananaPlugin` with a name of `banana` and an id defined by them in the `someIdCreationMethod` method (or however they want to create it

maniator: Unless I am misunderstanding you.

qhanam: Backwards compatibility is another issue. Existing third-party plugins could break.

Adding an abstract class on top of the interface will also give you more flexibility to make changes to the abstract class in the future, while if you replace the interface you will be limited in what you can do because of issues backwards compatibility.

Let's keep this as an interface.

src/plugins/Plugin.ts
qhanam: I don't see any use case for `pluginName`. Let's stick with `pluginId` only.

src/plugins/Plugin.ts
maniator: kk can do 

Will make an abstract class just for the internal plugins for now :-)

maniator:
@qhanam @adebayor123 

Let me know if you have any more questions

src/plugins/Plugin.ts
```diff
@@ -42,11 +27,11 @@ export interface Plugin {
...
-    record?(data: any): void;
+    record?<D extends unknown>(data: D): void;
 
     /**
      * Update the plugin.
-     * @param config Data that the plugin will use to update its config.
+     * @param updateWith Data that the plugin will use to update its config.
      */
-    update?(config: object): void;
+    update?<O extends unknown>(updateWith: O[]): void;
```
qhanam, line 36: issue: The array is specific to the DomEventPlugin -- the intent was that this could be any type. I suggest leaving this type unknown.
```suggestion
    update?<O extends unknown>(updateWith: O): void;
```

qhanam (approved):
One suggestion -- otherwise lgtm!

src/plugins/Plugin.ts
maniator: Sure, would just need to update the implementation of it for the DOM plugin call.

maniator:
@qhanam all up to date and passing :-)

qhanam:
Very cool -- thank you!

PR #164, opened 2022-05-22
Ignore errors

### LINKED ISSUE #119: Ignore errors with matching pattern configuration please.
Expecting:

```
{
          sessionSampleRate: 1,
          guestRoleArn:
            'arn:aws:iam::xxxxxxxx',
          identityPoolId: 'us-west-xxxxxxxxx',
          endpoint: 'https://dataplane.rum.us-west-2.amazonaws.com',
          telemetries: ['performance', 'errors', 'http'],
          allowCookies: true,
          enableXRay: true,
          pageIdFormat: 'PATH_AND_HASH',
/////// if the following configuration matches an error then ignore it.
          ignoreErrors: [
              /ResizeObserver loop limit exceeded/g,  // <---------- Focus here please.
              /Script error\./g,  // <---------- Focus here please.
          ]
},
```

adebayor123:
## Details
Previously, CW RUM did not provide an option to allow customers to filter out error events that provided no monitoring value. 
This PR addresses the issue by allowing the customers to pass in a function that acts similar to `Array.filter` to indicate whether the error should be recorded or not.

Resolves #119 

## Testing
Added a new unit test and an integ test to cover the new behavior.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -60,7 +60,16 @@ telemetries: [ 'errors', 'performance', 'http' ]
 ```
 ```javascript
 telemetries: [ 
-    [ 'errors', { stackTraceLength: 500 } ], 
+    [ 'errors', { 
+        stackTraceLength: 500, 
+        filter: (errorEvent) => {
+            const patternsToIgnore = [/ResizeObserver loop/, /undefined/];
+            return (
+                patternsToIgnore.filter((pattern) =>
+                    patternsToIgnore.test(errorEvent.message)
+                ).length !== 0
+            );
+        }} ], 
     'performance',
     [ 'http', { stackTraceLength: 500, addXRayTraceIdHeader: true } ]
 ]
@@ -78,6 +87,8 @@ telemetries: [
 | Name | Type | Default | Description |
 | --- | --- | --- | --- |
 | stackTraceLength | Number | `200` | The number of characters to record from a JavaScript error's stack trace (if available). |
+| filter | Function | `() => false` | By default, the web client will record all errors. If you wish to filter out certain errors, pass in a function that will return true if the error should be ignored, similar to the example above. |
+
 
 ## HTTP
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -59,17 +59,8 @@ For example, the following telemetry config arrays are both valid. The one on th
 telemetries: [ 'errors', 'performance', 'http' ]
 ```
 ```javascript
-telemetries: [ 
-    [ 'errors', { 
-        stackTraceLength: 500, 
-        filter: (errorEvent) => {
-            const patternsToIgnore = [/ResizeObserver loop/, /undefined/];
-            return (
-                patternsToIgnore.filter((pattern) =>
-                    patternsToIgnore.test(errorEvent.message)
-                ).length !== 0
-            );
-        }} ], 
+telemetries: [
+    [ 'errors', { stackTraceLength: 500 } ],
     'performance',
     [ 'http', { stackTraceLength: 500, addXRayTraceIdHeader: true } ]
 ]
@@ -87,8 +78,29 @@ telemetries: [
 | Name | Type | Default | Description |
 | --- | --- | --- | --- |
 | stackTraceLength | Number | `200` | The number of characters to record from a JavaScript error's stack trace (if available). |
-| filter | Function | `() => false` | By default, the web client will record all errors. If you wish to filter out certain errors, pass in a function that will return true if the error should be ignored, similar to the example above. |
+| ignore | Function | `() => false` | A function which accepts an [`ErrorEvent`](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent) or a [`PromiseRejectionEvent`](https://developer.mozilla.org/en-US/docs/Web/API/PromiseRejectionEvent) and returns a value that coerces to true when the error should be ignored. By default, no errors are ignored. |
 
+For example, the following telemetry config array causes the web client to ignore all errors whose message begins with "Warning:".
+
+```javascript
+telemetries: [
+    [
+        'errors',
+        {
+            stackTraceLength: 500,
+            ignore: (errorEvent) => {
+                return (
+                    errorEvent &&
+                    errorEvent.message &&
+                    errorEvent.message.test(/^Warning:/)
+                );
+            }
+        }
+    ],
+    'performance',
+    'http'
+]
+```
 
 ## HTTP
 
--- a/src/loader/loader-js-error-event.js
+++ b/src/loader/loader-js-error-event.js
@@ -5,7 +5,18 @@ loader('cwr', 'abc123', '1.0', 'us-west-2', './rum_javascript_telemetry.js', {
     allowCookies: true,
     dispatchInterval: 0,
     metaDataPluginsToLoad: [],
-    eventPluginsToLoad: [new JsErrorPlugin()],
+    eventPluginsToLoad: [
+        new JsErrorPlugin({
+            filter: (errorEvent) => {
+                const patterns = [/ResizeObserver loop/];
+                return (
+                    patterns.filter((pattern) =>
+                        pattern.test(errorEvent.message)
+                    ).length !== 0
+                );
+            }
+        })
+    ],
     telemetries: [],
     clientBuilder: showRequestClientBuilder
 });
--- a/src/loader/loader-js-error-event.js
+++ b/src/loader/loader-js-error-event.js
@@ -7,7 +7,7 @@ loader('cwr', 'abc123', '1.0', 'us-west-2', './rum_javascript_telemetry.js', {
     metaDataPluginsToLoad: [],
     eventPluginsToLoad: [
         new JsErrorPlugin({
-            filter: (errorEvent) => {
+            ignore: (errorEvent) => {
                 const patterns = [/ResizeObserver loop/];
                 return (
                     patterns.filter((pattern) =>
--- a/src/plugins/event-plugins/JsErrorPlugin.ts
+++ b/src/plugins/event-plugins/JsErrorPlugin.ts
@@ -6,17 +6,17 @@ export const JS_ERROR_EVENT_PLUGIN_ID = 'js-error';
 
 export type PartialJsErrorPluginConfig = {
     stackTraceLength?: number;
-    filter?: (error: ErrorEvent | PromiseRejectionEvent) => boolean;
+    ignore?: (error: ErrorEvent | PromiseRejectionEvent) => boolean;
 };
 
 export type JsErrorPluginConfig = {
     stackTraceLength: number;
-    filter: (error: ErrorEvent | PromiseRejectionEvent) => boolean;
+    ignore: (error: ErrorEvent | PromiseRejectionEvent) => boolean;
 };
 
 const defaultConfig: JsErrorPluginConfig = {
     stackTraceLength: 200,
-    filter: () => false
+    ignore: () => false
 };
 
 export class JsErrorPlugin extends InternalPlugin {
@@ -56,13 +56,13 @@ export class JsErrorPlugin extends InternalPlugin {
     }
 
     private eventHandler = (errorEvent: ErrorEvent) => {
-        if (!this.config.filter(errorEvent)) {
+        if (!this.config.ignore(errorEvent)) {
             this.recordJsErrorEvent(errorEvent);
         }
     };
 
     private promiseRejectEventHandler = (event: PromiseRejectionEvent) => {
-        if (!this.config.filter(event)) {
+        if (!this.config.ignore(event)) {
             this.recordJsErrorEvent({
                 type: event.type,
                 error: event.reason
--- a/src/plugins/event-plugins/JsErrorPlugin.ts
+++ b/src/plugins/event-plugins/JsErrorPlugin.ts
@@ -6,14 +6,17 @@ export const JS_ERROR_EVENT_PLUGIN_ID = 'js-error';
 
 export type PartialJsErrorPluginConfig = {
     stackTraceLength?: number;
+    filter?: (error: ErrorEvent | PromiseRejectionEvent) => boolean;
 };
 
 export type JsErrorPluginConfig = {
     stackTraceLength: number;
+    filter: (error: ErrorEvent | PromiseRejectionEvent) => boolean;
 };
 
 const defaultConfig: JsErrorPluginConfig = {
-    stackTraceLength: 200
+    stackTraceLength: 200,
+    filter: () => false
 };
 
 export class JsErrorPlugin extends InternalPlugin {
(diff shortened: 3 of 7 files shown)


docs/configuration.md
```diff
@@ -78,6 +78,8 @@ telemetries: [
 | Name | Type | Default | Description |
 | --- | --- | --- | --- |
 | stackTraceLength | Number | `200` | The number of characters to record from a JavaScript error's stack trace (if available). |
+| errorsToExclude | RegExp[] | `[]` | A list of error messages that will be used to identify errors to ignore. Matched errors will not be recorded. |
```
adebayor123, line 81: Initially thought of accepting both string and regex, but seems like the pattern we have is accepting only regex input (whether intentional or unintentional). Decided to follow what we have.

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -6,26 +6,31 @@ export const JS_ERROR_EVENT_PLUGIN_ID = 'com.amazonaws.rum.js-error';
...
-    stackTraceLength: 200
+    stackTraceLength: 200,
+    errorsToExclude: []
 };
 
 export class JsErrorPlugin implements Plugin {
     private pluginId: string;
     private enabled: boolean;
     private config: JsErrorPluginConfig;
+    private ignoredErrors: Set<string>;
```
adebayor123, line 26: If the customer has many regex inputs in `errorsToExclude`, I believed we could cache the errors that were ignored already and simply speed up the checking without having to go through the entire list at the expense of using space.

After doing some research, it seems that iterating through hundreds of regex patterns seems to be still very fast, so this may not be necessary at all. What do you think?

docs/configuration.md
adebayor123: One thing that this PR doesn't contain is live update of list of errors to ignore. I personally don't think it would be a useful add-on to ignore errors, so I didn't add it to this scope.

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -90,4 +98,17 @@ export class JsErrorPlugin implements Plugin {
             this.promiseRejectEventHandler
         );
     }
+
+    private isErrorInIgnoreList(message: string): boolean {
+        if (this.ignoredErrors.has(message)) {
```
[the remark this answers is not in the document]
adebayor123: Yeah, as mentioned in the previous comment above, the idea was to avoid going through a large list of regex that the customer passes in, but I do think this may not be needed since I've found some posts that it takes up to couple thousand inputs to reach 1 second of process time. Also great point, I personally have not thought of that case (and that is the biggest point proving we should take it out)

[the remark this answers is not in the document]
maniator: Sounds good 👍

docs/configuration.md
adebayor123: When my initial prototyping had both string and regex, I had to convert the string inputs to `RegExp` objects to use the matching methods (if this is what you're referring to!)

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -6,14 +6,17 @@ export const JS_ERROR_EVENT_PLUGIN_ID = 'com.amazonaws.rum.js-error';
 
 export type PartialJsErrorPluginConfig = {
     stackTraceLength?: number;
+    errorsToExclude?: RegExp[];
```
qhanam, line 9: issue: A callback function would give application maintainers more control over which errors should be included/excluded. What if we make this a function whose behavior is similar to [Array.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)?

Specifically, the function would take as input an error and return `true` if the error should be *included* in the RUM events.

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -62,10 +65,13 @@ export class JsErrorPlugin implements Plugin {
     }
 
     private eventHandler = (errorEvent: ErrorEvent) => {
-        this.recordEvent(
-            JS_ERROR_EVENT_TYPE,
-            errorEventToJsErrorEvent(errorEvent, this.config.stackTraceLength)
+        const rumEvent = errorEventToJsErrorEvent(
+            errorEvent,
+            this.config.stackTraceLength
```
qhanam, line 70: question: Assuming we use a function filter, rather than a regex filter, what should the input to the `fitler` function be?

Specifically, do we want the `filter` parameter to be (1) the input to `eventHandler`, `promiseRejectEventHandler`, or `record`, or (2) the output of `errorEventToJsErrorEvent`?

Having the raw data (from eventHandler, promiseRejectEventHandler and record) would provide the most flexibility, although the `filter` function would potentially need to handle the most types.

I am in favor of forwarding the original error, to give devs the most flexibility to implement filters. We could mitigate some of the complexity by providing a sample implementation in the documentation.

AntiMoron: Hi, I wasn't expecting so many details in a code review. Great job!!!  To your question, I have some input: `forwarding original error event` would just work. Some non-standard implementation should be considered in a monitoring product because this is unpredictable if one put some extra data into an `Error` object or programmer throw string\number literal for some unknown reasons. By providing raw data, let the user makes decisions how to deal with an error captured.

src/plugins/event-plugins/JsErrorPlugin.ts
AntiMoron: Vote up for this one. BTW I also wondered some other topics: why supporting `string` is not needed? In my opinion, most web apps are running in different browser environments which means some native error messages from browsers will be in different formats.

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -90,4 +96,15 @@ export class JsErrorPlugin implements Plugin {
             this.promiseRejectEventHandler
         );
     }
+
+    private isErrorInIgnoreList(message: string): boolean {
+        for (const pattern of this.config.errorsToExclude) {
```
AntiMoron, line 101: I have to mention that you've defined `errorsToExclude?: RegExp[];` as an option property, which means you should make an early exit if errorsToExclude is not provided. Although `defaultConfig` provided one array literal but that still not making sense in semantics.

AntiMoron (commented):
**Great job**

src/plugins/event-plugins/JsErrorPlugin.ts
adebayor123: Great suggestion, I have updated the PR so that customers will pass in a "filter" like function that will `true` when the error should be ignored (thought this is more intuitive - more than welcome to change if anyone disagrees).

src/plugins/event-plugins/JsErrorPlugin.ts
adebayor123: Again, great points - I do agree that if we are letting the customers decide which errors to filter out, we should pass in the untouched `ErrorEvent` object.

src/plugins/event-plugins/JsErrorPlugin.ts
adebayor123: This is a good point! Hopefully this is no longer an issue :)

docs/configuration.md
```diff
@@ -60,7 +60,16 @@ telemetries: [ 'errors', 'performance', 'http' ]
 ```
 ```javascript
 telemetries: [ 
-    [ 'errors', { stackTraceLength: 500 } ], 
+    [ 'errors', { 
+        stackTraceLength: 500, 
+        shouldIgnoreError: (errorEvent) => {
```
adebayor123, line 65: I think based on the param name `shouldIgnoreError` and the simple function should be sufficient for how to use the feature - thoughts?

src/loader/loader-js-error-event.js
```diff
@@ -5,7 +5,18 @@ loader('cwr', 'abc123', '1.0', 'us-west-2', './rum_javascript_telemetry.js', {
     allowCookies: true,
     dispatchInterval: 0,
     metaDataPluginsToLoad: [],
-    eventPluginsToLoad: [new JsErrorPlugin()],
+    eventPluginsToLoad: [
+        new JsErrorPlugin({
+            shouldIgnoreError: (errorEvent) => {
+                const patterns = [/ResizeObserver loop/];
```
maniator, line 11: Why are we passing this default here?

Shouldn't this be in the plugin itself?

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -6,14 +6,17 @@ export const JS_ERROR_EVENT_PLUGIN_ID = 'com.amazonaws.rum.js-error';
...
 
 export type JsErrorPluginConfig = {
     stackTraceLength: number;
+    shouldIgnoreError?: (error: ErrorEvent) => boolean;
 };
 
 const defaultConfig: JsErrorPluginConfig = {
-    stackTraceLength: 200
+    stackTraceLength: 200,
+    shouldIgnoreError: () => false
```
maniator, line 19: Shouldn't need a default as it is not required

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -62,10 +65,15 @@ export class JsErrorPlugin implements Plugin {
     }
 
     private eventHandler = (errorEvent: ErrorEvent) => {
-        this.recordEvent(
-            JS_ERROR_EVENT_TYPE,
-            errorEventToJsErrorEvent(errorEvent, this.config.stackTraceLength)
-        );
+        if (!this.config.shouldIgnoreError(errorEvent)) {
```
maniator, line 68: Isn't 'shouldIgnoreError' optional?

maniator, line 68: Also doing an early exit instead of wrapping in an if statement may be better here

src/plugins/event-plugins/__tests__/JsErrorPlugin.test.ts
```diff
@@ -521,4 +521,37 @@ describe('JsErrorPlugin tests', () => {
...
+            columnNumber: 2
+        });
+        // This creates an error event with 'undefined' as message
+        plugin.record({});
+        plugin.record({});
+        plugin.record({});
+        plugin.disable();
+
+        // Assert
+        expect(record).toHaveBeenCalledTimes(0);
```
maniator, line 555: Also should have a test to make sure it is called a certain amount of times with a mix of errors

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -6,14 +6,17 @@ export const JS_ERROR_EVENT_PLUGIN_ID = 'com.amazonaws.rum.js-error';
 
 export type PartialJsErrorPluginConfig = {
     stackTraceLength?: number;
+    shouldIgnoreError?: (error: ErrorEvent) => boolean;
 };
 
 export type JsErrorPluginConfig = {
     stackTraceLength: number;
+    shouldIgnoreError?: (error: ErrorEvent) => boolean;
```
qhanam, line 14:
```suggestion
    filter: (error: ErrorEvent | PromiseRejectionEvent) => boolean;
```

src/plugins/event-plugins/JsErrorPlugin.ts
```diff
@@ -62,10 +65,15 @@ export class JsErrorPlugin implements Plugin {
     }
 
     private eventHandler = (errorEvent: ErrorEvent) => {
```
qhanam, line 67: issue: I think we want to add the filter to `promiseRejectEventHandler` separately, so that the filter will get the complete event in both cases. We'll also need to avoid calling `eventHandler` from `promiseRejectEventHandler` so the filter doesn't get applied twice (or something to that effect).

src/plugins/event-plugins/JsErrorPlugin.ts
qhanam: From the MDN docs on the `filter` function's callback:
> `callbackFn`
> Function is a predicate, to test each element of the array. Return a value that coerces to true to keep the element, or to false otherwise.
In our case, the `filter` function will be the callback function, but I think the intent will be clear.

src/plugins/event-plugins/JsErrorPlugin.ts
qhanam: Similarly, for `record`, we don't want the error to pass through the filter.

I suggest doing an extract method refactor on `eventHandler` so that the three functions still share common code, but, the original `eventHandler` only does the `filter` check.

src/loader/loader-js-error-event.js
adebayor123: Not sure what you mean here - can you clarify a bit more?

maniator: Where is this file used?

is it a test/fixture file?

if so, should be noted in the filename

src/plugins/event-plugins/JsErrorPlugin.ts
adebayor123: It's not clear to me as to why default values are not required for optional parameters. Is it anti-pattern? I thought it makes sense to have a default value in case the user does not want to pass in anything (seems to be what we currently have in Orchestration as well).

src/plugins/event-plugins/JsErrorPlugin.ts
adebayor123: So I have to admit, I had a wrong understanding of PartialConfigs and default values.. Now it makes sense that (1) default values should exist for non-optional parameters or else the code will break, (2) if a parameter is optional, the code should not rely on it.

src/loader/loader-js-error-event.js
adebayor123: Loader files are used to load the web client in our test applications - we have different loader files for different test applications, as it helps us isolate the functionalities and behaviors we want to test.

I think the naming convention we have indicates that decently

src/plugins/event-plugins/JsErrorPlugin.ts
adebayor123: Made the following changes:
1. Pull out the actual `event` creating activity into `recordJsErrorEvent(error: any)` method. This method simply records the error event, and is used across `record(), promiseRejectEventHandler(), eventHandler()`.
2. `record()` will directly invoke `recordJsErrorEvent` instead of going through `eventHandler` to avoid going through filtering logic.
3. `promiseRejectEventHandler` will now do its own filtering and call `recordJsErrorEvent` if the error should be recorded

src/plugins/event-plugins/JsErrorPlugin.ts
adebayor123: are you suggesting something in the lines of this?:
```
private promiseRejectEventHandler = (event: PromiseRejectionEvent) => {
        if (this.config.filter(event)) {
            return;
        }
        this.recordJsErrorEvent({
            type: event.type,
            error: event.reason
        } as ErrorEvent);
    };
```

qhanam (approved):

PR #166, opened 2022-05-25
refactor: 💡 update how configurations are created

maniator:
Created the `Configurable` abstract class since it seems there is some repeat code when it comes to defining configuration

---

Looks longer than what it is...

a lot of it was replacing things like `PartialConfigN` with `Partial<ConfigN>`

----

TODO in a followup --

remove the `getFullConfig` method from the root class and figure out how _not to_ pass a mutable config to everything (I tried having `getFullConfig` return a shallow copy of `this.config` but everything broke)

Right now everything in the root `Orchestration` config is mutable, and _is_ being changed in tests and other plugins, which we probably **do not want**

---

Notes: 

all objects are mutable in javascript -- not many people know that
and all objects are passed by reference (no matter what)
So if you change the object in one place, it will change everywhere -- hence the need to @ least shallow copy the config object (or maybe even deep copy) (in a future update)

---


src/orchestration/Orchestration.ts
```diff
@@ -359,26 +323,29 @@ export class Orchestration {
                 id: applicationId,
                 version: applicationVersion
             },
-            this.config
+            this.getFullConfig()
```
maniator, line 326: what needs to be fixed in a followup :-(

If I try making a shallow copy of the orchestration config, everything breaks (which really should not happen)

maniator:
@qhanam @adebayor123 

Can you review this?

maniator:
Looks longer than what it is...

a lot of it was replacing things like `PartialConfigN` with `Partial<ConfigN>`

src/abstract/Configurable.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public getConfigValue<K extends keyof ConfigType>(
+        key: K,
+        defaultValue?: ConfigType[K]
+    ): ConfigType[K] {
+        return this.config[key] ?? defaultValue;
+    }
+
+    protected getFullConfig() {
```
maniator, line 25: This is the method that we can _hopefully_ remove in the future

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -12,19 +12,12 @@ import {
...
-    }
-
     // tslint:disable-next-line:no-empty
     enable(): void {}
 
     // tslint:disable-next-line:no-empty
     disable(): void {}
 
-    // tslint:disable-next-line:no-empty
-    configure(config: any): void {}
```
maniator, line 26: Never used

src/dispatch/Authentication.ts
```diff
@@ -4,16 +4,16 @@ import { Credentials } from '@aws-sdk/types';
 import { FetchHttpHandler } from '@aws-sdk/fetch-http-handler';
 import { StsClient } from './StsClient';
 import { CRED_KEY, CRED_RENEW_MS } from '../utils/constants';
+import { Configurable } from '../abstract/Configurable';
 
-export class Authentication {
+export class Authentication extends Configurable<Config> {
```
qhanam, line 9: issue: While the built-in plugins all share a `config` property, I don't see the benefit we get by extending other classes like this one.

Perhaps there is no downside either, but I think we should be mindful of over-using inheritance and generics, as they can make the code less understandable (i.e., more difficult to maintain and more likely to introduce bugs).

src/abstract/Configurable.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public setConfigValue<K extends keyof ConfigType>(
+        key: K,
+        value: ConfigType[K]
+    ) {
+        this.config[key] = value;
+    }
+
+    public getConfigValue<K extends keyof ConfigType>(
```
qhanam, line 18: issue: In the interest of ease-of-typing, could we omit the getter and setter?

src/abstract/Configurable.ts
```diff
@@ -0,0 +1,30 @@
+export abstract class Configurable<ConfigType extends object = {}> {
+    private readonly config: ConfigType;
```
qhanam, line 2: issue: I feel like we should just make this a protected member of `InternalPlugin`. I don't think it adds value to the non-plugin classes.
```suggestion
    protected config: ConfigType;
```

maniator: the config should be read only and immuttable

there are many issues r8ght now where changes in one place to a config affect everythkng else (which is hard to test and maintain)

src/abstract/Configurable.ts
maniator: Should have a getter and not have a setter, I only added the setter because we do some weird things in tests to change config values, which really should not happen in general as the config should be set once and read only and immutable

src/dispatch/Authentication.ts
maniator: Really we should not be passing this global "config" around that is shared and mutated by N number of classes...

maniator:
@qhanam I rebased and responded to all comments

src/abstract/Configurable.ts
maniator: Also config _is_ used in many of the non-plugin files

maniator:
ping @qhanam

src/dispatch/Authentication.ts
qhanam: Right, so there is an increased risk that the web client or a plugin could modify the config.

I guess I'm not convinced that (1) this is a problem in practice, or that (2) the config should not be mutable. So adding a method call to access a config property seems unnecessary to me.

PR #167, opened 2022-05-25
chore: 🤖 mark package as sideEffect free

maniator:
Can read documentation of tree-shaking here: https://webpack.js.org/guides/tree-shaking/

and you can see here that even though we are using es2015, it will not really use tree-shaking until it is marked as sideEffect free

[image](https://bundlephobia.com/package/aws-rum-web)

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
     "name": "aws-rum-web",
     "version": "1.5.1",
+    "sideEffects": false,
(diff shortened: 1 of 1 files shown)


qhanam (approved):
I like the idea, however I'm not sure this will cause webpack to prune any additional code, because the built-in plugins are referenced in `Orchestration`.

Either way, the package is indeed side effect free (at the moment), so this change doesn't have any downside.

Issue #168, opened 2022-05-27
Potential slowdown of site due to listening for global keydown events


maniator:
In the [`VirtualPageLoadTimer` plugin](https://github.com/aws-observability/aws-rum-web/blob/main/src/sessions/VirtualPageLoadTimer.ts#L81) on line 81 it has a keydown event, which if a user holds down a key it will get hit every ms until they list their finger from the keyboard.

I think there are two options to fix this issue:

1.  debounce the event for keydown
     Down side to this is that we may not get the most accurate information, and the debounce method still gets hit every ms
2. figure out a better option for detecting a page interaction without hitting the same method n number of times per second

adebayor123:
As discussed offline, `keyup` event and `mouseup` event would be a better solution, as it would not be triggered every ms.

maniator:
I think I may have a solution without any mouse or key events.

May push PR today or Tuesday

PR #169, opened 2022-05-27
chore: 🤖 Update VirtualPageLoad to be a plugin with timings

maniator:
Closes https://github.com/aws-observability/aws-rum-web/issues/168

Updates `VirtualPageLoad` to follow the same plugin paradigm as everything else
Updates the popstate callback to use a `setTimeout` so that it gets called after the page loads (but still get the current start time before the timeout is called) Can see docs for that here: https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event#sect1

---


src/sessions/PageManager.ts
```diff
@@ -51,17 +52,13 @@ export class PageManager {
...
     private recordInteraction: boolean;
 
-    constructor(config: Config, record: RecordEvent) {
-        this.config = config;
+    constructor(
+        private config: Config,
+        private record: RecordEvent,
+        private pluginManager: PluginManager
+    ) {
         this.record = record;
```
maniator, line 60: line can be removed

src/sessions/PageManager.ts
```diff
@@ -51,17 +52,13 @@ export class PageManager {
...
-        this.page = undefined;
-        this.resumed = undefined;
         this.recordInteraction = false;
```
maniator, line 61: can be set on line 53

maniator:
@qhanam @adebayor123 

This PR is ready for review.

maniator:
> If #168 is urgent, I suggest de-coupling the fix from the refactor.
> 
> Otherwise, I'll wait for @adebayor123 to review :-)

Needed to do the refactor in order to decouple and make the fix 😂

All a little confusing

adebayor123:
I think before diving into the big changes, such as moving `VirtualPageLoadTimer` into a plugin that seem to be more of a refactoring change, we should talk about the intention/logic of `VirtualPageLoadTimer`.

First of all, we define the duration of a route change as the difference between `latest (DOM update or AJAX request completion)` and `user interaction that started the route change`. Based on this definition, we need to a mechanism to listen to user interactions that may lead to route change - specifically clicks and key presses. Using the time of `pushState/replaceState/popstate` does not give us this definition. However, I do agree that `keydown` may have performance issues, as a user could spam it. I will discuss this further in the following sections.

I did some research on possible performance impact `keydown` event may have, and seems that the biggest culprits are usually `scroll` and `mousemove` events as they get registered at a very high frequency. Moreover, the only warning I could find on MDN was to avoid doing complex operations in the event handler callbacks - our callback only updates a variable after calling `Date.now()`, so should not be a big deal.

I was not able to find specific data points on how bad `keydown` events could affect an application, so I had to do a manual testing to collect some datapoints. On average, the frequency of `keydown` being registered was around 1 ms interval. However, `mousemove` was being registered at an interval around 0.1 ms. This suggests that we perhaps need not worry about the performance impact of listening to `keydown` as much as `mousemove` or `scroll`. I believe debouncing may be a remedy, but as mentioned above, our callback is a very light operation as it only does `Date.now()`, so I don't think debouncing would provide a significant increase in performance as debouncing helps when the operation is heavy. We can use `keyup`, but I don't think `keydown` at the moment really poses a big performance issue. However, I am more than willing to change if we have datapoints proving otherwise ;-) 

Another issue is that according to the PR, we will only have the timing information (`Date.now()`) when `popstate` event is emitted. However, `popstate` is not fired all the time. According to MDN: `Note that just calling history.pushState() or history.replaceState() won't trigger a popstate event. The popstate event will be triggered by doing a browser action such as a click on the back or forward button (or calling history.back() or history.forward() in JavaScript).` As a result, the proposed change goes against our definition of `route change start time`.

Last but not least, finding where to place `VirtualPageLoadTimer` is tricky. I originally shaped it into a plugin, but after some discussions, we internally agreed that it is too small of a functionality to be a plugin on its own while it is too big of a functionality to be within `PageManager`. As a result, we ended up shaping it as a helper class that's used in `PageManager`. 

These are my thoughts on the proposed changes - let me know what you think!

PR #174, opened 2022-06-08
feat: Retry fetch requests and disable dispatch on failure.

qhanam:
The RUM web client does not currently have mechanisms to (1) retry failed requests, and (2) cease sending requests when there are repeated failures. This is problematic because repeating failed requests can negatively impact application and is unlikely to change the outcome of the requests.

This change causes the web client to retry failed fetch requests up to three times. If a request fails after three retries, the web client will cease sending data to the CloudWatch RUM service.

For example, failures may include: failure to retrieve credentials (e.g., from cognito/sts), failure to sign request (e.g., from malformed credentials), data plane authentication failure (i.e., 403 response code) and network failures (e.g., poor connectivity).

Related to #21

---

### DIFF
--- a/src/dispatch/Dispatch.ts
+++ b/src/dispatch/Dispatch.ts
@@ -6,6 +6,7 @@ import { FetchHttpHandler } from './FetchHttpHandler';
 import { PutRumEventsRequest } from './dataplane';
 import { Config } from '../orchestration/Orchestration';
 import { v4 } from 'uuid';
+import { RetryHttpHandler } from './RetryHttpHandler';
 
 type SendFunction = (
     putRumEventsRequest: PutRumEventsRequest
@@ -101,7 +102,7 @@ export class Dispatch {
      * Send meta data and events to the AWS RUM data plane service via fetch.
      */
     public dispatchFetch = async (): Promise<{ response: HttpResponse }> => {
-        return this.dispatch(this.rum.sendFetch);
+        return this.dispatch(this.rum.sendFetch).catch(this.handleReject);
     };
 
     /**
@@ -189,6 +190,14 @@ export class Dispatch {
         return send(putRumEventsRequest);
     }
 
+    private handleReject = (e: any): { response: HttpResponse } => {
+        // The handler has run out of retries. We adhere to our convention to
+        // fail safe by disabling dispatch. This ensures that we will not
+        // continue to attempt requests when the problem is not recoverable.
+        this.disable();
+        throw e;
+    };
+
     /**
      * The default method for creating data plane service clients.
      * @param endpoint Service endpoint.
@@ -201,9 +210,12 @@ export class Dispatch {
         credentials
     ) => {
         return new DataPlaneClient({
-            fetchRequestHandler: new FetchHttpHandler({
-                fetchFunction: this.config.fetchFunction
-            }),
+            fetchRequestHandler: new RetryHttpHandler(
+                new FetchHttpHandler({
+                    fetchFunction: this.config.fetchFunction
+                }),
+                this.config.retries
+            ),
             beaconRequestHandler: new BeaconHttpHandler(),
             endpoint,
             region,
--- a/src/dispatch/RetryHttpHandler.ts
+++ b/src/dispatch/RetryHttpHandler.ts
@@ -1,5 +1,7 @@
 import { HttpHandler, HttpRequest, HttpResponse } from '@aws-sdk/protocol-http';
 
+export type BackoffFunction = (retry: number) => number;
+
 /**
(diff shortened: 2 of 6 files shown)


src/dispatch/Dispatch.ts
```diff
@@ -101,7 +102,7 @@ export class Dispatch {
      * Send meta data and events to the AWS RUM data plane service via fetch.
      */
     public dispatchFetch = async (): Promise<{ response: HttpResponse }> => {
-        return this.dispatch(this.rum.sendFetch);
+        return this.dispatch(this.rum.sendFetch).catch(this.handleReject);
```
adebayor123, line 105: question: for automatic dispatches that will occur at a predefined interval, we will fail silently - however, if the retries failed after invoking dispatch directly from `orchestration`, we are not failing silently.
```
public dispatch(): void {
        this.dispatchManager.dispatchFetch();
    }
``` 

is this intended?

src/dispatch/Dispatch.ts
```diff
@@ -101,7 +102,7 @@ export class Dispatch {
      * Send meta data and events to the AWS RUM data plane service via fetch.
      */
     public dispatchFetch = async (): Promise<{ response: HttpResponse }> => {
```
adebayor123, line 104: question: is the retry only needed for `fetch`? what about `beacon`?

src/dispatch/RetryHttpHandler.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public async handle(
+        request: HttpRequest
+    ): Promise<{ response: HttpResponse }> {
+        let retriesLeft = this.retries;
+        while (true) {
+            try {
+                const response = await this.handler.handle(request);
+                if (this.isStatusCode2xx(response.response.statusCode)) {
```
adebayor123, line 25: issue: I think we should perhaps treat 4xx and 5xx from data plane separately - if it's 5xx, we know it's a service issue, and hopefully, the issue will be mitigated fairly quickly. It's something we can fix on our end, so once things are recovered on our end, the web client should be able to send requests from there on. On the other hand, if it's 4xx, most likely the setup or credentials are invalid, and without the customer's own fixes, there's no point in sending more events (which is the whole purpose of adding this feature).

src/dispatch/__tests__/RetryHttpHandler.test.ts
```diff
@@ -0,0 +1,150 @@
...
+        try {
+            // Run
+            await client.sendFetch(Utils.PUT_RUM_EVENTS_REQUEST);
+        } catch (e) {
+            // Assert
+            expect(e.message).toEqual('500');
+            return;
+        }
+
+        fail('Request should fail');
```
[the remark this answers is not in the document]
qhanam: Since the request at line 110 should throw an exception and exit at line 114, this causes the test to fail if the request at line 110 succeeds.

src/dispatch/Dispatch.ts
qhanam: I **think** my reasoning here was that if you're manually invoking it you may want to know if it fails.

There's a ticket somewhere to revisit our error handling strategy, but for this change it is out of scope.

src/dispatch/Dispatch.ts
```diff
@@ -189,6 +190,14 @@ export class Dispatch {
         return send(putRumEventsRequest);
     }
 
+    private handleReject = (e: any): { response: HttpResponse } => {
+        // The handler has run out of retries. We adhere to our convention to
+        // fail safe by disabling dispatch. This ensures that we will not
+        // continue to attempt requests when the problem is not recoverable.
+        this.disable();
```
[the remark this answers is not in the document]
qhanam: (1) Unknowable
(2) My thought here was separation of concerns. The cost of recording events is low (i.e., not noticeable), while disabling the web client completely requires a lot more thought. For example, if we did want to implement a self-recovery, we would want to continue recording events.

src/dispatch/Dispatch.ts
adebayor123: Ah, makes sense :)

adebayor123:
Synced offline, there will be a backoff to the retries - otherwise lgtm

jeffschulzusc:
hi ... example, if i go 'offline', 
so i see that after a couple of retry about 1 second apart, it seems like the AwsRum subsequently never dispatches any more ... is that correct ?
- seems it should handle the network becoming available in the future, maybe check and retry after seconds or minutes interval before giving up completely ?

if my js code detects network become available, could i AwsRum.disable then AwsRum.enable or such to affect the dispatching to resume ?

otherwise the AwsRum works great, thanks !

-jeff

PR #177, opened 2022-06-10
feat: Custom attributes

limhjgrace:
**Context**:
Currently, RUM Web Client does not provide a mechanism to add custom attributes (data in the form of <key, value> pairs) to RUM events. This is problematic because users want to add attributes in addition to the default attributes provided by the RUM event schemas for better data analysis when querying or using the RUM console.

**Changes**:
1. Allow the web client to accept session attributes during initialization. 
2. Allow the web client to accept session attributes during runtime through the new `addSessionAttributes` API.
3. Allow the web client to accept page attributes during runtime through the existing `recordPageViews` API.

Invalid custom attributes will be excluded from the event metadata. Custom attributes will be validated to check:
a. Invalid key length
b. Invalid value length
c. Invalid key chars
d. Total number of custom attributes

---

### DIFF
--- a/src/CommandQueue.ts
+++ b/src/CommandQueue.ts
@@ -38,8 +38,8 @@ export class CommandQueue {
         ): void => {
             this.orchestration.setAwsCredentials(payload);
         },
-        setCustomAttributes: (payload: { [k: string]: any }): void => {
-            this.orchestration.setCustomAttributes(payload);
+        addSessionAttributes: (payload: { [k: string]: any }): void => {
+            this.orchestration.addSessionAttributes(payload);
         },
         recordPageView: (payload: any): void => {
             this.orchestration.recordPageView(payload);
--- a/src/CommandQueue.ts
+++ b/src/CommandQueue.ts
@@ -38,6 +38,9 @@ export class CommandQueue {
         ): void => {
             this.orchestration.setAwsCredentials(payload);
         },
+        setCustomAttributes: (payload: { [k: string]: any }): void => {
+            this.orchestration.setCustomAttributes(payload);
+        },
         recordPageView: (payload: any): void => {
             this.orchestration.recordPageView(payload);
         },
--- a/src/event-schemas/meta-data.json
+++ b/src/event-schemas/meta-data.json
@@ -66,7 +66,7 @@
         "pageTags": { "type": "array", "items": { "type": "string" } }
     },
     "patternProperties": {
-        "^(?!pageTags)[a-zA-Z0-9_]{1,128}$": {
+        "^(?!pageTags).{1,128}$": {
             "maxLength": 256,
             "type": ["string", "boolean", "number"]
         },
--- a/src/event-schemas/meta-data.json
+++ b/src/event-schemas/meta-data.json
@@ -65,6 +65,18 @@
         },
         "pageTags": { "type": "array", "items": { "type": "string" } }
     },
+    "patternProperties": {
+        "^(?!pageTags)[a-zA-Z0-9_]{1,128}$": {
+            "maxLength": 256,
+            "type": ["string", "boolean", "number"]
+        },
+        "pageTags": {
+            "type": "array",
+            "items": {
+                "type": "string"
+            }
+        }
+    },
     "additionalProperties": false,
     "required": ["version", "domain"]
 }
--- a/src/orchestration/Orchestration.ts
+++ b/src/orchestration/Orchestration.ts
@@ -63,6 +63,7 @@ export type PartialConfig = {
     batchLimit?: number;
     clientBuilder?: ClientBuilder;
     cookieAttributes?: PartialCookieAttributes;
+    customAttributesMap?: { [k: string]: string | number | boolean };
     disableAutoPageView?: boolean;
     dispatchInterval?: number;
     enableRumClient?: boolean;
@@ -150,6 +151,7 @@ export type Config = {
     batchLimit: number;
     clientBuilder?: ClientBuilder;
     cookieAttributes: CookieAttributes;
+    customAttributesMap?: { [k: string]: string | number | boolean };
     disableAutoPageView: boolean;
     dispatchInterval: number;
     enableRumClient: boolean;
@@ -277,6 +279,14 @@ export class Orchestration {
         credentials: Credentials | CredentialProvider
     ): void {
         this.dispatchManager.setAwsCredentials(credentials);
+    }
+
+    /**
+     * Set custom session attributes to add them to all event metadata.
+     * @param payload object containing custom attribute data in the form of key, value pairs
+     */
+    public setCustomAttributes(payload: { [key: string]: any }): void {
+        this.eventCache.setCustomAttributes(payload);
     }
 
     /**
--- a/src/orchestration/Orchestration.ts
+++ b/src/orchestration/Orchestration.ts
@@ -63,7 +63,7 @@ export type PartialConfig = {
     batchLimit?: number;
     clientBuilder?: ClientBuilder;
     cookieAttributes?: PartialCookieAttributes;
-    customAttributesMap?: { [k: string]: string | number | boolean };
+    sessionAttributes?: { [k: string]: string | number | boolean };
     disableAutoPageView?: boolean;
     dispatchInterval?: number;
     enableRumClient?: boolean;
@@ -151,7 +151,7 @@ export type Config = {
     batchLimit: number;
     clientBuilder?: ClientBuilder;
     cookieAttributes: CookieAttributes;
-    customAttributesMap?: { [k: string]: string | number | boolean };
+    sessionAttributes?: { [k: string]: string | number | boolean };
     disableAutoPageView: boolean;
     dispatchInterval: number;
     enableRumClient: boolean;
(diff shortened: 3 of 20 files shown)


src/orchestration/Orchestration.ts
```diff
@@ -276,6 +278,14 @@ export class Orchestration {
         this.dispatchManager.setAwsCredentials(credentials);
     }
 
+    /**
+     * Set custom session attributes to add them to all event metadata.
+     * @param payload object containing custom attribute data in the form of key, value pairs
+     */
+    public setCustomAttributes(payload: { [key: string]: any }): void {
```
qhanam, line 285: todo: We'll need to brainstorm some names for this API. E.g.,
- `addAttribute`
- `setAttribute`
- `setSessionAttribute`
- `setSessionMetadata`

src/orchestration/Orchestration.ts
```diff
@@ -63,6 +63,7 @@ export type PartialConfig = {
     batchLimit?: number;
     clientBuilder?: ClientBuilder;
     cookieAttributes?: PartialCookieAttributes;
+    customAttributesMap?: { [k: string]: string | number | boolean };
```
qhanam, line 66: todo: We should brainstorm some names for this . E.g.,
- `sessionAttributes`

src/orchestration/Orchestration.ts
qhanam: issue: Could we replace the `any` type with `string | boolean | number`? This would help communicate the expected attribute types to callers.

src/sessions/PageManager.ts
```diff
@@ -17,11 +17,19 @@ export type Attributes = {
     parentPageId?: string;
     interaction?: number;
     pageTags?: string[];
+    // The value types of custom attributes are restricted to the types: string | number | boolean
+    // However, given that pageTags is a string array, we need to include it as a valid type
+    // Events will be verified by our service to validate attribute value types where
+    //  1) pageTags attribute value must be of type string[]
+    //  2) any other attribute value must be of the following types:
+    //      string | number | boolean
+    [k: string]: string | number | boolean | string[];
```
qhanam, line 26: thought: Thanks for documenting this. Maybe in the future we will support arrays in the general case as well :-).

src/sessions/SessionManager.ts
```diff
@@ -125,6 +132,19 @@ export class SessionManager {
...
+     * Adds custom session attributes to the session's attributes
+     * @param customAttributesMap object containing custom attribute data in the form of key, value pairs
+     */
+    public setCustomAttributes(customAttributesMap: {
+        [k: string]: string | number | boolean;
+    }) {
+        const attributeKeys = Object.keys(customAttributesMap);
+        for (const attribute of attributeKeys) {
+            this.attributes[attribute] = customAttributesMap[attribute];
+        }
```
qhanam, line 145: suggestion: The [spread operator ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) can help here. Note that the second spread operator (`...customAttributeMap`) takes precedence, so attributes will be overridden with new attributes in this example.
```suggestion
        this.attributes = { ...this.attributes, ...customAttributeMap };
```

src/sessions/PageManager.ts
```diff
@@ -17,11 +17,19 @@ export type Attributes = {
...
 };
 
 export type PageAttributes = {
     pageId: string;
     pageTags?: string[];
+    [k: string]: string | number | boolean | string[];
```
qhanam, line 32: nit: Additional attributes can only be primitives, right?
```suggestion
    [k: string]: string | number | boolean;
```

src/sessions/SessionManager.ts
```diff
@@ -92,6 +94,11 @@ export class SessionManager {
         // Collect the user agent and domain
         this.collectAttributes();
 
+        // Set custom session attributes
+        if (this.config.customAttributesMap) {
+            this.setCustomAttributes(this.config.customAttributesMap);
+        }
```
qhanam, line 100: nit: The spread operator will ignore customAttributesMap if it is falsey.
```suggestion
        this.setCustomAttributes(this.config.customAttributesMap);
```

src/orchestration/Orchestration.ts
limhjgrace: Updated to use `sessionAttributes`

src/orchestration/Orchestration.ts
limhjgrace: Updated to use `setSessionAttribute`. Replaced `any` with `string | boolean | number`

src/sessions/PageManager.ts
limhjgrace: Yes, additional attributes can only be primitives but if we  remove that from line 26, `pageTags` can't be a `string[]` because of the type collision. Metadata schema will validate attributes where we drop any attributes with `string[]` type if it's not the `pageTags` attribute.

src/sessions/SessionManager.ts
limhjgrace: Updated to use suggestion, thanks!

src/sessions/PageManager.ts
limhjgrace: As mentioned above, I think we need to still allow values to have `string[]` type to avoid the error
`Property 'pageTags' of type 'string[]' is not assignable to 'string' index type 'string | number | boolean'.` 
Metadata schema will validate attributes where we drop any attributes with `string[]` type if it's not the `pageTags` attribute.

src/sessions/SessionManager.ts
limhjgrace: Updated to use suggestion, thanks!

src/sessions/__integ__/SessionManager.test.ts
```diff
@@ -27,6 +30,15 @@ const button1: Selector = Selector(`#${BUTTON_ID_1}`);
...
 fixture('Session Handler usage').page('http://localhost:8080/');
 
+const removeUnwantedEvents = (json: any) => {
+    for (let i = 0; i < json.RumEvents.length; i++) {
+        if (/(session_start_event)/.test(json.RumEvents[i].type)) {
+            json.RumEvents.splice(i, 1);
+        }
+    }
+    return json;
+};
```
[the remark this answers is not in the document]
limhjgrace: Updated accordingly

src/orchestration/Orchestration.ts
```diff
@@ -276,6 +278,16 @@ export class Orchestration {
         this.dispatchManager.setAwsCredentials(credentials);
     }
 
+    /**
+     * Set custom session attributes to add them to all event metadata.
+     * @param payload object containing custom attribute data in the form of key, value pairs
+     */
+    public setCustomAttributes(payload: {
```
qhanam, line 285: nit: `setSessionAttributes` to be consistent with the config option + rename parameter
```suggestion
    public setSessionAttributes(attributes: {
```

src/CommandQueue.ts
```diff
@@ -38,6 +38,9 @@ export class CommandQueue {
         ): void => {
             this.orchestration.setAwsCredentials(payload);
         },
+        setCustomAttributes: (payload: { [k: string]: any }): void => {
```
qhanam, line 41:
```suggestion
        setSessionAttributes: (payload: { [k: string]: any }): void => {
```

src/orchestration/Orchestration.ts
```diff
@@ -148,6 +149,7 @@ export type Config = {
     batchLimit: number;
     clientBuilder?: ClientBuilder;
     cookieAttributes: CookieAttributes;
+    sessionAttributes?: { [k: string]: string | number | boolean };
```
qhanam, line 152: nit: Make this required and make it default to an empty object.
```suggestion
    sessionAttributes: { [k: string]: string | number | boolean };
```

src/sessions/SessionManager.ts
```diff
@@ -125,6 +130,16 @@ export class SessionManager {
         return this.attributes;
     }
 
+    /**
+     * Adds custom session attributes to the session's attributes
+     * @param sessionAttributes object containing custom attribute data in the form of key, value pairs
+     */
+    public setCustomAttributes(sessionAttributes: {
```
qhanam, line 137:
```suggestion
    public setSessionAttributes(sessionAttributes: {
```

qhanam:
We'll need to update the documentation as well:
1. The `sessionAttributes` config option doc
1. The `addSessionAttributes` API doc

limhjgrace:
> We'll need to update the documentation as well:
> 
>     1. The `sessionAttributes` config option doc
> 
>     2. The `addSessionAttributes` API doc

david-siqi-liu:
`sessionAttributes` would be extremely useful for our use case. May I know if the current status of this PR?

src/event-schemas/meta-data.json
```diff
@@ -65,6 +65,18 @@
         },
         "pageTags": { "type": "array", "items": { "type": "string" } }
     },
+    "patternProperties": {
+        "^(?!pageTags)[a-zA-Z0-9_]{1,128}$": {
```
qhanam, line 69: question: Do we need these restrictions? Will these be enforced by the service?

qhanam:
You can update `ci.yaml` to run the CI workflow on this branch.

For example, see [this commit](https://github.com/aws-observability/aws-rum-web/commit/b71aaa9e5d34fc12ac81f0edf2cdc1de375698a0#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR5).

This might be a good time to think about adopting [conventions for branching](https://nvie.com/posts/a-successful-git-branching-model/).

src/event-schemas/meta-data.json
limhjgrace: Don't need character restrictions but will want to keep length and type validations, which will be enforced by backend. Updated to remove char restrictions.

qhanam (approved):

PR #180, opened 2022-07-13
chore: Make pagesToInclude consistent with urlsToInclude

adebayor123:
## Details 

Even though the intended behavior and mechanism between `pagesToInclude` and `urlsToInclude` are similar to each other, the default value of `pagesToInclude` was an empty array and the default value of `urlsToInclude` was `[/.*/]`. As a result, this inconsistency in default values leads to confusion for users while reading the configuration documentation.

This PR changes the default value of `pagesToInclude` from `[]`  to `[/.*/]` to maintain parity and consistency between the two configuration values.

## Testing
No unit tests were introduced, as the change preserves the existing logic of `isCurrentUrlAllowed`.

---


qhanam (approved):
Nice

PR #186, opened 2022-07-26
ci: 🎡 run integ tests in headless on windows as well

maniator:
---

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,8 +43,43 @@ jobs:
             - name: Run Unit Tests
               run: npm run test
 
-    integ-test:
-        name: Run Integ Tests on ${{ matrix.os }}
+    integ-test-browser:
+        name: Run Browser Integ Tests on ${{ matrix.os }}
+        runs-on: ${{ matrix.os }}
+        permissions: read-all
+        env:
+            NODE_OPTIONS: --max-http-header-size=80000
+        strategy:
+            fail-fast: false
+            matrix:
+                os: [windows-latest]
+
+        steps:
+            - name: Checkout AWS RUM Web Client Repository
+              uses: actions/checkout@v2
+
+            - name: Setup Node
+              uses: actions/setup-node@v2
+              with:
+                  node-version: '16.x'
+
+            - name: Cache NPM modules
+              uses: actions/cache@v2
+              with:
+                  path: ~/.npm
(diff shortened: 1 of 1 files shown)


qhanam:
Hi @maniator what is the motivation for this change?

Running the windows integ test headless might speed up the workflow (and maybe reduce the windows test flakiness), but is there a benefit to running both the headless and non-headless windows integ tests?

maniator:
> Hi @maniator what is the motivation for this change?
> 
> Running the windows integ test headless might speed up the workflow (and maybe reduce the windows test flakiness), but is there a benefit to running both the headless and non-headless windows integ tests?

Really prob better to run both in headless

But the headless config does not include Microsoft edge (which we can probably just update and create a Windows edge config and a unix edge config

qhanam:
So with this change, the windows CI workflow will be:
1. Run integ tests with UI
   1. Chrome
   1. Firefox
   1. Edge
1. Run integ tests in headless
   1. Chrome
   1. Firefox

> But the headless config does not include Microsoft edge (which we can probably just update and create a Windows edge config and a unix edge config

Should we do this before merging so we don't re-run the tests? Or would you rather merge this and make the change later?

.github/workflows/ci.yaml
```diff
@@ -73,13 +73,41 @@ jobs:
...
+        env:
+            NODE_OPTIONS: --max-http-header-size=80000
+        strategy:
+            fail-fast: false
+            matrix:
+                os: [ubuntu-latest, windows-latest]
+
+        steps:
+            - name: Checkout AWS RUM Web Client Repository
+              uses: actions/checkout@v2
```
piwysocki, line 94:
```suggestion
              uses: actions/checkout@v3
```

.github/workflows/ci.yaml
```diff
@@ -73,13 +73,41 @@ jobs:
...
+
+            - name: Setup Node
+              uses: actions/setup-node@v2
```
piwysocki, line 97:
```suggestion
              uses: actions/setup-node@v3
```

.github/workflows/ci.yaml
```diff
@@ -73,13 +73,41 @@ jobs:
...
+              with:
+                  node-version: '16.x'
+
+            - name: Cache NPM modules
+              uses: actions/cache@v2
```
piwysocki, line 102:
```suggestion
              uses: actions/cache@v3
```

qhanam:
Closing, as Windows browser integ tests are now running headless.

PR #185, opened 2022-07-26
feat: 🎸 add recordEvent to orchestration and command queue

maniator:
Add `recordEvent` to RUM which will be added to the `interaction` telemetry on page load.

Closes #139 

---

Example JSON sent via [PutRumEvents](https://docs.aws.amazon.com/cloudwatchrum/latest/APIReference/API_PutRumEvents.html) api (in an integ test):

```
{
  BatchId: 'b18cb16b-47f8-4d16-a67d-831e1f17ae3c',
  AppMonitorDetails: { id: 'abc123', version: '1.0' },
  UserDetails: {
    userId: 'e77ad812-00cd-4fdb-a051-a247e5618a7b',
    sessionId: '9cd9bec8-c5f7-4e9b-a8ae-a75fad02f749'
  },
  RumEvents: [
    {
      id: 'bad26fe3-e763-4429-87de-df43b49cf917',
      timestamp: 1658867045,
      type: 'com.amazon.rum.js_general_event',
      metadata: '{"version":"1.0.0","browserLanguage":"en-US","browserName":"Chrome Headless","browserVersion":"103.0.5060.134","osName":"Mac OS","osVersion":"10.15.7","deviceType":"desktop","platformType":"web","domain":"localhost","title":"RUM Integ Test","pageId":"/js_general_event.html","interaction":0}',
      details: '{"version":"1.0.0","name":"random-event","data":"Some random event"}'
    }
  ]
}
```

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #139, opened 2022-04-28
Add log level support


Anntoin:
We've found that `recordError` is a useful replacement for custom client error logging. However to completely replace what we have now we'd need support for other log levels, e.g. info, warn.

This could be provided by a `recordLog` or `recordEvent` command that supports arbitrary logging. Is this something that would be considered in scope for the rum client?

qhanam:
Hi @Anntoin thanks for raising this (and sorry taking so long to respond)! I agree these are important use cases.

For both `recordLog` and `recordEvent`, the CloudWatch RUM service must first support these application-specific event types. Currently, it only supports [a fixed set of events](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-RUM-datacollected.html).

For example, it is currently possible to use the CloudWatch RUM web client (by creating a plugin) or the [CloudWatch RUM SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-rum/index.html) to send arbitrary events to CloudWatch RUM, however only those events which conform to [an accepted schema](https://github.com/aws-observability/aws-rum-web/tree/main/src/event-schemas) will be ingested (the rest will be dropped).

We will update this thread when CloudWatch RUM has added support, but I can't provide an ETA.

maniator:
I am going to see if it is possible to add a recordEvent method to go according to the RUM schema here: https://docs.aws.amazon.com/cloudwatchrum/latest/APIReference/API_PutRumEvents.html

maniator:
@Anntoin see https://github.com/aws-observability/aws-rum-web/pull/185 for possible solution :-)

Anntoin:
That's awesome 🎉 thanks for following up!

This should cover our use case. I'm on vacation at the moment but looking
forward to checking it out when I get back.

This makes RUM a one-stop shop for all our client-side visibility needs!

williazz:
`recordEvent` was introduced in https://github.com/aws-observability/aws-rum-web/pull/188

williazz:
Closing becuase we are only interested in events

PR #188, opened 2022-08-01
feat: Add recordEvent API and expose Plugin to enable recording of custom events

adebayor123:
## Detail
This PR exposes the Plugin Interface and two new APIs, `recordEvent` and `addPlugin`, to allow customers to implement customized plugins and record custom event using the web client. With this change, customers can record custom events by either making a direct API call to record events or implementing and registering their customized plugins.

Closes #185 and #139 

## Testing
Added new unit test cases and integ test cases to verify behavior.

---


docs/custom_events.md
```diff
@@ -0,0 +1,50 @@
+# Using CloudWatch RUM Web Client to Record Custom Events
```
adebayor123, line 1: Thought a dedicated document would be better than appending onto `docs/cdn_commands.md`. Let me know if you think otherwise.

src/orchestration/Orchestration.ts
```diff
@@ -352,6 +353,17 @@ export class Orchestration {
         );
     }
 
+    /**
+     * Record an event directly to eventCache.
+     * @param type
+     * @param eventData
+     */
+    public recordEvent(eventType: string, eventData: object) {
+        if (!EVENT_TYPE_NAMESPACE.has(eventType)) {
```
adebayor123, line 362: Removing this check allows users to send built-in events by creating their own data (assuming it conforms to the schema), which addresses a use case brought up before, and in the case where they do send wrong data format, at the end of the day, it will be handled by the Data Plane. Let me know what you think.

src/event-cache/EventCache.ts
```diff
@@ -88,7 +91,11 @@ export class EventCache {
             const session: Session = this.sessionManager.getSession();
             this.sessionManager.incrementSessionEventCount();
 
-            if (this.canRecord(session)) {
+            if (
+                this.canRecord(session) &&
+                this.isEventTypePatternAllowed(type) &&
+                Object.keys(eventData).length !== 0
```
adebayor123, line 97: Since `eventCache.recordEvent` was only accessible via internal implementations, we would never have to worry about empty data. Now that customers can directly send in data and record "custom" events, it is possible that empty data would be passed in, and I thought the benefit of preventing the ingestion of empty data outweighed the use case of sending empty data.

src/CommandQueue.ts
```diff
@@ -47,6 +56,20 @@ export class CommandQueue {
...
             this.orchestration.registerDomEvents(payload);
         },
+        recordEvent: (payload: RecordEventInput) => {
+            const type = payload.event_type;
+            const eventData = payload.event_data;
+            if (
+                type === undefined ||
+                eventData === undefined ||
+                typeof type !== 'string' ||
+                typeof eventData !== 'object'
```
maniator, line 66: Can we make event data _not_ required?

Sometimes we may want to record a straight event with no data

src/event-cache/EventCache.ts
maniator: Why are we preventing empty data?

Sometimes we might just want to know how many times one event happened (with no data attached to that event)

src/orchestration/__tests__/Orchestration.test.ts
```diff
@@ -6,6 +6,7 @@ import { DomEventPlugin } from '../../plugins/event-plugins/DomEventPlugin';
 import { JsErrorPlugin } from '../../plugins/event-plugins/JsErrorPlugin';
 import { PluginManager } from '../../plugins/PluginManager';
 import { PageIdFormatEnum } from '../Orchestration';
+import { JS_ERROR_EVENT_TYPE } from '../../plugins/utils/constant';
```
maniator, line 9: why is this added?

src/plugins/utils/constant.ts
```diff
@@ -28,3 +26,20 @@ export const PAGE_VIEW_EVENT_TYPE = `${RUM_AMZ_PREFIX}.page_view_event`;
 
 // Session start event
 export const SESSION_START_EVENT_TYPE = `${RUM_AMZ_PREFIX}.session_start_event`;
+
+// Protected namespace for built-in events
+export const EVENT_TYPE_NAMESPACE = new Set<string>([
```
maniator, line 31: You may want to wrap this in `Object.freeze` otherwise it is mutable.

src/CommandQueue.ts
```diff
@@ -21,6 +22,14 @@ export type AwsRumClientInit = {
     u?: string;
 };
 
+/**
+ * The input for recordEvent command
+ */
+export type RecordEventInput = {
+    event_type: string;
+    event_data: object;
```
maniator, line 30: can we make the data optional?

maniator: (see other comments)

app/custom_event.html
```diff
@@ -0,0 +1,136 @@
...
+            function recordEvent() {
+                cwr('recordEvent', {
+                    event_type: 'custom_event_api',
+                    event_data: { customEventVersion: 255 }
+                });
+            }
+
+            // Add plugin
+            function addPlugin() {
+                class newPlugin {
```
maniator, line 28: Can we use js paradigm of naming classes with uppercase letters?

app/custom_event.html
```diff
@@ -0,0 +1,136 @@
...
+                    }
+                    getPluginId() {
+                        return this.id;
+                    }
+                    record(data) {
+                        this.context.record(this.id, data);
+                    }
+                }
+                this.plugin = new newPlugin();
+                cwr('addPlugin', this.plugin);
```
maniator, line 51: shouldn't `addPlugin` make sure that the plugin is an instance of the plugin `Plugin` class?

maniator: oooo we have it as an interface...

We may want to change that to make sure passed in plugins are extending the plugin class type (https://github.com/aws-observability/aws-rum-web/blob/main/src/plugins/Plugin.ts)

docs/cdn_commands.md
```diff
@@ -61,7 +61,8 @@ if(awsCreds) awsRum.setAwsCredentials(credentialProvider);
...
 | setAwsCredentials | [Credentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Credentials.html) \|&nbsp;[CredentialProvider](https://www.npmjs.com/package/@aws-sdk/credential-providers) | `cwr('setAwsCredentials', cred);`<br/><br/>`awsRum.setAwsCredentials(cred);` | Forward AWS credentials to the web client. The web client requires AWS credentials with permission to call the `PutRumEvents` API. If you have not set `identityPoolId` and `guestRoleArn` in the web client configuration, you must forward AWS credentials to the web client using this command.
-
+| recordEvent | Object | `cwr('recordEvent', {event_type: 'your_event_type', event_data: {field1: 1, field2: 2}})` <br/><br/> `awsRum.recordEvent('your_event_type', {field1 1, field2: 2})` | Record an event. Learn more in [Using CloudWatch RUM Web Client to Record Custom Events](custom_events.md)
+| addPlugin | Plugin | `cwr('addPlugin', this.myCustomPlugin)` <br/><br/> `awsRum.addPlugin(this.myCustomPlugin)` | Adds a plugin to the Web Client. Learn more in [Using CloudWatch RUM Web Client to Record Custom Events](custom_events.md)
```
qhanam, line 65: I'm probably contradicting something I said earlier, but it might be better to do docs in a separate commit. This way we can separate the technical conversation from the doc revisions.

src/CommandQueue.ts
```diff
@@ -47,6 +56,20 @@ export class CommandQueue {
         registerDomEvents: (payload: any): void => {
             this.orchestration.registerDomEvents(payload);
         },
+        recordEvent: (payload: RecordEventInput) => {
+            const type = payload.event_type;
+            const eventData = payload.event_data;
+            if (
+                type === undefined ||
+                eventData === undefined ||
```
qhanam, line 64: nit: These checks are redundant.
```
typeof undefined === 'undefined'
```

src/CommandQueue.ts
```diff
@@ -47,6 +56,20 @@ export class CommandQueue {
         registerDomEvents: (payload: any): void => {
             this.orchestration.registerDomEvents(payload);
         },
+        recordEvent: (payload: RecordEventInput) => {
+            const type = payload.event_type;
+            const eventData = payload.event_data;
```
qhanam, line 61: nit: Using `payload.type` and `payload.data` is fine, no need to re-assign.

src/CommandQueue.ts
```diff
@@ -47,6 +56,20 @@ export class CommandQueue {
         registerDomEvents: (payload: any): void => {
             this.orchestration.registerDomEvents(payload);
         },
+        recordEvent: (payload: RecordEventInput) => {
```
qhanam, line 59: issue: If we're going to check the payload for type safety, the proper type is `unknown` (indicates the type could be anything but is the opposite of `any` in that `unknown` values cannot be assigned to any other type).

src/CommandQueue.ts
```diff
@@ -47,6 +56,20 @@ export class CommandQueue {
...
+            const type = payload.event_type;
+            const eventData = payload.event_data;
+            if (
+                type === undefined ||
+                eventData === undefined ||
+                typeof type !== 'string' ||
+                typeof eventData !== 'object'
+            ) {
+                throw new Error('IncorrectParametersException: invalid input.');
```
qhanam, line 68: nit: For consistency.
```suggestion
                throw new Error('IncorrectParametersException')
```

src/event-cache/__tests__/EventCache.integ.test.ts
```diff
@@ -25,9 +25,9 @@ describe('EventCache tests', () => {
         const eventCache: EventCache = Utils.createEventCache(config);
 
         // Run
-        eventCache.recordEvent(EVENT1_SCHEMA, {});
+        eventCache.recordEvent(EVENT1_SCHEMA, { field: 'value' });
```
qhanam, line 28: nit: Once we revert to allowing empty objects in `recordEvent`, please revert these :-)

src/orchestration/Orchestration.ts
adebayor123: Sounds good, as it also allows users to send built-in events on demand. Will reflect in next revision.

src/event-cache/__tests__/EventCache.integ.test.ts
adebayor123: Done!

src/orchestration/Orchestration.ts
```diff
@@ -352,6 +353,17 @@ export class Orchestration {
         );
     }
 
+    /**
+     * Record an event directly to eventCache.
+     * @param type
+     * @param eventData
+     */
+    public recordEvent(eventType: string, eventData: object) {
```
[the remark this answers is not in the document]
adebayor123: I think object is better due to two reasons:
1. We will serialize in `eventCache.addRecordToCache`
2. Accepting object will lower the chances of users sending invalid JSON objects in the `details` field.

app/custom_event.html
adebayor123: Agreed!

src/CommandQueue.ts
adebayor123: Good suggestion

src/event-cache/EventCache.ts
adebayor123: I admit, while implementing this I was mistaking empty data with null values. I'll be making `{}` as default value if not passed in.

src/CommandQueue.ts
adebayor123: Updated CommandQueue to be able to send events without containing eventData field, and Orchestration to give default value of empty object when not passed in.

src/event-cache/EventCache.ts
```diff
@@ -228,4 +235,12 @@ export class EventCache {
 
         return include && !exclude;
     }
+
+    /**
+     * Returns true if the eventType consists of valid characters. Returns false otherwise.
+     */
+    private isEventTypePatternAllowed(eventType: string) {
+        const valid = new RegExp('^(?!\\.)[\\.\\-_#A-Za-z0-9]+$');
```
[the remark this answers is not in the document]
adebayor123: (1) These are the characters that are allowed for AppMonitor names, so it somewhat maintains consistency in naming convention,  (2) the only difference between existing event types and this pattern is the addition of `#, -, _`.

From what I found, it seems there's no so called `invalid` characters for field values in ES, but I think it's better to follow the syntax we have for AppMonitor names.

app/custom_event.html
adebayor123: It seems the best way to check if an object implements an interface is to manually do user defined type guard (checking if the object has access to all the methods), which means any Object that has the methods in the Interface will pass the type guard. Do you have suggestions for this?

docs/custom_events.md
adebayor123, line 1: Add example document that shows how to do x,y,z

src/event-cache/EventCache.ts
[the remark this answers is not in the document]
adebayor123: Synced up offline, decided to go without the pattern check

app/custom_event.html
adebayor123: Synced up with Quinn, whether we add the type check or not, the plugin will most likely fail during `addPlugin` stage.

src/CommandQueue.ts
```diff
@@ -65,6 +80,9 @@ export class CommandQueue {
             } else {
                 throw new Error('IncorrectParametersException');
             }
+        },
+        addPlugin: (plugin: Plugin) => {
+            this.orchestration.addPlugin(plugin);
```
qhanam, line 85: issue: We have an interface for adding plugins in the [`AwsRumConfiguration` object](https://github.com/aws-observability/aws-rum-web/blob/main/src/orchestration/Orchestration.ts#L72) ([example usage](https://github.com/aws-observability/aws-rum-web/blob/main/src/loader/loader-js-error-event.js#L8)). 

Is there benefit for having this as well?

src/event-cache/EventCache.ts
```diff
@@ -75,11 +75,11 @@ export class EventCache {
      * Add an event to the cache and reset the session timer.
      *
      * If the session is being recorded, the event will be recorded.
-     * If the session is not being recorded, the event will not be recorded.
+     * If the session is not being recorded, the event will not be recorded
      *
      * @param type The event schema.
      */
-    public recordEvent = (type: string, eventData: object) => {
+    public recordEvent = (type: string, eventData = {}) => {
```
qhanam, line 82: nit: Missing type

src/orchestration/Orchestration.ts
```diff
@@ -352,6 +352,15 @@ export class Orchestration {
         );
     }
 
+    /**
+     * Records a custom event.
+     * @param type A unique name for the type of event being recorded.
+     * @param eventData A JSON object containing the event's attributes.
+     */
+    public recordEvent(eventType: string, eventData = {}) {
```
qhanam, line 360: nit: Missing explicit type declaration

docs/cdn_commands.md
```diff
@@ -61,7 +61,8 @@ if(awsCreds) awsRum.setAwsCredentials(credentialProvider);
...
 | setAwsCredentials | [Credentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Credentials.html) \|&nbsp;[CredentialProvider](https://www.npmjs.com/package/@aws-sdk/credential-providers) | `cwr('setAwsCredentials', cred);`<br/><br/>`awsRum.setAwsCredentials(cred);` | Forward AWS credentials to the web client. The web client requires AWS credentials with permission to call the `PutRumEvents` API. If you have not set `identityPoolId` and `guestRoleArn` in the web client configuration, you must forward AWS credentials to the web client using this command.
-
+| recordEvent | Object | `cwr('recordEvent', {event_type: 'your_event_type', event_data: {field1: 1, field2: 2}})` <br/><br/> `awsRum.recordEvent('your_event_type', {field1: 1, field2: 2})` | Record a custom event. Learn more in [Using CloudWatch RUM Web Client to Record Custom Events](custom_events.md)
```
qhanam, line 64: nit: Broken link.

src/event-cache/EventCache.ts
adebayor123: From what I saw, default values declares the type as well: https://www.typescriptlang.org/docs/handbook/functions.html - The reason for adding a default value was due to this comment: https://github.com/aws-observability/aws-rum-web/pull/188#discussion_r937359770

src/orchestration/Orchestration.ts
adebayor123: Answered above

docs/cdn_commands.md
adebayor123: The link should be broken because the documentations were separated into a different file - #195

src/CommandQueue.ts
adebayor123: Good point. One case that would be beneficial of having an `addPlugin` method is that it gives the users the flexibility to initialize a plugin at any point in time. Initializing plugins only via the config does not allow that as far as I know.

src/event-cache/EventCache.ts
adebayor123: Actually, just tested, it does not declare the type explicitly. Will be making changes

src/CommandQueue.ts
qhanam: Plugins lifecycle management happens through enable, disable and update, which I think covers any situation where `addPlugin` would be used. Since there is no `deletePlugin`, the expectation is that plugins live for the duration of the web client instance.

maniator, line 85: Do we want type checks here to make sure the passed value is a plugin? (not every consumer will be using typescript)

app/custom_event.html
```diff
@@ -0,0 +1,153 @@
...
+                cwr('recordEvent', {
+                    event_type: 'custom_event_api',
+                    event_data: { customEventVersion: 255 }
+                });
+            }
+
+            function recordEmptyEvent() {
+                cwr('recordEvent', {
+                    event_type: 'custom_event_api',
+                    event_data: {}
```
maniator, line 29: will you always have to send an empty object if you dont want to send any data?

adebayor123: Yes, that's the idea.

maniator:
Any updates on when this might get deployed to the next version.

src/orchestration/Orchestration.ts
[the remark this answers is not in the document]
maniator: I would say make it generic:

```typescript
public recordEvent<T extends unknown = object>(eventType: string, eventData: T) {
...

File: src/CommandQueue.ts [hunk 23]
```diff
@@ -61,6 +63,20 @@ export class CommandQueue {
...
         },
+        recordEvent: (payload: any) => {
+            if (
+                typeof payload === 'object' &&
+                typeof payload.event_type === 'string' &&
+                typeof payload.event_data === 'object'
+            ) {
+                this.orchestration.recordEvent(
+                    payload.event_type,
+                    payload.event_data
```
qhanam, line 74: issue: Convention is to use camel case for object property names.

limhjgrace (approved):
Synced offline to clarify some questions, lgtm!

qhanam (approved):

docs/cdn_commands.md
```diff
@@ -62,6 +62,7 @@ if(awsCreds) awsRum.setAwsCredentials(credentialProvider);
...
 | setAwsCredentials | [Credentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Credentials.html) \|&nbsp;[CredentialProvider](https://www.npmjs.com/package/@aws-sdk/credential-providers) | `cwr('setAwsCredentials', cred);`<br/><br/>`awsRum.setAwsCredentials(cred);` | Forward AWS credentials to the web client. The web client requires AWS credentials with permission to call the `PutRumEvents` API. If you have not set `identityPoolId` and `guestRoleArn` in the web client configuration, you must forward AWS credentials to the web client using this command.
+| recordEvent | Object | `cwr('recordEvent', {type: 'your_event_type', data: {field1: 1, field2: 2}})` <br/><br/> `awsRum.recordEvent('your_event_type', {field1: 1, field2: 2})` | Record a custom event. Learn more in [Using CloudWatch RUM Web Client to Record Custom Events](examples.md)
```
qhanam, line 65:
```suggestion
| recordEvent | Object | `cwr('recordEvent', {type: 'your_event_type', data: {field1: 1, field2: 2}})` <br/><br/> `awsRum.recordEvent('your_event_type', {field1: 1, field2: 2})` | Record a custom event. 
```

PR #192, opened 2022-08-04
fix: Set appropriate error codes in X-Ray segments

qhanam:
When an error, fault or throttle event occurs, the web client does not currently set the appropriate status flag in the X-Ray segment or subsegment. When viewing the trace in X-Ray, this causes the request to look like it succeeded when it failed.

According to the [X-Ray segment documentation](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-errors):
> Record errors in segments when your application returns an error to the user, and in subsegments when a downstream call returns an error.

This change sets the appropriate `error`, `fault` or `throttle` flag when (1) a 4xx or 5xx status code is returned, or (2) the http request fails.

---

### DIFF
--- a/src/plugins/event-plugins/FetchPlugin.ts
+++ b/src/plugins/event-plugins/FetchPlugin.ts
@@ -16,7 +16,10 @@ import {
     createXRaySubsegment,
     requestInfoToHostname,
     addAmznTraceIdHeaderToHeaders,
-    resourceToUrlString
+    resourceToUrlString,
+    is429,
+    is4xx,
+    is5xx
 } from '../utils/http-utils';
 import { HTTP_EVENT_TYPE, XRAY_TRACE_EVENT_TYPE } from '../utils/constant';
 import {
@@ -24,7 +27,6 @@ import {
     isErrorPrimitive
 } from '../utils/js-error-utils';
 import { HttpEvent } from '../../events/http-event';
-import { InternalPlugin } from '../InternalPlugin';
 
 type Fetch = typeof fetch;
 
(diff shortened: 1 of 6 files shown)


src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -143,14 +145,31 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
                 xRayTraceEvent.subsegments[0].http.response = {
                     status: response.status
                 };
+
+                if (is429(response.status)) {
+                    xRayTraceEvent.subsegments[0].throttle = true;
```
adebayor123, line 150: question: I'm assuming `subsegments[0]` refers to the very first downstream call from the application. If that call returns an error/fault/throttle, we update `subsegments[0]` accordingly. How is `subsegment[1..n]` updated?

limhjgrace (approved):

src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -143,14 +145,31 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
...
+                }
+
                 const cl = parseInt(response.headers.get('Content-Length'), 10);
                 if (!isNaN(cl)) {
                     xRayTraceEvent.subsegments[0].http.response.content_length = cl;
                 }
             }
 
             if (error) {
-                xRayTraceEvent.subsegments[0].error = true;
```
adebayor123, line 153: question: are `Error` objects only created for 5xx?

adebayor123 (approved):

src/plugins/event-plugins/FetchPlugin.ts
adebayor123: Looking at the code, it seems like we're responsible for generating the subsegment of the first downstream call, which makes sense.

src/plugins/event-plugins/FetchPlugin.ts
adebayor123: If an error is passed in along with a status code of 4xx, it will lead to `xRayTraceEvent.subsegments[0].error = true, xRayTraceEvent.subsegments[0].fault=true, xRayTraceEvent.fault = true, xRayTraceEvent.fault = true`. Is it a case we do not have to worry about?

qhanam: In this branch a network error has occurred. There is no response and by extension no status code.

src/plugins/event-plugins/FetchPlugin.ts
qhanam: Correct -- there is only one subsegment.

PR #194, opened 2022-08-08
feat: Add useBeacon config for visibilitychange dispatch behavior.

qhanam:
When the page transitions to a hidden state and fires the `visibilitychange` event, the page [may be in the process of unloading](https://developer.chrome.com/blog/page-lifecycle-api/), so the web client flushes it's event cache. In this scenario, the approach [recommended by MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event) is to use `sendBeacon` so that the request will complete after the page is unloaded. However, ad blockers prevent `sendBeacon` from functioning, so this request may not be sent.

We therefore have two bad options:
1. Use sendBeacon and accept missing data when ad blockers are used and the page loses visibility
2. Use fetch and accept missing data when the page is unloaded before fetch completes
A third option is to send both, however this would increase bandwidch and require deduping server side.

This change adds a config option that controls whether `fetch` or `sendBeacon` is used in this scenario, so that application owners can select the behavior that best fits their application.

---

### DIFF
--- a/src/dispatch/Dispatch.ts
+++ b/src/dispatch/Dispatch.ts
@@ -152,13 +152,22 @@ export class Dispatch {
             // case. However, ad-blockers prevent sendBeacon from functioning.
             // We therefore have two bad options:
             //
-            // (1) Use sendBeacon and accept missing data when ad blockers are
+            // (1) Use sendBeacon. Data will be lost when ad blockers are
             //     used and the page loses visibility
-            // (2) Use fetch and accept missing data when the page is unloaded
+            // (2) Use fetch. Data will be lost when the page is unloaded
             //     before fetch completes
             //
             // A third option is to send both, however this would increase
             // bandwitch and require deduping server side.
+            this.config.useBeacon
+                ? this.dispatchBeaconFailSilent
+                : this.dispatchFetchFailSilent
+        );
+        // Using 'pagehide' is redundant most of the time (visibilitychange is
+        // always fired before pagehide) but older browsers may support
+        // 'pagehide' but not 'visibilitychange'.
+        document.addEventListener(
+            'pagehide',
             this.config.useBeacon
                 ? this.dispatchBeaconFailSilent
                 : this.dispatchFetchFailSilent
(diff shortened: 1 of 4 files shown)


src/dispatch/Dispatch.ts
```diff
@@ -142,9 +147,22 @@ export class Dispatch {
     public startDispatchTimer() {
         document.addEventListener(
             'visibilitychange',
-            this.dispatchBeaconFailSilent
+            // The page is moving to the hidden state, which means it may be
+            // unloaded. The sendBeacon API would typically be used in this
+            // case. However, ad-blockers prevent sendBeacon from functioning.
+            // We therefore have two bad options:
+            //
+            // (1) Use sendBeacon and accept missing data when ad blockers are
```
adebayor123, line 155: question: What does `accept missing data` mean?

src/dispatch/Dispatch.ts
```diff
@@ -142,9 +147,22 @@ export class Dispatch {
...
+            // (2) Use fetch and accept missing data when the page is unloaded
+            //     before fetch completes
+            //
+            // A third option is to send both, however this would increase
+            // bandwitch and require deduping server side.
+            this.config.useBeacon
+                ? this.dispatchBeaconFailSilent
+                : this.dispatchFetchFailSilent
         );
-        document.addEventListener('pagehide', this.dispatchBeaconFailSilent);
```
adebayor123, line 147: question: Are we removing this since `visibilitychange` implies `pagehide`?

adebayor123: If so, should we add a unit test case to verify `pagehide` is handled properly as intended?

adebayor123 (approved):

src/dispatch/Dispatch.ts
qhanam: That was the idea, although as I look at the compatibility tables, `pagehide` is supported in older browser versions which the web client technically supports, so perhaps we should keep the event listener for `pagehide`.

PR #205, opened 2022-08-12
feat: Custom attributes

limhjgrace:
**Context**:
Currently, RUM Web Client does not provide a mechanism to add custom attributes (data in the form of <key, value> pairs) to RUM events. This is problematic because users want to add attributes in addition to the default attributes provided by the RUM event schemas for better data analysis when querying or using the RUM console.

**Changes**:
1. Allow the web client to accept session attributes during initialization. 
2. Allow the web client to accept session attributes during runtime through the new `addSessionAttributes` API.
3. Allow the web client to accept page attributes during runtime through the existing `recordPageViews` API.

Invalid custom attributes will be excluded from the event metadata. Custom attributes will be validated to check:
a. Invalid key length
b. Invalid value length
c. Invalid key chars
d. Total number of custom attributes

---


limhjgrace:
Duplicate of https://github.com/aws-observability/aws-rum-web/pull/177

PR #228, opened 2022-09-14
Refactored recordPageView to add custom attributes in one param

limhjgrace:
Previously, users would add custom page attributes to page view events by adding them as individual fields into the object sent in the `recordPageView()` command. However, to keep parity with how users send in custom session attributes, this change will allow users to send the custom page attributes together in a separate field.

For example:

```
//Before
{
  pageId: '/page_view_two',
  pageTags: ['pageGroup1'],
  customPageAttributeString: 'customPageAttributeValue',
  customPageAttributeNumber: 1,
  customPageAttributeBoolean: true
}
```

```
//After
{
  pageId: '/page_view_two',
  pageTags: ['pageGroup1'],
  pageAttributes: {
    customPageAttributeString: 'customPageAttributeValue',
    customPageAttributeNumber: 1,
    customPageAttributeBoolean: true
  } 
}
```
---

### DIFF
--- a/src/sessions/PageManager.ts
+++ b/src/sessions/PageManager.ts
@@ -3,6 +3,7 @@ import { RecordEvent } from '../plugins/types';
 import { PageViewEvent } from '../events/page-view-event';
 import { VirtualPageLoadTimer } from '../sessions/VirtualPageLoadTimer';
 import { PAGE_VIEW_EVENT_TYPE } from '../plugins/utils/constant';
+import { DEFAULT_PAGE_ATTRIBUTES } from '../utils/constants';
 
 export type Page = {
     pageId: string;
@@ -193,7 +194,7 @@ export class PageManager {
         if (customPageAttributes?.pageAttributes) {
             Object.keys(customPageAttributes.pageAttributes).forEach(
                 (attribute) => {
-                    if (attribute !== 'pageId') {
+                    if (!DEFAULT_PAGE_ATTRIBUTES.includes(attribute)) {
                         this.attributes[attribute] =
                             customPageAttributes.pageAttributes[attribute];
                     }
--- a/src/sessions/PageManager.ts
+++ b/src/sessions/PageManager.ts
@@ -29,7 +29,7 @@ export type Attributes = {
 export type PageAttributes = {
     pageId: string;
     pageTags?: string[];
-    [k: string]: string | number | boolean | string[];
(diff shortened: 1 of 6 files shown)


src/sessions/PageManager.ts
```diff
@@ -188,9 +188,17 @@ export class PageManager {
...
+                this.attributes.pageTags = customPageAttributes.pageTags;
+            }
+            if (customPageAttributes.pageAttributes) {
+                Object.keys(customPageAttributes.pageAttributes).forEach(
+                    (attribute) => {
+                        this.attributes[attribute] =
+                            customPageAttributes.pageAttributes[attribute];
+                    }
+                );
+            }
```
qhanam, line 201:
```suggestion
```suggestion
        if (customPageAttributes?.pageTags) {
            this.attributes.pageTags = customPageAttributes.pageTags;
        }
        if (customPageAttributes?.pageAttributes) {
            Object.keys(customPageAttributes.pageAttributes).forEach(
                (attribute) => {
                    this.attributes[attribute] =
                        customPageAttributes.pageAttributes[attribute];
                }
            );
        }
```
```

**Comment by qhanam** [comment]
What happens if I add an attribute named `pageTags` to the `pageAttributes` object?

**Comment by limhjgrace** [comment]
> What happens if I add an attribute named `pageTags` to the `pageAttributes` object?

Given implementation, the one in the `pageAttributes` object would override the value provided by `pageTags`

File: src/sessions/PageManager.ts (hunk 1 above)
Comment (limhjgrace) [reply src/sessions/PageManager.ts]: Refactored accordingly

File: src/sessions/PageManager.ts [hunk 2]
```diff
@@ -187,10 +188,18 @@ export class PageManager {
...
+        }
+        if (customPageAttributes?.pageAttributes) {
+            Object.keys(customPageAttributes.pageAttributes).forEach(
+                (attribute) => {
+                    if (!DEFAULT_PAGE_ATTRIBUTES.includes(attribute)) {
+                        this.attributes[attribute] =
+                            customPageAttributes.pageAttributes[attribute];
+                    }
+                }
+            );
```
qhanam, line 202:
```suggestion
            this.attributes = {
                ...customPageAttributes.pageAttributes,
                ...this.attributes
            };
```

qhanam (approved):

PR #237, opened 2022-09-27
chore: Enable typescript strict mode

### LINKED ISSUE #12: Enforce strict type checks with TypeScript
Strict type checks in TypeScript gives stronger guarantees about types, but are not currently enforced.

This task is to (1) update tsconfig.json to include `"strict": true` in `compilerOptions` and (2) fix any TypeScript compiler errors with the new configuration.

qhanam:
The web client does not currently use TypeScript's strict mode, which provides better type safety.

This change enables strict mode for production code (and resolves strict mode type violations). Strict mode remains disabled for test code.

Resolves #12 .

---

### DIFF
--- a/src/index-browser.ts
+++ b/src/index-browser.ts
@@ -12,5 +12,5 @@ if (typeof fetch === 'function' && typeof navigator.sendBeacon === 'function') {
     new CommandQueue().init(window.AwsRumClient);
 } else {
     // eslint-disable-next-line @typescript-eslint/no-empty-function
-    window[window.AwsRumClient.n] = () => {};
+    (window as any)[window.AwsRumClient.n] = () => {};
 }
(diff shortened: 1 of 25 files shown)


src/index-browser.ts
```diff
@@ -12,5 +12,5 @@ if (typeof fetch === 'function' && typeof navigator.sendBeacon === 'function') {
     new CommandQueue().init(window.AwsRumClient);
 } else {
     // tslint:disable-next-line:no-empty
-    window[window.AwsRumClient.n] = () => {};
+    (window as any)[window.AwsRumClient.n] = () => {};
```
adebayor123, line 15: Would this (https://dev.to/zapkub/stop-using-window-as-any-41i1) be beneficial?

adebayor123: Seems (var as any) is commonly used pattern throughout - nevermind!

adebayor123 (approved):
LGTM

PR #253, opened 2022-10-07
fix: Default attributes not overwritten by custom attributes

limhjgrace:
Previously, users could overwrite built-in default attributes. However, this may lead to unexpected behavior in the CW RUM console and poor user experience. This change will ensure that built-int attributes will overwrite any custom attributes if the attribute key is the same.

Changes passed integ tests: https://github.com/limhjgrace/aws-rum-web/pull/1

---

### DIFF
--- a/src/event-cache/EventCache.ts
+++ b/src/event-cache/EventCache.ts
@@ -207,9 +207,9 @@ export class EventCache {
         // objects with their own attribute sets. Instead, we store session
         // attributes and page attributes together as 'meta data'.
         const metaData: MetaData = {
-            version: '1.0.0',
             ...this.sessionManager.getAttributes(),
-            ...this.pageManager.getAttributes()
+            ...this.pageManager.getAttributes(),
+            version: '1.0.0'
         };
 
         this.events.push({
--- a/src/sessions/SessionManager.ts
+++ b/src/sessions/SessionManager.ts
@@ -137,7 +137,7 @@ export class SessionManager {
     public addSessionAttributes(sessionAttributes: {
         [k: string]: string | number | boolean;
     }) {
-        this.attributes = { ...this.attributes, ...sessionAttributes };
+        this.attributes = { ...sessionAttributes, ...this.attributes };
     }
 
     public getUserId(): string {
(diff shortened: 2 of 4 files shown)


src/event-cache/EventCache.ts
```diff
@@ -207,9 +207,9 @@ export class EventCache {
         // objects with their own attribute sets. Instead, we store session
         // attributes and page attributes together as 'meta data'.
         const metaData: MetaData = {
-            version: '1.0.0',
             ...this.sessionManager.getAttributes(),
-            ...this.pageManager.getAttributes()
+            ...this.pageManager.getAttributes(),
+            version: '1.0.0'
```
qhanam, line 212: issue: Why do we set the application version here? Won't the application version always be `1.0.0`, since the last entry takes precedence?

src/sessions/SessionManager.ts
```diff
@@ -137,7 +137,7 @@ export class SessionManager {
     public addSessionAttributes(sessionAttributes: {
         [k: string]: string | number | boolean;
     }) {
-        this.attributes = { ...this.attributes, ...sessionAttributes };
+        this.attributes = { ...sessionAttributes, ...this.attributes };
```
qhanam, line 140: I'm still skeptical about not letting applications overwrite the built-in metadata.

While not permitting overwrites might prevent users from accidentally overwriting built-in metadata, there may be cases where we want to overwrite the built-in metadata.

Likelihood of overwriting accidents  vs likelihood of use cases overwriting.

One example is if one wanted to spoof the domain, since RUM doesn't currently support multiple domains in the app monitor config.

qhanam: We will apply this change as-is, which is to disallow overwriting built-in metadata. This is under the assumption that we can change the behavior to be more permissive in the future.

src/event-cache/EventCache.ts
qhanam: An never mind I missed that we're just moving the the property assignment.

qhanam (approved):

PR #273, opened 2022-10-25
feat: Omit request signature when in proxy mode.

qhanam:
When using a proxy, requests need not be signed by SigV4, however, the web client still requires credentials to dispatch requests. A workaround to this has been to add dummy credentials. 

This change adds a `proxy` config option. When `proxy = true`, the web client will dispatch data without waiting for credentials, and without signing requests.

While we could have simply added dummy credentials as a solution, omitting the SigV4 step entirely will allow us to remove this part of the AWS SDK in the future, reducing the size of the web client.

---

### DIFF
--- a/src/dispatch/__tests__/BeaconHttpHandler.test.ts
+++ b/src/dispatch/__tests__/BeaconHttpHandler.test.ts
@@ -68,7 +68,7 @@ describe('BeaconHttpHandler tests', () => {
         );
     });
 
-    test('when proxy is tue then sendBeacon omits the signature', async () => {
+    test('when proxy is true then sendBeacon omits the signature', async () => {
         // Init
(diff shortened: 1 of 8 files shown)


src/dispatch/__tests__/BeaconHttpHandler.test.ts
```diff
@@ -75,4 +67,20 @@ describe('BeaconHttpHandler tests', () => {
             'https://rumservicelambda.us-west-2.amazonaws.com/appmonitors/application123?X-Amz-Algorithm=AWS4-HMAC-SHA256'
         );
     });
+
+    test('when proxy is tue then sendBeacon omits the signature', async () => {
```
limhjgrace, line 71: nit: `tue` -> `true` ?

limhjgrace (approved):

adebayor123 (approved):

PR #275, opened 2022-10-27
chore: Smoke tests for custom attributes feature

limhjgrace:
Currently, the custom attributes feature doesn't have smoke test to test e2e. This PR will add a test scenario to detect regression.

Smoke tests tested in fork repo release flow: https://github.com/limhjgrace/aws-rum-web/actions/runs/3341540493/jobs/5532996009

---

### DIFF
--- a/src/test-utils/smoke-test-utils.ts
+++ b/src/test-utils/smoke-test-utils.ts
@@ -26,6 +26,27 @@ const builtInAttributes = [
     'pageTags'
 ];
 
+export const customAttributesAddedAtInit = [
+    'customAttributeKeyAtInit1=customAttributeValueAtInit1',
+    'customAttributeKeyAtInit2=customAttributeValueAtInit2',
+    'custom_attribute_key_at_init=customAttributeValueAtInit',
+    'valid:customAttributeKeyAtInit=customAttributeValueAtInit'
+];
+
+export const customAttributesAddedAtRuntime = [
+    'customAttributeKeyAtRuntime1=customAttributeValueAtRuntime1',
+    'customAttributeKeyAtRuntime2=customAttributeValueAtRuntime2',
+    'custom_attribute_key_at_runtime=customAttributeValueAtRuntime',
+    'valid:customAttributeKeyAtRuntime=customAttributeValueAtRuntime'
+];
+
+export const customPageAttributesAddedAtRuntime = [
+    'customPageAttributeKeyAtRuntime1=customPageAttributeValueAtRuntime1',
+    'customPageAttributeKeyAtRuntime2=customPageAttributeValueAtRuntime2',
+    'custom_page_attribute_key_at_runtime=customPageAttributeValueAtRuntime',
+    'valid:customPageAttributeKeyAtRuntime=customPageAttributeValueAtRuntime'
+];
+
 /** Returns filtered events by type */
 export const getEventsByType = (
     requestBody: { RumEvents: any[] },
@@ -145,7 +166,7 @@ export const isEachEventIngested = async (
                     `Expected attributes: ${JSON.stringify(metadataAttributes)}`
                 );
                 console.log(
-                    `Actual metadata: ${JSON.stringify(
+                    `Retrieved metadata: ${JSON.stringify(
                         ingestedEvents.get(eventId)
                     )}`
                 );
@@ -167,6 +188,11 @@ export const isEachEventIngested = async (
                 console.log(
                     `Invalid custom attributes ingested: ${JSON.stringify(
                         invalidAttributes
+                    )}`
+                );
+                console.log(
+                    `Retrieved metadata: ${JSON.stringify(
+                        ingestedEvents.get(eventId)
                     )}`
                 );
                 throw new Error(`Invalid custom attributes ingested`);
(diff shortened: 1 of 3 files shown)


adebayor123:
suggestion: I think the 4 different cases you've listed can be summed up to a single E2E case, which is: `can events with custom attributes be ingested?`. The 4 different scenarios should be covered in unit/integ test cases.

src/test-utils/smoke-test-utils.ts
```diff
@@ -101,5 +153,50 @@ export const isEachEventIngested = async (
         if (!ingestedEvents.has(eventId)) {
             throw new Error(`Event ${eventId} not ingested.`);
         }
+
+        // validate custom attributes
+        if (metadataAttributes) {
```
qhanam, line 158: issue: This method is difficult to read due to its length and complexity. I suggest implementing the steps to retrieve the events, and then verify the events, as separate functions. See comment in the `verifyIngestionWithRetry` function.

src/test-utils/smoke-test-utils.ts
```diff
@@ -56,7 +104,8 @@ export const verifyIngestionWithRetry = async (
                 rumClient,
                 eventIds,
                 timestamp,
-                monitorName
+                monitorName,
+                metadataAttributes
             );
             return true;
```
qhanam, line 110: suggestion: Refactor this method so that each step (getting data and verifying data) is implemented as a separate function. We could also further decompose the verification step.
```javascript
const actual: Map<string, string[]> = await getIngestedEvents(
    rumClient,
    eventIds,
    timestamp,
    monitorName,
    metadataAttributes
);
return expectEvents(actual)
    && expectValidAttributes(actual)
    && expectInvalidAttributes(actual);
```

src/test-utils/smoke-test-utils.ts
```diff
@@ -101,5 +153,50 @@ export const isEachEventIngested = async (
...
+                console.log(
+                    `Expected attributes: ${JSON.stringify(metadataAttributes)}`
+                );
+                console.log(
+                    `Retrieved metadata: ${JSON.stringify(
+                        ingestedEvents.get(eventId)
+                    )}`
+                );
+                throw new Error(`Did not find expected metadata attribute(s).`);
+            }
```
qhanam, line 174: thought: I wonder if we could use *expect* expressions instead of  logging to console and throwing errors. Playwright's `expect` is simple but could work, or we could use a library like [expect](https://www.npmjs.com/package/expect).

Non-blocking. The code as-implemented is consistent with the existing code and works fine.

src/test-utils/smoke-test-utils.ts
```diff
@@ -96,10 +141,58 @@ export const isEachEventIngested = async (
...
+
+/** Returns true when no invalid custom attributes are found */
+export const expectInvalidAttributes = async (
+    ingestedEvents: Map<string, string[]>,
+    eventIds: string[],
+    metadataAttributes: string[] | undefined = undefined
+) => {
+    if (metadataAttributes) {
+        eventIds.forEach((eventId) => {
+            const invalidAttributes = ingestedEvents
```
adebayor123, line 184: nit: `eventsWithInvalidAttributes` would be more accurate IMO

src/test-utils/smoke-test-utils.ts
```diff
@@ -96,10 +141,58 @@ export const isEachEventIngested = async (
...
+};
+
+/** Returns true when no invalid custom attributes are found */
+export const expectInvalidAttributes = async (
+    ingestedEvents: Map<string, string[]>,
+    eventIds: string[],
+    metadataAttributes: string[] | undefined = undefined
+) => {
+    if (metadataAttributes) {
+        eventIds.forEach((eventId) => {
```
adebayor123, line 183: question: Do we need this iteration? If the ingestedEvents has events with invalid metadata, it will always be there without having to go through each event. It will make this a lot simpler

src/test-utils/smoke-test-utils.ts
```diff
@@ -40,25 +71,38 @@ export const isDataPlaneRequest = (response, targetUrl) => {
...
-            await isEachEventIngested(
+            const actual: Map<string, string[]> = await getIngestedEvents(
                 rumClient,
-                eventIds,
                 timestamp,
                 monitorName
             );
-            return true;
+
+            return (
```
adebayor123, line 93: nit: Could we refactor this so that the boolean is returned by a single method, and the single method contains the three different methods checking for 1) whether event is ingested, 2) whether valid attributes are there, 3) whether invalid attributes are there?

adebayor123: Later on, if we introduce another condition to check, the original `verifyRetryWithIngestion` will get longer and longer. Having a dedicated method such as `verifyIngestion` that calls those three different methods will make it look a lot cleaner in my opinion - also adding logs to indicate which of the checks fail will be beneficial for debugging

src/test-utils/smoke-test-utils.ts
limhjgrace: No, technically only need to check metadata of one event. Good catch! Will remove for loop here and above when checking for valid attributes.

adebayor123: What I meant was you could just do filtering on `ingestedEvents` without having to do extra for loop on `eventIds` array :P

qhanam (approved):

adebayor123 (approved):

PR #280, opened 2022-11-02
fix: Populate method from Request Info

codermpl:
- When in createHttpEvent, if there is no method on the RequestInit, then try to get the method from the RequestInfo

---


qhanam (approved):
lgtm; thank you!

PR #314, opened 2022-11-29
feat: Create a session ID when cookies are disabled

ps863:
Presently, session ID is not created in all instances.  Without a session ID, we are unable link events from the same page together. 

This change will allow a session ID to be created in all instances. This allows all events tied to a single page to be linked together. 

---

### DIFF
--- a/src/sessions/__tests__/SessionManager.test.ts
+++ b/src/sessions/__tests__/SessionManager.test.ts
@@ -127,7 +127,7 @@ describe('SessionManager tests', () => {
         expect(sessionManager.getSession().sessionId).toEqual(sessionId);
     });
 
-    test('when cookies are not allowed and existing session does not exist then getSession returns a new session', async () => {
+    test('when cookies are not allowed then getSession returns a new session', async () => {
         // Init
         const config = {
             ...DEFAULT_CONFIG,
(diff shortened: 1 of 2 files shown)


src/sessions/__tests__/SessionManager.test.ts
```diff
@@ -221,7 +268,7 @@ describe('SessionManager tests', () => {
         expect(sessionManager.getSession()).toBeTruthy();
     });
 
-    test('when cookies are disabled after being enabled then getSession returns the nil session', async () => {
+    test('when cookies are disabled after being enabled and session not expired then getSession returns the existing session', async () => {
```
qhanam, line 271:
```suggestion
    test('when cookies are disabled after being enabled then getSession returns the existing session', async () => {
```

src/sessions/__tests__/SessionManager.test.ts
```diff
@@ -127,7 +127,7 @@ describe('SessionManager tests', () => {
         expect(sessionManager.getSession().sessionId).toEqual(sessionId);
     });
 
-    test('when cookies are not allowed then getSession returns the nil session', async () => {
+    test('when cookies are not allowed and existing session does not exist then getSession returns a new session', async () => {
```
qhanam, line 130:
```suggestion
    test('when cookies are not allowed then getSession returns a new session', async () => {
```

qhanam (approved):

limhjgrace:
> This change will allow a session ID to be created in all instances. This allows all events tied to a single page to be linked together.

This could inflate sessionCount where individual pages in one session would be considered as separate unique sessions. Not sure if we need to document this anywhere?

limhjgrace (approved):

PR #321, opened 2022-12-11
feat: webclient version and installation method

limhjgrace:
Currently, AWS RUM has no way to determine how RUM users are installing the web client or the version of the installed web client. This update will add this information to the metadata of all events. The web client version will be stored in a new built-in field, `aws:clientVersion`, while the installation method will be stored in a new built-in field, `aws:client`. 

This update also updates the deployment script to track all official web client versions in a file located in the S3 bucket in the AWS RUM service account (`content/versions.csv`).

---


.github/scripts/deploy.sh
```diff
@@ -3,6 +3,11 @@
 version=$(npm pkg get version | sed 's/"//g')
 bucket=$1
 
+versionsFile="../versions.txt";
+echo $version >> $versionsFile
+
+allVersions=$(<$versionsFile)
```
qhanam, line 9: issue: These changes to *versions.txt* won't persist.

Instead, could we read the current list of versions from s3 (i.e., `s3://$bucket/content/versions`)? This way the deployment bucket maintains the historic list of versions.

.github/scripts/deploy.sh
```diff
@@ -26,3 +32,8 @@ then
     aws s3api put-object --bucket $bucket --key "content/$patchUpdate/LICENSE-THIRD-PARTY" --body LICENSE-THIRD-PARTY --cache-control max-age=7200
     aws s3api put-object --bucket $bucket --key "content/$patchUpdate/LICENSE" --body LICENSE --cache-control max-age=7200
 fi
+
+echo "1 min bake time"
+sleep 60
+
+echo "Completed deployment"
```
qhanam, line 39: issue: I think the intent here is to wait until the backend has fetched the new `versions` file before deploying the version, in which case this should be at line 22.

However, I don't think we should add this sleep for the following reasons:
1. I think `current` and `versions` should only contain versions which are available (for example, if some automation reads the value of `current`, but that version of `cwr.js` is not available, it could break the automation).
2. Sleeping for one minute wastes compute time in the CI/CD pipeline.

In practice, I doubt the delay will have significant impact on monitoring the health of the web client, and if it does, we can re-visit this.

.github/scripts/deploy.sh
```diff
@@ -13,6 +18,7 @@ aws s3api put-object --bucket $bucket --key "content/$version/LICENSE" --body LI
 if [[ $version =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
 then
     echo $version | aws s3 cp - s3://$bucket/content/current
+    echo "$allVersions" | aws s3 cp - s3://$bucket/content/versions
```
qhanam, line 21: issue: To prevent race conditions, I think we should update the `current` and `versions` files after deploying the new version (i.e., immediately before the `fi` statement at line 34).

src/event-cache/__tests__/EventCache.integ.test.ts
```diff
@@ -50,7 +59,8 @@ describe('EventCache tests', () => {
 
         const eventCache: EventCache = Utils.createEventCache(config);
         const expectedMetaData = {
-            version: '1.0.0',
+            version: WEB_CLIENT_VERSION,
```
qhanam, line 62: issue: Should this be `aws:clientVersion`?

src/orchestration/Orchestration.ts
```diff
@@ -228,7 +228,8 @@ export class Orchestration {
         {
             cookieAttributes: configCookieAttributes,
             ...partialConfig
-        }: PartialConfig = {}
+        }: PartialConfig = {},
+        isFromScript?: boolean | undefined
```
qhanam, line 232: suggestion: Instead of adding a new parameter, use the `config` object.

Options to consider:
1. Add `"aws:client": "arw-module"` to the `sessionAttributes` object.
2. Add a new field called `client` to the config object. Set the default value of `client` to `arw-module`.

(1) would require having the application-provided `aws:` attributes overwrite the default `aws:` attributes, which I still think we should do. The advantage of (1) is that we don't need to add any new fields to the `config` object.

qhanam: suggestion (alternate): Instead of adding a new parameter, use the build script to differentiate between the two at build time.

For example, during build time we could (a) set an environment variable containing the client name and version, or (b) generate a file containing the client name and version.

I'm hesitant to add a new parameter to the constructor.

src/test-utils/test-utils.ts
```diff
@@ -58,20 +58,30 @@ export const PUT_RUM_EVENTS_REQUEST: PutRumEventsRequest = {
     ]
 };
 
+export const WEB_CLIENT_VERSION = '2.0.0';
```
qhanam, line 61: question: This is unused. Are you planning on adding the version info in a future PR?

src/orchestration/Orchestration.ts
limhjgrace: Added new field `client` to config object with default value set to `arw-module`

src/event-schemas/meta-data.json
```diff
@@ -63,7 +62,10 @@
         "domain": {
             "type": "string"
         },
-        "pageTags": { "type": "array", "items": { "type": "string" } }
+        "pageTags": { "type": "array", "items": { "type": "string" } },
+        "aws:client": {
```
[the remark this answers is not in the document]
limhjgrace: Yes, will add

src/test-utils/test-utils.ts
limhjgrace: Refactored

.github/scripts/deploy.sh
limhjgrace: Agree, removed.

.github/scripts/deploy.sh
limhjgrace: Refactored

.github/scripts/deploy.sh
limhjgrace: Refactored

.github/scripts/deploy.sh
```diff
@@ -5,13 +5,59 @@ bucket=$1
...
+        delete_local_file
+
+        
+        echo ERROR: INVALID VERSION
+        echo ${singlePairArr[1]}
+        exit 1
+    fi
+    
+done
+
```
qhanam, line 52: question: What are we trying to achieve here?

If the goal is to verify the file, I recommend doing so in a separate script, as a separate step, after the new file has been uploaded. The script would (1) download the from S3, (2) verify its correctness, and (3) fail the build.

Although such a step seems like overkill, so I'm OK with having it or not.

src/event-cache/EventCache.ts
```diff
@@ -8,6 +8,10 @@ import {
     UserDetails,
     RumEvent
 } from '../dispatch/dataplane';
+import { INSTALL_SCRIPT, INSTALL_MODULE } from '../utils/constants';
+
+import packageInfo from '../../package.json';
```
qhanam, line 13: issue: package.json is pretty large to be including in the JS bundle. Is there a way we could automatically create a minimal 'version.json' file at build time?

limhjgrace: Using the `GenerateJsonPlugin` webpack plugin, was able to create a separate `version.json` file at build time

.github/scripts/deploy.sh
limhjgrace: Given we decided to remove the check from DP, I wanted to make sure we verify the file is formatted as expected during web client release. Will refactor to do in as separate script as a separate step.

update-version.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+version=$(jq -r '.version' package.json)
+jq --arg version $version '.version |= $version' version.json >file.json.tmp && cp file.json.tmp version.json
```
qhanam, line 4: question: Why do we need to write to `file.json.tmp` first?

package.json
```diff
@@ -29,7 +29,7 @@
         "build:dist:es": "tsc --project ./tsconfig.es.json",
         "build:dist:cjs": "tsc --project ./tsconfig.cjs.json",
         "build:assets": "npm-run-all -p build:dist:* build:webpack:prod",
-        "build": "npm-run-all -s build:schemas build:assets",
+        "build": "npm-run-all -s build:schemas update:version build:assets",
```
qhanam, line 32: suggestion: If we run any of the targets independently, like `npm run build:dist:es`, then the correct version may not be used.

This is fine for the deployment script in this repository, because it runs `build`, but there is are scenarios where this step could be missed. For example, a third party runs their own web client build, but runs only the subset they need: `build:schemas && build:dist:es`.

One way to help mitigate this would be to omit `version.json` from version control (by adding it to `.gitignore`). This way, like the schemas, `build:version` would need to run before the other build steps, helping ensure the file is up to date.

package.json
```diff
@@ -57,7 +57,8 @@
         "smoke:local": "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT NAME=$NAME VERSION=$npm_package_version npx playwright test --config=playwright.local.config.ts --headed",
         "smoke": "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT NAME=$NAME VERSION=$npm_package_version npx playwright test --config=playwright.config.ts --headed",
         "smoke:headless": "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT NAME=$NAME VERSION=$npm_package_version npx playwright test --config=playwright.config.ts",
-        "prepare": "husky install"
+        "prepare": "husky install",
+        "update:version": "bash update-version.sh"
```
qhanam, line 61: nit: Rename `build:version` for consistency.

qhanam (approved):

update-version.sh
limhjgrace: If we write directly to the `version.json` file, it appends onto the content instead of overwriting the file so we have to write to a temp file and then copy the file over to overwrite the existing file.

package.json
limhjgrace: Added to `.gitignore`

package.json
limhjgrace: Renamed

update-version.sh
limhjgrace: Given we're ignoring the version.json file and removed file from changes, I updated the script so that we just write to a new file.

PR #327, opened 2022-12-12
feat:referrer

ps863:
Currently, AWS CloudWatch RUM does not support referrer. Information about the referrer for page views is not presently available. 

This change adds the ability for the RUM web client to record and dispatch referrer information. The referrer will be stored in the `referrer` field in event details of a page view event. The referrer's domain will be stored in the `referrerDomain` field in event details of a page view event.

---

### DIFF
--- a/src/sessions/PageManager.ts
+++ b/src/sessions/PageManager.ts
@@ -8,6 +8,8 @@ export type Page = {
     pageId: string;
     interaction: number;
     parentPageId?: string;
+    referrer?: string | null;
+    referrerDomain?: string | null;
     start: number;
 };
 
@@ -129,6 +131,11 @@ export class PageManager {
             pageId,
             parentPageId: resumed.pageId,
             interaction: resumed.interaction + 1,
+            referrer:
+                document.referrer !== undefined && document.referrer !== ''
+                    ? document.referrer
+                    : null,
+            referrerDomain: this.getDomainFromReferrer(),
             start: Date.now()
         };
         this.resumed = undefined;
@@ -165,6 +172,11 @@ export class PageManager {
             pageId,
             parentPageId: currentPage.pageId,
             interaction: currentPage.interaction + 1,
+            referrer:
+                document.referrer !== undefined && document.referrer !== ''
+                    ? document.referrer
+                    : null,
+            referrerDomain: this.getDomainFromReferrer(),
             start: startTime
         };
     }
@@ -173,6 +185,11 @@ export class PageManager {
         this.page = {
             pageId,
             interaction: 0,
+            referrer:
+                document.referrer !== undefined && document.referrer !== ''
+                    ? document.referrer
+                    : null,
+            referrerDomain: this.getDomainFromReferrer(),
             start: Date.now()
         };
     }
@@ -220,6 +237,14 @@ export class PageManager {
                 pageViewEvent.parentPageInteractionId =
                     page.parentPageId + '-' + (page.interaction - 1);
             }
+
+            if (page.referrer !== null) {
+                pageViewEvent.referrer = page.referrer;
+
+                if (page.referrerDomain !== null) {
+                    pageViewEvent.referrerDomain = page.referrerDomain;
+                }
+            }
         }
 
         return pageViewEvent;
@@ -234,5 +259,22 @@ export class PageManager {
      */
     private useCookies() {
         return navigator.cookieEnabled && this.config.allowCookies;
+    }
+
+    /*
+    Parses the domain from the referrer, if it is available
+    */
+    private getDomainFromReferrer() {
+        if (document.referrer !== undefined || document.referrer !== '') {
+            const domainName = document.referrer.match(
+                '([A-Za-z]+://[^/]+)[/]?'
+            );
+
+            if (domainName) {
+                return domainName[1];
+            }
+        } else {
+            return null;
+        }
     }
 }
--- a/src/sessions/PageManager.ts
+++ b/src/sessions/PageManager.ts
@@ -17,8 +17,8 @@ export type Attributes = {
     parentPageId?: string;
     interaction?: number;
     pageTags?: string[];
-    referrer?: string;
-    referrerDomain?: string;
+    'aws:referrer'?: string;
+    'aws:referrerDomain'?: string;
     // The value types of custom attributes are restricted to the types: string | number | boolean
     // However, given that pageTags is a string array, we need to include it as a valid type
     // Events will be verified by our service to validate attribute value types where
@@ -194,8 +194,10 @@ export class PageManager {
                 this.attributes.parentPageId = page.parentPageId;
             }
             if (document.referrer !== undefined) {
-                this.attributes.referrer = document.referrer;
-                this.attributes.referrerDomain = this.getDomainFromReferrer();
+                this.attributes['aws:referrer'] = document.referrer;
+                this.attributes[
+                    'aws:referrerDomain'
+                ] = this.getDomainFromReferrer();
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/sessions/__tests__/PageManager.test.ts
```diff
@@ -487,3 +488,86 @@ describe('PageManager tests', () => {
         expect(pageManager.getPage().start).toEqual(2500);
     });
 });
+
+test('when complete referrer is available from the DOM', async () => {
```
limhjgrace, line 492: suggestion: `when complete referrer is available from the DOM` --> `when complete referrer is available from the DOM then is recorded in page view event`

src/sessions/__tests__/PageManager.test.ts
```diff
@@ -487,3 +488,86 @@ describe('PageManager tests', () => {
...
+        pageId: '/console/home'
+    });
+
+    window.removeEventListener(
+        'popstate',
+        (pageManager as any).popstateListener
+    );
+});
+
+test('when only domain level referrer is available from the DOM', async () => {
```
limhjgrace, line 521: suggestion: `when only domain level referrer is available from the DOM` --> `when only domain level referrer is available from the DOM then is recorded in page view event`

src/sessions/__tests__/PageManager.test.ts
```diff
@@ -487,3 +488,86 @@ describe('PageManager tests', () => {
...
+        pageId: '/console/home'
+    });
+
+    window.removeEventListener(
+        'popstate',
+        (pageManager as any).popstateListener
+    );
+});
+
+test('when referrer is not available from the DOM', async () => {
```
limhjgrace, line 549: suggestion: `when referrer is not available from the DOM` --> `when referrer from the DOM is not valid then it is not recorded in page view event`

src/sessions/PageManager.ts
```diff
@@ -235,4 +251,32 @@ export class PageManager {
...
+            const domainName = document.referrer.match(
+                '([A-Za-z]+://[^/]+)[/]?'
+            );
+
+            if (domainName) {
+                return domainName[1];
+            }
+        } else {
+            return null;
+        }
```
qhanam, line 269:
```suggestion
        try {
            return new URL(document.referrer).hostname;
         } catch (e) {
             return "";
         }
```

src/sessions/PageManager.ts
```diff
@@ -220,6 +228,14 @@ export class PageManager {
...
                     page.parentPageId + '-' + (page.interaction - 1);
             }
+
+            if (page.referrer !== null) {
+                pageViewEvent.referrer = page.referrer;
+
+                if (page.referrerDomain !== null) {
+                    pageViewEvent.referrerDomain = page.referrerDomain;
+                }
+            }
```
qhanam, line 238:
```suggestion
            pageViewEvent.referrer = document.referrer;
            pageViewEvent.referrerDomain = getDomainFromReferrer()
```

src/sessions/PageManager.ts
```diff
@@ -235,4 +260,21 @@ export class PageManager {
...
+    */
+    private getDomainFromReferrer() {
+        if (document.referrer !== undefined || document.referrer !== '') {
+            const domainName = document.referrer.match(
+                '([A-Za-z]+://[^/]+)[/]?'
+            );
+
+            if (domainName) {
+                return domainName[1];
+            }
```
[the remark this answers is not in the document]
ps863: 1. Yes. However, will be using URL interface instead of regex for this. (See next revision) 
2. Excluded in next revision 
3. Next revision use of URL interface should handle localhost as well

src/sessions/PageManager.ts
```diff
@@ -235,4 +251,32 @@ export class PageManager {
...
+    /*
+    Get the referrer, if it can be read from the DOM
+    */
+    private getReferrer() {
+        if (document.referrer !== undefined && document.referrer !== '') {
+            return document.referrer;
+        } else {
+            return null;
+        }
+    }
```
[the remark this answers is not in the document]
ps863: fixed in next revision. The helper function is not needed

src/sessions/PageManager.ts
ps863: added in next revision

src/sessions/PageManager.ts
ps863: added in next revision

src/sessions/__tests__/PageManager.test.ts
```diff
@@ -487,3 +488,97 @@ describe('PageManager tests', () => {
...
+
+    // Assert
+    expect(pageManager.getPage()).toMatchObject({
+        pageId: '/console/home'
+    });
+
+    expect(pageManager.getAttributes()).toMatchObject({
+        'aws:referrer': '',
+        'aws:referrerDomain': ''
+    });
```
limhjgrace, line 578: question: do we want to record in metadata if it's empty? is this a conscience decision we're making?

src/sessions/__tests__/PageManager.test.ts
```diff
@@ -487,3 +488,97 @@ describe('PageManager tests', () => {
...
+
+    window.removeEventListener(
+        'popstate',
+        (pageManager as any).popstateListener
+    );
+});
```
limhjgrace, line 584: question: do we need a unit test for when `localhost` is the referrer?

src/sessions/__integ__/PageManager.test.ts
```diff
@@ -1,6 +1,7 @@
 import { Selector } from 'testcafe';
 import { REQUEST_BODY } from '../../test-utils/integ-test-utils';
 import { PAGE_VIEW_EVENT_TYPE } from '../../plugins/utils/constant';
+import { create } from 'lodash';
```
adebayor123, line 4: nit: unused

src/sessions/__tests__/PageManager.test.ts
adebayor123: https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer
States that  -`The value is an empty string if the user navigated to the page directly (not through a link, but, for example, by using a bookmark). `, so we should record it

adebayor123 (approved):

src/sessions/__tests__/PageManager.test.ts
ps863: handled in next revision

src/sessions/__integ__/PageManager.test.ts
ps863: removed in next revision

limhjgrace (approved):

src/sessions/PageManager.ts
```diff
@@ -235,4 +251,19 @@ export class PageManager {
...
+    */
+    private getDomainFromReferrer() {
+        try {
+            return new URL(document.referrer).hostname;
+        } catch (e) {
+            if (document.referrer === 'localhost') {
+                // Handle special case for localhost
+                return 'localhost';
+            }
+            return '';
```
qhanam, line 266:
```suggestion
            return document.referrer === 'localhost' ? document.referrer : '';
```

src/sessions/PageManager.ts
```diff
@@ -235,4 +246,15 @@ export class PageManager {
...
+    /*
+    Parses the domain from the referrer, if it is available
+    */
+    private getDomainFromReferrer() {
+        try {
+            return new URL(document.referrer).hostname;
+        } catch (e) {
+            return document.referrer === 'localhost' ? document.referrer : '';
+        }
+    }
```
qhanam, line 259: nit: Consider covering these (localhost and empty referrer) with unit tests.

qhanam (approved):

PR #341, opened 2023-01-12
feat:Time Spent on Page

ps863:
Currently, AWS CloudWatch RUM does not record the time spent on a certain page. This change records the time spent on a page as part of the `parentTimeSpentOnPage` field in the page view event details. This field records the time spent on the parent page to the page view event it is recorded in.

---

### DIFF
--- a/src/__integ__/cookiesEnabled.test.ts
+++ b/src/__integ__/cookiesEnabled.test.ts
@@ -27,18 +27,8 @@ test('when page is re-loaded with cookies enabled, session start is not dispatch
         .click(clear)
         .eval(() => location.reload());
 
-    await t
-        .wait(300)
-        .click(dispatch)
-        .expect(REQUEST_BODY.textContent)
-        .contains('BatchId');
-
-    const jsonB = JSON.parse(await REQUEST_BODY.textContent);
-    const sessionStartEventsB = jsonB.RumEvents.filter(
-        (e) => e.type === SESSION_START_EVENT_TYPE
-    );
-
-    await t.expect(sessionStartEventsB.length).eql(0);
+    // No new events should be recorded, thus no request body
+    await t.wait(300).click(dispatch).expect(REQUEST_BODY.textContent).eql('');
 });
 
 test('when page is loaded with cookies enabled, session start includes meta data', async (t: TestController) => {
--- a/src/sessions/PageManager.ts
+++ b/src/sessions/PageManager.ts
@@ -48,10 +48,11 @@ export class PageManager {
     private config: Config;
     private record: RecordEvent;
     private page: Page | undefined;
-    private resumed: Page | undefined;
+    private resumed: boolean;
     private attributes: Attributes | undefined;
     private virtualPageLoadTimer: VirtualPageLoadTimer;
     private TIMEOUT = 1000;
+    private timeOnParentPage: number | undefined;
 
     /**
      * A flag which keeps track of whether or not cookies have been enabled.
@@ -65,7 +66,7 @@ export class PageManager {
         this.config = config;
         this.record = record;
         this.page = undefined;
-        this.resumed = undefined;
+        this.resumed = false;
         this.recordInteraction = false;
         this.virtualPageLoadTimer = new VirtualPageLoadTimer(
             this,
@@ -82,13 +83,12 @@ export class PageManager {
         return this.attributes;
     }
 
-    public resumeSession(pageId: string, interaction: number) {
+    public resumeSession(page: Page | undefined) {
         this.recordInteraction = true;
-        this.resumed = {
-            pageId,
-            interaction,
-            start: 0
-        };
+        if (page) {
+            this.page = page;
+            this.resumed = true;
+        }
     }
 
     public recordPageView(payload: string | PageAttributes) {
@@ -103,9 +103,7 @@ export class PageManager {
             this.recordInteraction = true;
         }
 
-        if (!this.page && this.resumed) {
-            this.createResumedPage(pageId, this.resumed);
-        } else if (!this.page) {
+        if (!this.page) {
             this.createLandingPage(pageId);
         } else if (this.page.pageId !== pageId) {
             this.createNextPage(this.page, pageId);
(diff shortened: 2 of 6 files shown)


src/sessions/PageManager.ts
```diff
@@ -165,6 +167,7 @@ export class PageManager {
             pageId,
             parentPageId: currentPage.pageId,
             interaction: currentPage.interaction + 1,
+            parentTimeSpentOnPage: timeSpentOnParentPage,
```
adebayor123, line 170: suggestion: `parentTimeSpentOnPage` -> `timeSpentOnParentPage`

src/sessions/PageManager.ts
```diff
@@ -8,6 +8,7 @@ export type Page = {
     pageId: string;
     interaction: number;
     parentPageId?: string;
+    parentTimeSpentOnPage?: number;
```
adebayor123, line 11: suggestion: `parentTimeSpentOnPage` -> `timeSpentOnParentPage`

adebayor123:
thought: Since we cannot retrieve the last page's duration and we decided to not create a new form of page view, what about sending a page view event with `pageId: EndVisit` or something similar to indicate that this was the last page to visit? I believe we have beacon to send events in the case of browser closing (feel free to correct me if I'm wrong), and if we can detect a user closing the browser/page and utilize the beacon, we would capture the last page's duration.

ps863:
> thought: Since we cannot retrieve the last page's duration and we decided to not create a new form of page view, what about sending a page view event with `pageId: EndVisit` or something similar to indicate that this was the last page to visit? I believe we have beacon to send events in the case of browser closing (feel free to correct me if I'm wrong), and if we can detect a user closing the browser/page and utilize the beacon, we would capture the last page's duration.

Had thought about this. From what I know, I see three main issues with this approach: 

1) In order to send this last page view event on user closing browser, we would need to listen for the [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) or [unload](https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event) event. Listening for both events is known to be unreliable (not supported by certain browsers) and can cause performance issues. 


2) Following (1), since the existing page view event is recorded on route change and unload event is fired whenever user navigates away from a page, we would not be able to distinguish and would end up duplicating page view events recorded (as a unload event occurs with every navigation between pages and so does a route change).
We could have transition to capturing page view event exclusively when the page is navigated away from/closed by relying on the unload event, but this requires sendBeacon to work reliably. (refer to (3) on why it does not) 

3) To flush the event cache reliably on the user closing the page, we need to use sendBeacon. However, as noted [here](https://github.com/aws-observability/aws-rum-web/pull/194), we noticed ad blockers were preventing sendBeacon from functioning correctly. Therefore, users have an option to configure web client to send events using fetch or sendBeacon (both approaches have their downsides). 
If fetch is used, a situation can arise wherein the page has not fully unloaded before the fetch is completed and events are dispatched to the dataplane. As such any page view event that relies on listening for page unload may not get dispatched.

src/sessions/PageManager.ts
ps863: agree. Added to the revision

adebayor123:
question: How do we handle the case where the session is picked up from the cookie? For example, if user left the browser at page B, the web client may hold the "page.start" value in the cache, and when you move to subsequent pages, it may return an incorrect `timeSpendOnParentPage` information. If `page.start` is not stored in the cookie, what if the page manager's `page.start` is `undefined`? How do we handle this case?

src/sessions/PageManager.ts
```diff
@@ -136,6 +137,7 @@ export class PageManager {
 
     private createNextPage(currentPage: Page, pageId: string) {
         let startTime = Date.now();
+        const timeOnParentPage = startTime - currentPage.start;
```
qhanam, line 140: suggestion: `timeOnParentPage` is more concise. Use `timeOnParentPage` instead of `timeSpentOnParentPage`.

src/sessions/PageManager.ts
```diff
@@ -165,6 +167,7 @@ export class PageManager {
             pageId,
             parentPageId: currentPage.pageId,
             interaction: currentPage.interaction + 1,
+            timeSpentOnParentPage: timeOnParentPage,
```
qhanam, line 170: question: What happens when the page is resumed?

We may want to change the behavior so that page view events are not recorded for resumed pages. RUM users have been confused by seeing users navigate from "PageA" to "PageA" in user journeys.

src/sessions/PageManager.ts
```diff
@@ -8,6 +8,7 @@ export type Page = {
     pageId: string;
     interaction: number;
     parentPageId?: string;
+    timeSpentOnParentPage?: number;
```
qhanam, line 11: question: Is there a way we can omit `timeOnParentPage` from this structure? This is stored on browser storage, so if we don't need it, we should omit it.

I feel like it's something we can compute during `recordPageView()`.

ps863: Fixed in revision. Now its done during ```recordPageView```

src/sessions/PageManager.ts
ps863: Handled this case in revision. Page view events are no longer recorded for resumed pages if page ID is same. Start time from the Session cookie is used to compute the time on page if resumed page is different than when it was when session stopped.

src/sessions/PageManager.ts
ps863: updated in revision and schemas

ps863:
> question: How do we handle the case where the session is picked up from the cookie? For example, if user left the browser at page B, the web client may hold the "page.start" value in the cache, and when you move to subsequent pages, it may return an incorrect `timeSpendOnParentPage` information. If `page.start` is not stored in the cookie, what if the page manager's `page.start` is `undefined`? How do we handle this case?



Handled this in latest revision. The following change should handle the concerns you had: 

- If the resumed page Id is same as the page id stored when session temporarily stopped, then we don't dispatch a new Page View Event. Only the PageManager state is updated. The start time is updated as the time stored in the cookie (ie the start time is not reset like it was previously, and therefore the correct start time is recorded). Therefore when user moves to a new page, we will use the correct start time to compute time on page.

src/__integ__/cookiesEnabled.test.ts
```diff
@@ -27,18 +27,8 @@ test('when page is re-loaded with cookies enabled, session start is not dispatch
...
-        .contains('BatchId');
-
-    const jsonB = JSON.parse(await REQUEST_BODY.textContent);
-    const sessionStartEventsB = jsonB.RumEvents.filter(
-        (e) => e.type === SESSION_START_EVENT_TYPE
-    );
-
-    await t.expect(sessionStartEventsB.length).eql(0);
+    // No new events should be recorded, thus no request body
+    await t.wait(300).click(dispatch).expect(REQUEST_BODY.textContent).eql('');
```
qhanam, line 31: question: I'm curious why did this test start failing? And why is this correct behavior now?

qhanam (approved):

limhjgrace (approved):

PR #348, opened 2023-01-24
chore: Fix windows+edge integ tests.

qhanam:
Windows integ tests are currently broken because TestCafe is timing out waiting for browser connections.

This change fixes the breakage by:
1. Only running edge browser integ tests in windows
2. Using headless for windows integ tests
3. Increasing concurrency for windows integ tests from one to three
4. Increasing browserInitTimeout to six minutes

This change also refactors the WebVitalsPlugin integ test.

---


src/plugins/event-plugins/__integ__/WebVitalsPlugin.test.ts
```diff
@@ -17,24 +17,6 @@ fixture('WebVitalEvent Plugin').page(
 // "FID is not reported if the user never interacts with the page."
 // It doesn't seem like TestCafe actions are registered as user interactions, so cannot test FID
 
-const removeUnwantedEvents = (json: any) => {
```
ps863, line 20: question: The changes to this file (`WebVitalsPlugin.test.js`) look good to me. However, are these related to fixing the integ tests? If so, maybe we should include in PR description for future reference.

package.json
```diff
@@ -42,7 +42,8 @@
         "preinteg": "npm run build:dev",
         "integ": "testcafe --app \"http-server ./build/dev -s\"",
         "integ:browser": "npm run integ -- --config-file=./.testcafe.browser.js",
-        "integ:headless": "npm run integ -- --config-file=./.testcafe.headless.js",
+        "integ:headless-linux": "npm run integ -- --config-file=./.testcafe.headless-linux.js",
+        "integ:headless-windows": "npm run integ -- --config-file=./.testcafe.headless-windows.js",
```
limhjgrace, line 46: nit: We can update our internal runbooks regarding the changes in how we can run the integ tests.

limhjgrace (approved):

src/plugins/event-plugins/__integ__/WebVitalsPlugin.test.ts
qhanam: You are correct, I am pushing an unrelated change with this PR, which is generally bad practice.

The test was failing, so I refactored it while debugging. It turned out to be something else, however since the change improves the test's maintainability and is small, I didn't want to open a second PR.

ps863 (approved):

PR #350, opened 2023-01-27
Docs/configs

limhjgrace:
Updating config docs as some existing entries are outdated/incorrect and there are new entries to add.
---


docs/configuration.md
```diff
@@ -21,8 +21,8 @@ For example, the config object may look similar to the following:
...
-| cookieAttributes | [CookieAttributes](#cookieattributes) | `{ domain: window.location.hostname, path: '/', sameSite: 'Strict', secure: true } ` | Cookie attributes are applied to all cookies stored by the web client, including `cwr_s` and `cwr_u`. |
-| sessionAttributes | [MetadataAttributes](#metadataattributes) | `{ color: 'blue' }` | Session attributes will be added the metadata of all events in the session.|
+| cookieAttributes | [CookieAttributes](#cookieattributes) | `{ domain: window.location.hostname, path: '/', sameSite: 'Strict', secure: true, unique: false } ` | Cookie attributes are applied to all cookies stored by the web client, including `cwr_s` and `cwr_u`. |
```
adebayor123, line 24: question: What does setting `unique` do? Is that something the users can configure?

docs/configuration.md
```diff
@@ -36,10 +36,14 @@ For example, the config object may look similar to the following:
...
 | sessionSampleRate | Number | `1` | The proportion of sessions that will be recorded by the web client, specified as a unit interval (a number greater than or equal to 0 and less than or equal to 1). When this field is `0`, no sessions will be recorded. When this field is `1`, all sessions will be recorded. |
 | signing | Boolean | true | When this field is `true`, the web client signs [RUM data](https://docs.aws.amazon.com/cloudwatchrum/latest/APIReference/API_PutRumEvents.html) using [SigV4](https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html).<br/><br/>When this field is `false`, the web client does not sign [RUM data](https://docs.aws.amazon.com/cloudwatchrum/latest/APIReference/API_PutRumEvents.html).<br/><br/>Set this field to `false` only when sending RUM data to CloudWatch RUM through an unauthenticated proxy. This field **must be `true`** when sending RUM data directly to CloudWatch RUM. |
 | telemetries | [Telemetry Config Array](#telemetry-config-array) | `[]` | See [Telemetry Config Array](#telemetry-config-array) |
+| batchLimit | Number | `100` | The maximum number of events that will be sent in one batch of RUM events. |
```
adebayor123, line 43: thought: I do agree these are configurable values but do we really want to recommend configuring these values?

docs/configuration.md
limhjgrace: The details are included in the `CookieAttributes` section below: "When this field is `false`, the session cookie name is `cwr_s`. When this field is `true`, the session cookie name is `cwr_s_[AppMonitor Id]`"

Yes, users can configure.

docs/configuration.md
limhjgrace: While I don't recommend configuring these values, there have been some users who asked about the `batchLimit` value so I wanted to include it in the docs so that they can see what the default value is.

adebayor123 (approved):

Issue #371, opened 2023-03-01
bug: Cannot find module 'aws-rum-web' or its corresponding type declarations.


Naes0:
Hi, I'm getting the following error when importing AwsRum and AwsRumConfig on v1.13.2:

**Type error: Cannot find module 'aws-rum-web' or its corresponding type declarations.**

Workaround has been to downgrade to v1.12.0.
Very weird because I can see them being exported in the index.d.ts in node_modules. :/

Anyone else have this problem?

targumon:
Yes, seeing this too.
Maybe related to #363? ("fix: Transpile @aws-sdk to ES5.")

ivanpepelko:
Additionally: this happens when trying to `bump aws-rum-web from 1.12.0 to 1.13.2`

with `1.13.2` a workaround is this:
```typescript
import { AwsRum } from 'aws-rum-web/dist/es/src';
```

... which is something I don't want to do. It seems import root is missing index file.

qhanam:
We've [deprecated v1.3.2](https://www.npmjs.com/package/aws-rum-web/v/1.13.2?activeTab=versions) in NPM, and are working on a patch.

qhanam:
Patched v1.13.3 is [available on NPM](https://www.npmjs.com/package/aws-rum-web?activeTab=versions).

Naes0:
Just tried v1.13.3, got the following error:

**TypeError: Cannot read properties of undefined (reading 'version')
 at Object.<anonymous> (/app/node_modules/aws-rum-web/dist/cjs/src/event-cache/EventCache.js:19:47)**

looks like it trying to access a 'default' property here which causes it to fail:
`var webClientVersion = version_json_1.default.version;`

qhanam:
@Naes0 what build system and module  (i.e., es/cjs) are you using?

Naes0:
@qhanam We're using Next.js 13 (not turbopack) so webpack 5 and cjs on the server and esm in the browser :)

PR #381, opened 2023-03-07
fix: Hardcode webclient version

limhjgrace:
Applications using cjs run into a `TypeError` due to import issues involving the `version.json` file. Related issue: https://github.com/aws-observability/aws-rum-web/issues/371

As a temporary fix, we will hardcode the webclient version.

---


ps863:
question: does "resolveJsonModule": true remain true if we are hardcoding the web client version?

limhjgrace:
> question: does "resolveJsonModule": true remain true if we are hardcoding the web client version?

has been removed

ps863 (approved):

qhanam:
Related to #381

qhanam (approved):

PR #391, opened 2023-03-17
cd: Create direct CDN deployment and NPM deprecation workflow

adebayor123:
## Description
This PR introduces two new workflows to help with the deprecation process for both CDN and NPM releases.
For CDN:
- Previous method involved re-releasing the previous stable version to overwrite the 1.x versions. This had to rely on the existing CD workflow, which triggers smoke test and NPM release as well.
- Instead, this PR creates a workflow dedicated to just re-releasing with a previous stable version without the smoke test or the NPM release. This helps reduce the time to deprecate for CDN releases. Verified in GAMMA stage.
- To use this workflow, the contributor will create a release branch of a previously safe commit, trigger this workflow.

For NPM:
- Previously to deprecate, we relied on manual work. This PR creates a workflow that requires two inputs, 1) deprecating version and 2) deprecating reason. With these inputs, the workflow will automatically deprecate the NPM version.
- Tested with https://www.npmjs.com/package/aws-rum-web-experimental/v/1.9.0-alpha.0 

---


limhjgrace:
question: In our rollback steps to deprecate CDN, the runbook indicates that we need to "Revert the affected major (M.x) and minor (M.m.x) auto-update versions to the previous good version by deleting the bad (most recent) version." Do we not have to still delete the bad (most recent) version? Or does deploying the previous stable version do this under the hood?

limhjgrace:
> question: In our rollback steps to deprecate CDN, the runbook indicates that we need to "Revert the affected major (M.x) and minor (M.m.x) auto-update versions to the previous good version by deleting the bad (most recent) version." Do we not have to still delete the bad (most recent) version? Or does deploying the previous stable version do this under the hood?

Synced offline that the intended behavior should be to only release previous stable version to update 1.x or 1.x.x, not remove the bad version. LGTM.

.github/workflows/cdn_release.yaml
```diff
@@ -0,0 +1,51 @@
+name: AWS RUM CDN Release
```
ps863, line 1: nit: Should we name this workflow something that informs us its meant for deprecation (eg - AWS RUM CDN - For Deprecation) ? 
The name right now, might make it seem this is a workflow we run to do a normal release for CDN but the normal release process is the same (using AWS Rum Web Client release workflow)

ps863 (approved):

qhanam:
Regarding the CDN workflow, is there any disadvantage to re-running the smoke tests? I feel like doing that validation after deployment can only be a good thing.

What do you think about merging these two scripts into a single "deprecate" workflow?

.github/workflows/cdn_release.yaml
qhanam: If it's going to be a separate workflow, I agree we should be clear that it is to be used only for rolling back the 'live' CDN distributions.

adebayor123: Changed the workflow name to `Rollback CDN Live Version`

qhanam (approved):

PR #395, opened 2023-03-23
cd: smoke tests for CJS and ES modules

ps863:
This change introduces the following: 

- Sample smoke test application that uses ES Modules
- Sample smoke test application that uses CJS module
- Automation in existing release workflow that adds 
    - Steps to build sample applications via NPM installation method (pre and post release)
    - Steps to update and upload the sample applications to AWS resources to perform testing (no new resources are created. Existing resources are used) 
    - Steps to perform existing smoke tests on CJS and ES module applications before release on NPM 
- All smoke test apps moved to new subdirectory - smoke. This keeps it separate from other apps

There are a few known issues: 
- 2 Tests are failing for ES and CJS smoke tests. This is due to inability to access the AwsRum object to create custom attributes and create a Dom event. This will be fixed in a future update. Operators should use the applications manually and exercise the behavior until the automation works correctly.
- Tests are lot more flaky and overall release process will take longer. Due to previous point, the timeout for ES and CJS tests has been set to 10 minutes. This will be removed once the application is fixed to access AwsRum. Additionally, more changes will be made to improve parallelization that would reduce overall flakiness. 
- Credentials may become invalid, so we retrieve them after every smoke test 

[Sample run of workflow
](https://github.com/ps863/aws-rum-web/actions/runs/4498683511/jobs/7915603316)

---


qhanam:
> 2 Tests are failing for ES and CJS smoke tests

Have you removed these, or do they cause the release workflow to fail?

> Operators should use the applications manually and exercise the behavior until the automation works correctly.

Does running these tests in the ES/CJS smoke tests add significant protection against regressions? If they do, we should automate it now instead of introducing a manual step. If they don't, we can omit them from the ES/CJS smoke tests.

> Due to previous point, the timeout for ES and CJS tests has been set to 10 minutes.

Do you mean because of the *“inability to access the AwsRum object”*? How could we fix this?

.github/workflows/cd.yaml
```diff
@@ -82,19 +106,80 @@ jobs:
...
+                  INSTALL_METHOD: 'CDN'
               run: npm run smoke:headless
+              if: always()
 
             - name: Publish to NPM
               run: npm publish
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+            - name: Build Smoke Test Application - NPM/ES (Post NPM Release)
```
adebayor123, line 171: question: I understand the idea behind this, but is it necessary? Technically if the packed ES/CJS tests succeed, wouldn't NPM most likely succeed as well?

adebayor123: question: Based on the sample workflow you've provided, it seems we continued to run all three different types of smoke tests even if the ES one failed - did the workflow stop at NPM because it is a gamma stage and the credentials were wrong, or is it because the tests have failed?

adebayor123:
> 2 Tests are failing for ES and CJS smoke tests. This is due to inability to access the AwsRum object to create custom attributes and create a Dom event. This will be fixed in a future update. Operators should use the applications manually and exercise the behavior until the automation works correctly.

Is it the application unable to access AwsRum object? If it's not accessible, shouldn't it apply to all types of tests? Could you clarify further?

I also agree with Quinn that we should hold off on merging this until we fix the test for full automation - are we tight on deadline?

> Tests are lot more flaky and overall release process will take longer. Due to previous point, the timeout for ES and CJS tests has been set to 10 minutes. This will be removed once the application is fixed to access AwsRum. Additionally, more changes will be made to improve parallelization that would reduce overall flakiness.

Is the timeout increased due to the flakiness or is it due to the issue of not being able to access AwsRum object?

> Credentials may become invalid, so we retrieve them after every smoke test

This is somewhat expected - should we consider using the `Duration` option to increase the overall duration of a credential? https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html

ps863:
> Have you removed these, or do they cause the release workflow to fail?

No the tests are still there and not removed. Reasoning of retaining the test is primarily to be aware that they are failing when running in gamma. 

The workflow, however, will still continue running the next set of tests and not cause the entire workflow to stop. 


> Does running these tests in the ES/CJS smoke tests add significant protection against regressions? If they do, we should automate it now instead of introducing a manual step. If they don't, we can omit them from the ES/CJS smoke tests.

I believe the other smoke tests are already testing that the package loads correctly and ingestion is occurring. Most regressions could happen at the build step or initialization stage. We are testing that web client initializes correctly, by checking for a 200 response from dataplane. We also build the application itself prior to this step. 


Additionally, CDN smoke tests all are functioning and would still catch a regression with the actual code in this case. 


> Do you mean because of the _“inability to access the AwsRum object”_? How could we fix this?

Yes. The fix here is to identify how the library can be accessed from the web bundle javascript that smoke.html loads so that the custom attributes and DOM event can be recorded (`eg - awsRum.registerDomEvents() or awsRum.addSessionAttributes`)

.github/workflows/cd.yaml
ps863: > question: I understand the idea behind this, but is it necessary? Technically if the packed ES/CJS tests succeed, wouldn't NPM most likely succeed as well?

You are correct that theoretically its unlikely there would be an issue. 

This is more of a sanity check that the publishing process is done correctly and replicates the behavior seen when installing the package locally during development. It decreases the overall time to mitigation if the release had an issue or for some reason NPM install is not working correctly

.github/scripts/build_npm_applications.sh
```diff
@@ -0,0 +1,46 @@
+PRE_OR_POST_RELEASE=$1
+MODULE_TYPE=$2
+
+
+if [ "$PRE_OR_POST_RELEASE" = "PRE" ]; then
```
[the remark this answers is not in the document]
ps863: Described in below [comment](https://github.com/aws-observability/aws-rum-web/pull/395#discussion_r1146683292)

limhjgrace:
Are we running the tests sequentially e.g. once for CDN, once for ES, once for CJS? Do we want to/can we parallelize this to reduce the increase in runtime? But this would require additional infra so may not be worth the effort.

ps863:
> Are we running the tests sequentially e.g. once for CDN, once for ES, once for CJS? Do we want to/can we parallelize this to reduce the increase in runtime? But this would require additional infra so may not be worth the effort.

Agree. Makes sense to do at a later time. Right now, priority is to have 
- Capability to check if new releases build correctly before release 
- Verify that web client initialization does not have any regressions by checking if we can send and verify events

adebayor123 (approved):
We can always improve in further iterations. lgtm

playwright.config.ts
```diff
@@ -2,22 +2,42 @@
...
+            trace: 'on-first-retry'
+        },
+        projects: [
+            {
+                name: 'chromium',
+                use: { ...devices['Desktop Chrome'] }
+            }
+        ]
+    };
+}
```
qhanam, line 41:
```suggestion
var config = {
    forbidOnly: !!process.env.CI,
    reporter: 'list',
    workers: process.env.CI ? 4 : undefined,
    testDir: process.env.INSTALL_METHOD === 'CDN' ?
        'src/__smoke-test__' :
        'src/__smoke-test-npm__',
    retries: process.env.CI ? 2 : 2,
    timeout: 300000,
    use: {
        trace: 'on-first-retry'
    },
    projects: [
        {
            name: 'chromium',
            use: { ...devices['Desktop Chrome'] }
        }
    ]
};
```

package.json
```diff
@@ -57,7 +57,7 @@
         "smoke:local:headless": "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT NAME=$NAME VERSION=$npm_package_version npx playwright test --config=playwright.local.config.ts",
         "smoke:local": "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT NAME=$NAME VERSION=$npm_package_version npx playwright test --config=playwright.local.config.ts --headed",
         "smoke": "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT NAME=$NAME VERSION=$npm_package_version npx playwright test --config=playwright.config.ts --headed",
-        "smoke:headless": "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT NAME=$NAME VERSION=$npm_package_version npx playwright test --config=playwright.config.ts",
+        "smoke:headless": "cross-env URL=$URL MONITOR_ID=$MONITOR ENDPOINT=$ENDPOINT NAME=$NAME VERSION=$npm_package_version INSTALL_METHOD=$INSTALL_METHOD npx playwright test --config=playwright.config.ts",
```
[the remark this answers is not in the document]
ps863: done

smoke/smoke-test-application-NPM-CJS/package.json
```diff
@@ -0,0 +1,34 @@
...
+        "copy-webpack-plugin": "^6.3.2",
+        "ts-loader": "^9.2.6",
+        "typescript": "^4.4.3",
+        "webpack": "^5.58.1",
+        "webpack-dev-middleware": "^5.2.1",
+        "webpack-dev-server": "^4.3.1"
+    },
+    "dependencies": {
+        "@babel/runtime": "^7.16.0",
+        "aws-rum-web": "^1.13.6",
```
qhanam, line 27: issue: I think this should be something like `*.*.*` ... here and in the other ES application.

.github/scripts/build_npm_applications.sh
```diff
@@ -0,0 +1,46 @@
...
+        echo "Not a valid module type"
+    fi
+elif [ "$PRE_OR_POST_RELEASE" = "POST" ]; then
+
+    if [ "$MODULE_TYPE" = "NPM-ES" ]; then  
+        cd smoke/smoke-test-application-NPM-ES
+        npm uninstall aws-rum-web
+        npm run clean 
+        npm install 
+        npm install aws-rum-web #Install latest released direct from NPM
```
qhanam, line 30: nit: We already installed aws-rum-web in the previous `npm install` step. Here and elsewhere.

.github/scripts/build_npm_applications.sh
```diff
@@ -0,0 +1,46 @@
...
+        npm run clean 
+        npm install 
+        npm install /home/runner/work/aws-rum-web/aws-rum-web #Install locally 
+        npm run build
+    elif [ "$MODULE_TYPE" = "NPM-CJS" ]; then  
+        cd smoke/smoke-test-application-NPM-CJS
+        npm uninstall aws-rum-web
+        npm run clean 
+        npm install 
+        npm install $(npm pack /home/runner/work/aws-rum-web/aws-rum-web | tail -1) #Install locally
```
qhanam, line 18: question: Why does this command differ from the `NPM-ES` build?

.github/scripts/build_npm_applications.sh
qhanam: OK I see what you're doing here. Since we've specified the actual version of the web client (i.e., `1.13.6` instead of `*`), we need to run a separate `npm install aws-rum-web`.

qhanam (approved):

.github/scripts/build_npm_applications.sh
ps863: This needs to be done as `npm install <path to local>` creates a symlink to the directory instead of actually copying the files over. When used with webpack and babel, if there is only a symlink, it's known to be problematic when the web bundle is created and the dependencies are resolved. The solution therefore is to use `npm pack ` which will copy over the files to the directory and install them. 

With the CJS build, I kept facing issues using the `npm install <path to local>` option. It was not  an issue with ES build (unclear to me why the difference exists). But therefore, there are two different commands being used. I believe the ES build will work with `npm pack` as well. 


Reference: 

https://stackoverflow.com/questions/51845893/error-importing-local-npm-package
https://stackoverflow.com/questions/21425980/npm-install-without-symlinks-option-not-working

.github/scripts/build_npm_applications.sh
ps863: Yes `npm install ` should get the latest released rum web client version. This just double confirms the behavior and if the new version builds correctly, before ending the release flow. 
I also npm uninstall to completely ensure the installation is that of the latest version and there is nothing left from the last version

.prettierignore
```diff
@@ -6,3 +6,4 @@ src/events
 package-lock.json 
 docs/
 .vscode
+smoke
```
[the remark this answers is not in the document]
ps863: removed. Was added during testing

smoke/smoke-test-application-NPM-CJS/src/loader-npm-rum.ts
```diff
@@ -0,0 +1,30 @@
...
+    awsRum: AwsRum = new AwsRum(
+        APPLICATION_ID,
+        APPLICATION_VERSION,
+        APPLICATION_REGION,
+        config
+    );
+} catch (error) {
+    // Ignore errors thrown during CloudWatch RUM web client initialization
+    console.log(error);
+}
```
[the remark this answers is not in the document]
ps863: agree. Edited to re-throw the error

PR #412, opened 2023-05-30
feat: enable recordAllRequests by default 

williazz:
### **_Breaking change_**

All RUM HTTP events will now be published by default, instead of just those with errors. This is breaking because the default cost will increase as more RUM events get ingested. Users may opt out of this by setting `trackAllRequests` to false. 

This change is necessary because we will provide observability into the latency of HTTP requests initiated by Fetch and Xhr. In consequence, non-error status codes like 2xx need to be published so that customers can know if they are meeting their SLAs. 

These changes will be staged in `release/2.0.x` first before merging into main with other features. 

---


src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
```diff
@@ -527,7 +527,10 @@ describe('XhrPlugin tests', () => {
         plugin.disable();
 
         // Assert
-        expect(record).not.toHaveBeenCalled();
```
adebayor123, line 530: question: did this change due to the default behavior change?

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
```diff
@@ -568,7 +571,10 @@ describe('XhrPlugin tests', () => {
...
```
adebayor123, line 571: question: if the session is not being recorded, shouldn't `record` not be invoked at all?

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
williazz: Sorry for the late response. `record(type, event)` can record any event, including both XrayTrace and HttpEvent. Here, I expect an HttpEvent to be recorded because it is enabled by default, even though trace is disabled. I do not expect `record(XRAY_TRACE_EVENT_TYPE, ...)` to be recorded

qhanam:
An open question is whether we will have a major version bump or a minor version bump for this change.

Unless an app monitor has explicitly set `recordAllRequests`, this change may cause additional HTTP events to be recorded and exported to CloudWatch RUM.

*Do we consider this change backwards compatible, or is it a breaking change?*

williazz:
@qhanam Using this [breaking change policy](https://learn.microsoft.com/en-us/linkedin/shared/breaking-change-policy), I'd argue that this change qualifies as "changes to the intended functionality of an endpoint". So I'd advocate for a major version upgrade

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
williazz: I changed the logic to check for session recording inside Xhr before reading this comment. I can change it back :/

qhanam:
Re. the commit *"fix: if session is not recorded, then Fetch and Xhr do not record any"*

My understanding is that [the current behavior is correct](https://github.com/aws-observability/aws-rum-web/pull/412#discussion_r1224555467). That is, if a session is not being recorded, then EventCache will prevent all events from being recorded (including Fetch and XHR).

Is my understanding correct? If it is, I don't think we need this change.

qhanam:
> @qhanam Using this [breaking change policy](https://learn.microsoft.com/en-us/linkedin/shared/breaking-change-policy), I'd argue that this change qualifies as "changes to the intended functionality of an endpoint". Even though these changes can be opted out, the intended functionality is so different that I think it deserves a major version upgrade

This seems reasonable. We may want to wait to merge this until we are ready to start building major version 2.

williazz:
> Is my understanding correct? If it is, I don't think we need this change.

Yes that's correct. Reverted

williazz:
> suggestion: https://google.github.io/eng-practices/review/developer/cl-descriptions.html

updated

PR #416, opened 2023-06-09
chore: Run smoke tests in pre-prod before publishing

qhanam:
The release workflow does not currently test the release in a pre-prod environment before executing the release. This means that (1) the release technician must manually test the release in a pre-prod environment before executing a release, and (2) there is an increased risk of a regression.

This change adds steps to the release workflow to test the release in a pre-prod environment before running the release.

This change adds the following three secrets secrets to the release environment.
1. `ROLE_GAMMA`
2. `BUCKET_GAMMA`
3. `CDN_GAMMA`

I've verified the new release workflow succeeds in pre-prod up to the NPM publishing step, which is unchanged. See https://github.com/qhanam/aws-rum-web/actions/runs/5226500122/jobs/9437220189

---

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -19,18 +19,6 @@ jobs:
                   node-version: '16.x'
                   registry-url: 'https://registry.npmjs.org'
 
-            - name: Fetch AWS Credentials for Deployment
-              run: |
-                  export AWS_ROLE_ARN=${{ secrets.ROLE }}
-                  export AWS_WEB_IDENTITY_TOKEN_FILE=/tmp/awscreds
-                  export AWS_DEFAULT_REGION=us-east-1
-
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yaml
```diff
@@ -116,9 +104,28 @@ jobs:
               run: npm run smoke:headless
               timeout-minutes: 10
 
-            - name: Fetch AWS Credentials for Smoke Test
```
ps863, line 119: In a previous change, we would fetch the credentials before every smoke test because the CJS and ES smoke tests are long running and creds used to time out before test finished. 

If we are removing the fetch before every test, we may want to increase the duration that the temporary credentials are valid when they are initially fetched.

ps863 (approved):

limhjgrace (approved):

.github/workflows/cd.yaml
qhanam: Fixed -- we now re-fetch credentials before running each smoke test (ES, CJS and CDN).

PR #417, opened 2023-06-09
feat: measure latency of HTTP events

williazz:
# Why

As a single page app owner, I rely on AJAX requests to fetch data in the background so that I may present a different view without refreshing. In order to insist on the highest standard, I need observability into the performance of Fetch and Xhr requests, which is best measured by latency. 

# Changes

* Merged with #412, which includes breaking change of enabling `trackAllRequests` by default
* Fetch and Xhr Plugins use `Date.now()` timestamps to measure latency
* Schema change to HttpEvent, includes required fields `startTime` and `duration`
* Timestamps are shared between Trace and Http Events
* Performance API was NOT used in this PR because PerformanceResourceTiming entries and monkey patched HttpEvents can only be mapped by request URL, which will have collisions on repeated requests. Collisions can be resolved by creation order, but a stress test is needed to investigate to what extent incorrect mappings actually occur. The [PRT implementation](https://github.com/williazz/aws-rum-web/blob/AjaxLatencyPRT/src/plugins/HttpPlugin.ts#LL13C23-L13C33) will be released later as apart of another feature, once tested. 

# Tests
```
 PASS  src/sessions/__tests__/SessionManager.test.ts
 PASS  src/plugins/event-plugins/__tests__/DomEventPlugin.test.ts
 PASS  src/dispatch/__tests__/RetryHttpHandler.test.ts
 PASS  src/dispatch/__tests__/EnhancedAuthentication.test.ts
 PASS  src/plugins/__tests__/plugins.test.ts
 PASS  src/orchestration/__tests__/Orchestration.test.ts
 PASS  src/event-cache/__tests__/EventCache.test.ts
 PASS  src/dispatch/__tests__/StsClient.test.ts
 PASS  src/event-cache/__tests__/EventCache.integ.test.ts
 PASS  src/dispatch/__tests__/DataPlaneClient.test.ts
 PASS  src/dispatch/__tests__/BeaconHttpHandler.test.ts
 PASS  src/dispatch/__tests__/CognitoIdentityClient.test.ts
 PASS  src/sessions/__tests__/PageManager.test.ts
 PASS  src/dispatch/__tests__/Dispatch.test.ts
 PASS  src/dispatch/__tests__/Authentication.test.ts

Test Suites: 28 passed, 28 total
Tests:       366 passed, 366 total
Snapshots:   0 total
Time:        21.958 s
```

---


src/plugins/MonkeyPatched.ts
```diff
@@ -15,8 +15,17 @@ export abstract class MonkeyPatched<
     Nodule extends object,
     FieldName extends keyof Nodule
 > extends InternalPlugin {
-    public enable = this.patch.bind(this, true);
-    public disable = this.patch.bind(this, false);
+    constructor(pluginId: string) {
+        super(pluginId);
+    }
+
+    enable() {
```
williazz, line 22: Enable / disable were refactored from properties to methods. This way they can be overriden while still having access to the originale methods via `super.enable/disable`

src/plugins/event-plugins/XhrPlugin.ts
```diff
@@ -122,6 +123,14 @@ export class XhrPlugin extends MonkeyPatched<XMLHttpRequest, 'send' | 'open'> {
...
     }
 
+    /** Updates endTime if empty and returns it in seconds */
+    private fillEndTimeIfEmpty(xhrDetails: XhrDetails) {
+        if (!xhrDetails.endTime) {
+            xhrDetails.endTime = Date.now();
+        }
+        return xhrDetails.endTime / 1000;
+    }
+
```
limhjgrace, line 133: suggestion: Seems like we've combined a getter and setter method into one method. Can we clarify what singular action this method should do and refactor accordingly? Also want to note that we're setting end time with `ms` as the unit but returning the value in seconds which may be confusing at the first glance. Could we unify to use one unit and do the unit conversion outside this method?

src/plugins/event-plugins/XhrPlugin.ts
```diff
@@ -203,25 +212,31 @@ export class XhrPlugin extends MonkeyPatched<XMLHttpRequest, 'send' | 'open'> {
...
+        }
     };
 
     private handleXhrTimeoutEvent = (e: Event) => {
         const xhr: XMLHttpRequest = e.target as XMLHttpRequest;
         const xhrDetails = this.xhrMap.get(xhr);
-        const errorName = 'XMLHttpRequest timeout';
-        this.handleXhrDetailsOnError(xhrDetails, errorName);
+        if (xhrDetails) {
+            this.fillEndTimeIfEmpty(xhrDetails);
```
limhjgrace, line 226: question: Do we need this refactoring? I see that we have the same check at line 236-237. Is this to get the most accurate end time?

src/plugins/event-plugins/XhrPlugin.ts
```diff
@@ -203,25 +212,31 @@ export class XhrPlugin extends MonkeyPatched<XMLHttpRequest, 'send' | 'open'> {
     private handleXhrAbortEvent = (e: Event) => {
         const xhr: XMLHttpRequest = e.target as XMLHttpRequest;
         const xhrDetails = this.xhrMap.get(xhr);
-        const errorName = 'XMLHttpRequest abort';
-        this.handleXhrDetailsOnError(xhrDetails, errorName);
+        if (xhrDetails) {
+            this.fillEndTimeIfEmpty(xhrDetails);
```
limhjgrace, line 216: question: Do we need this refactoring? I see that we have the same check at line 236-237. Is this to get the most accurate end time?

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
```diff
@@ -88,7 +91,7 @@ describe('XhrPlugin tests', () => {
         plugin.disable();
 
         // Assert
-        expect(record).toHaveBeenCalledTimes(1);
+        expect(record).toHaveBeenCalledTimes(2);
```
limhjgrace, line 94: question: Why do we expect this to now be called twice?

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
```diff
@@ -609,7 +618,7 @@ describe('XhrPlugin tests', () => {
         expect(record).toHaveBeenCalledTimes(1);
     });
 
-    test('when recordAllRequests is false then the plugin does record a request with status OK', async () => {
+    test('when recordAllRequests is true then the plugin records a request with status OK', async () => {
```
limhjgrace, line 621: suggestion: It appears that we only updated the test name without changing the test content. Do we want to update the test to add validations?

src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -258,6 +260,13 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
         this.context.record(HTTP_EVENT_TYPE, httpEvent);
     };
 
+    /** Manually updates the http duration and returns the end time in seconds */
+    private updateDurationManually(httpEvent: HttpEvent): number {
+        const endTime = Date.now();
+        httpEvent.duration = endTime - httpEvent.startTime!;
+        return endTime / 1000;
+    }
```
limhjgrace, line 268: suggestion: Seems like we've combined a getter and setter method into one method. Can we clarify what singular action this method should do and refactor accordingly? Also want to note that we're setting end time with `ms` as the unit but returning the value in seconds which may be confusing at the first glance. Could we unify to use one unit and do the unit conversion outside this method?

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
limhjgrace: Never mind, I see your comment in the other PR: https://github.com/aws-observability/aws-rum-web/pull/412/files#r1210735142

limhjgrace (commented):
In the PR description, you mention that this is merged with https://github.com/aws-observability/aws-rum-web/pull/412. Will we still need the other PR or will we close that in favor of having this PR include all changes?

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
williazz: oh good catch, i think it accidentally got lost when resolving merge conflicts

williazz: On second look, I believe this change was to fix a typo. `recordAllRequests` is true in the configuration but was described to be false in the test name

src/plugins/event-plugins/FetchPlugin.ts
williazz: Changes:
* Specificed ms / sec in all timestamp vars for clarity 
* Removed `setDurationManually()`

My intention behind `setDurationManually()` was code reuse, but I can see how it broke SOLID and was confusing.

docs/configuration.md
```diff
@@ -135,7 +135,7 @@ telemetries: [
 | urlsToInclude | RegExp[] | `[/.*/]` | A list of HTTP request (`XMLHttpRequest` or `fetch`) URLs. These requests will be recorded, unless explicitly excluded by `urlsToExclude`. |
 | urlsToExclude | RegExp[] | `[]` | A list of HTTP request (`XMLHttpRequest` or `fetch`) URLs. These requests will not be recorded. |
 | stackTraceLength | Number | `1000 ` | The number of characters to record from a JavaScript error's stack trace (if available). |
-| recordAllRequests | boolean | `false` | By default, only HTTP failed requests (i.e., those with network errors or status codes which are not 2xx) are recorded. When this field is `true`, the http telemetry will record all requests, including those with successful 2xx status codes. <br/><br/>This field does **does not apply** to X-Ray traces, where all requests are recorded. |
+| recordAllRequests | boolean | `true` | By default, all HTTP requests are recorded. When this field is `false`, http telemetry will only record failed requests (i.e., those with network errors or status codes which are not 2xx). <br/><br/>This field does **does not apply** to X-Ray traces, where all requests are recorded. |
```
qhanam, line 138: issue: PRs should be [*"one self-contained change"*](https://google.github.io/eng-practices/review/developer/small-cls.html).

I believe that the best practice when using [GitHubFlow](https://gitversion.net/docs/reference/modes/mainline) is to merge #412 into a branch until we are preparing to release the next major version.

I've created a release branch for the next major revision -- `release/2.0.x`. We can merge #412 into that branch and rebase this change on top of it. You'll need to update the target branch for #412 before it can be merged in.

src/plugins/event-plugins/XhrPlugin.ts
```diff
@@ -244,11 +259,14 @@ export class XhrPlugin extends MonkeyPatched<XMLHttpRequest, 'send' | 'open'> {
         xhr: XMLHttpRequest
     ) {
         if (this.config.recordAllRequests || !this.statusOk(xhr.status)) {
-            this.context.record(HTTP_EVENT_TYPE, {
+            const httpEvent: HttpEvent = {
```
qhanam, line 262: nit: There are a lot of minor stylistic changes in this CR (e.g., creating a temporary variables, renaming variables, re-arranging imports). These changes don't add value to the package, and make it more difficult to review.

Also keep in mind that saving bytes matters in this package. The packager/minimizer can only do so much, so adding additional statements and long variable names can increase the size of the package.

docs/configuration.md
williazz: Target branch for #412 has been updated to release/2.0.x

src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -227,7 +227,9 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
                     : request.method
                     ? request.method
                     : 'GET'
-            }
+            },
+            startTime: Date.now(),
+            duration: 0
```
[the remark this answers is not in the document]
williazz: * `startTime` is used and shared. 
* `duration` was just a default value back when these fields were required, and I forgot to delete this initialization after we made them optional

updated by deleting the duration here

src/plugins/event-plugins/XhrPlugin.ts
williazz: * updated to remove `fillEndTimeIfEmpty()`
* moved unit conversion out as well

src/plugins/event-plugins/XhrPlugin.ts
williazz: yes. date.now() must be called as early as possible

src/plugins/event-plugins/XhrPlugin.ts
williazz: updated to remove temp vars and renamings

src/plugins/event-plugins/XhrPlugin.ts
williazz: same as the other

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
williazz: Yeah sorry these PR's shouldn't have been combined. I wanted to keep the 2.0 changes together. Luckily, Quinn suggested to just set the target branch to `release/2.0.x`

limhjgrace (approved):

williazz:
I have rebased `release/2.0.x` onto this PR so that #412 is not longer included in these changes

williazz:
In [fix: remove redundant trace latency tests](https://github.com/aws-observability/aws-rum-web/pull/417/commits/114201331cfd37ee66e7bb5476996ce5c2a840d6), I realized that unit tests to check if trace events capture latency are not needed. `start_time` and `end_time` are required on trace events, so previous unit tests already provide enough coverage.

williazz:
> question: In the PR description, you mention that this is merged with #412. Will we still need the other PR or will we close that in favor of having this PR include all changes?

Yeah we'll keep these separate! #412 has been merged into `release/2.0.x`. I've rebased those changes onto this PR, so that #417 will only contain the latency changse.

When we are ready to release 2.0, we will do a merge fast forward of 2.0 onto main. 

> note: I see the commit history is a bit long, please make sure we squash before merging!

💀

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
```diff
@@ -905,4 +907,115 @@ describe('FetchPlugin tests', () => {
...
+        // Assert
+        expect(record).toHaveBeenCalledTimes(1);
+        const eventType = record.mock.calls[0][0];
+        const httpEvent = record.mock.calls[0][1] as HttpEvent;
+        expect(eventType).toEqual(HTTP_EVENT_TYPE);
+        expect(httpEvent).toEqual(
+            expect.objectContaining({
+                response: expect.anything(),
+                startTime: expect.any(Number),
+                duration: expect.any(Number)
```
qhanam, line 930: issue: Verify that startTime and duration are correct here.

The same applies to the test at line 938, and the tests in *XhrPlugin.test.ts*.

qhanam (approved):

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
```diff
@@ -820,4 +823,133 @@ describe('XhrPlugin tests', () => {
...
+        expect(httpEvent).toEqual(
+            expect.objectContaining({
+                response: expect.anything(),
+                startTime: expect.any(Number),
+                duration: expect.any(Number)
+            })
+        );
+        // expected results depend on mockNow()
+        expect(httpEvent.startTime).toEqual(0);
+        expect(httpEvent.duration).toEqual(2000);
```
williazz, line 858: Xhr plugin makes an extra call to Date.now(), so the mocked time is a second later

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
williazz: Updated to check for specific values. However, this introduces some brittleness because of the way I wrote `mockNow()`. The expected results will be affected by the number of calls to the mocked `Date.now()`.

PR #419, opened 2023-06-15
feat: resource event filtering 

williazz:
# Why (updated)
Users currently have no way to filter out unwanted performance events. One story is that customers are receiving a lot of noisy resource events from chrome extensions. 

# Changes
Users can filter resource events by supplying a `ignore(entry: PerformanceEntry)` to PerformanceConfig. This enables PerformanceObserver to ignore specified PerformanceResourceTiming entries, which are used to create RUM ResourceEvents.

In addition, `ignore()` also configures the NavigationPlugin, which uses a PerformanceObserver to create NavigationEvents. Here, customers can look for PerformanceNavigationTiming entries and ignore them if need be. 

The default configuration is that only resource events with non-http URL schemas are ignored. This will ignore resource events from browser extensions, which typically follow non-http URL schemas (e.g. "chrome-extension://*"). 

# Revisions
Revision 4:
* [fix: remove performance mock data helpers](https://github.com/aws-observability/aws-rum-web/pull/419/commits/b8d07e10e5f724f71a3c9b638b9a40603659fd67)
* [fix: remove unecessary defaultIgnore() tests](https://github.com/aws-observability/aws-rum-web/pull/419/commits/7c9f8db5595baa60d9de6a7dfbff9d9168153c94)

Revision 3:
* add unit tests for subclasses of PerformanceEntry to verify that only resource events with non-http(s) url schema are ignored by default
* moved unit tests to performance-utils

Revision 2:
* `ignore()` receives PerformanceEntry
* `ignore()` is defined in PerformanceConfig and shared between NavigationPlugin and ResourcePlugin. Note: only resource level 2 events are subject to ignoring
* Added unit test to ignore navigation events
* Removed version bump for a separate PR
* Ignore chrome extension events by default
* Rename ignoreEvent() to ignore()
* Remove unrelated refactors

---


CHANGELOG.md
```diff
@@ -2,6 +2,29 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [1.14.0](https://github.com/aws-observability/aws-rum-web/compare/v1.12.0...v1.14.0) (2023-06-21)
```
qhanam, line 5: issue: Use separate PRs to prepare a new release version.

The best practice is to use a separate review process for each unrelated change. This (1) makes it easier to review the change, (2) ensures issues with one change don't block the review/release process for another, and (3) makes it easy to revert changes.

docs/configuration.md
```diff
@@ -138,6 +138,26 @@ telemetries: [
 | recordAllRequests | boolean | `false` | By default, only HTTP failed requests (i.e., those with network errors or status codes which are not 2xx) are recorded. When this field is `true`, the http telemetry will record all requests, including those with successful 2xx status codes. <br/><br/>This field does **does not apply** to X-Ray traces, where all requests are recorded. |
 | addXRayTraceIdHeader | boolean | `false` | By default, the `X-Amzn-Trace-Id` header will not be added to the HTTP request. This means that the client-side trace and server-side trace will **not be linked** in X-Ray or the ServiceLens graph.<br/><br/> When this field is `true`, the `X-Amzn-Trace-Id` header will be added to HTTP requests (`XMLHttpRequest` or `fetch`). **Adding the header is dangerous and you must test your application before setting this field to `true` in a production environment.** The header could cause CORS to fail or invalidate the request's signature if the request is signed with sigv4.
 
+## Resource
```
qhanam, line 141: issue: I feel like the `Performance` configuration is the correct place to put this configuration option. This would (1) give it a similar pattern to the `Errors` configuration, and (2) give users control over all PerformanceEntry .

## Performance

| Name | Type | Default | Description |
| --- | --- | --- | --- |
| eventLimit | Number | `10` | The maximum number of resources to record ... |
| ignore | Function | `(entry: PerformanceEntry) => { ... }` |  A function which accepts a [PerformanceEntry](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry) and returns a boolean that determines if the PerformanceEntry should be ignored. <br/><br/> By default, [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) entries from browser extensions are ignored. |

qhanam (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,29 @@
...
+
+
+### Features
+
+* add ignore callback to resource plugin config ([3fd4d7b](https://github.com/aws-observability/aws-rum-web/commit/3fd4d7b184645e21e66f0ab199d3823c7a4435ee))
+* Create a session ID when cookies are disabled ([#314](https://github.com/aws-observability/aws-rum-web/issues/314)) ([6943587](https://github.com/aws-observability/aws-rum-web/commit/6943587259b8d7623742656199dc26b26b5bcd5d))
```
ps863, line 11: It looks like some features are being duplicated from 1.13.0 and 1.14.0
Please double check what is included in this release that was not included 1.13.0 

Eg - referrer change went out in 1.13.0

src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
```diff
@@ -250,4 +247,24 @@ describe('ResourcePlugin tests', () => {
             })
         );
     });
+
+    test('resource plugin ignores resources based on a custom condition', async () => {
```
ps863, line 251: nit: write test description in format ```when <action> then <expected behavior>```  (see above unit tests for reference).

docs/configuration.md
```diff
@@ -138,6 +138,26 @@ telemetries: [
...
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| ignore | Function(event: ResourceEvent) : boolean | `() => false` | A function which accepts a ResourceEvent and returns a boolean that determines if the [resource event](https://github.com/aws-observability/aws-rum-web/blob/main/src/event-schemas/resource-event.json) should be ignored. By default, no resources are ignored. |
```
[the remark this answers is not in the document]
williazz: Good catch. Originally, I had planned to do `ignore(PerformanceResourceTiming)`, and then changed my mind a forgot to fix the name

docs/configuration.md
williazz: That can work. `ignore()` can  be shared between `NavigationPlugin` and `ResourcePlugin` in perfromance config, since both use the `PerformanceResourceTiming` interface. (PerformanceEntry doesn't contain many useful fields for users to ignore).

However, does that mean `eventLimit` should also be shared and implemented in `NavigationPlugin` and `WebVitalsPlugin`? Currently, performance config is only supplied to `ResourcePlugin`

src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -117,20 +120,22 @@ export class ResourcePlugin extends InternalPlugin {
         }
     };
 
-    recordResourceEvent = (entryData: PerformanceResourceTiming): void => {
+    private isPutRumEventCall(eventData: PerformanceResourceTiming) {
```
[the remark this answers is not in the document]
williazz: I kept it for now, but in the future I will put them in a separate PR. Or at the very least, keep the commits separate

CHANGELOG.md
williazz: Removed the version bump

src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
williazz: Refactored

docs/configuration.md
williazz: Refactored!

src/plugins/event-plugins/ResourcePlugin.ts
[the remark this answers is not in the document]
qhanam: Why not fix it now?

docs/configuration.md
williazz: I'm not sure how to ignore PerformanceEntries that come from browser extensions at the moment. I have to do some digging

qhanam: > PerformanceEntry doesn't contain many useful fields for users to ignore

PerformanceEntry has multiple subclasses, and is actively being extended. Rather than provide a configuration option for each subclass, why not provide a single one that can be overloaded?

qhanam: > Currently, only ResourcePlugin receives the performance config

Yes the other plugins will need access to the config.

> However, does that mean eventLimit should also be shared and implemented in NavigationPlugin and WebVitalsPlugin?

No, all navigation events and web vitals events should be recorded.

williazz: > > PerformanceEntry doesn't contain many useful fields for users to ignore
> 
> PerformanceEntry has multiple subclasses, and is actively being extended. Rather than provide a configuration option for each subclass, why not provide a single one that can be overloaded?

My refactor stores ignore(performanceEntry) in performance config. And performanceEntry can be type cast as `PerformanceResourceTiming` for resource events or `PerformanceNavigationTiming` for navigation events.

src/plugins/event-plugins/ResourcePlugin.ts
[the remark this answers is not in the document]
williazz: Fixed, along with other unrelated refactors

docs/configuration.md
williazz: > I'm not sure how to ignore PerformanceEntries that come from browser extensions at the moment. I have to do some digging

This has been solved by just looking for a `chrome-extension*` prefix

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -1,20 +1,35 @@
...
+    ignore?: (event: PerformanceEntry) => any;
+};
+
+export type NavigationPluginConfig = {
+    ignore: (event: PerformanceEntry) => any;
+};
+
+export const defaultNavigationPluginConfig = {
+    ignore: defaultPerformanceIgnore
+};
```
qhanam, line 21: suggestion: Rename these types like `PerformancePluginConfig` and move them to `performance-utils` where they can be shared.

This should be non-breaking, because we don't export this type.

src/plugins/utils/performance-utils.ts
```diff
@@ -0,0 +1,2 @@
+export const defaultPerformanceIgnore = (entry: PerformanceEntry) =>
+    entry.name.startsWith('chrome-extension');
```
qhanam, line 2:
```suggestion
export const defaultPerformanceIgnore = (entry: PerformanceEntry) =>
    /^https?:/.test(entry.name);
```

src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
```diff
@@ -105,6 +108,20 @@ describe('NavigationPlugin tests', () => {
         // Assert
         expect(record).toHaveBeenCalledTimes(0);
     });
+    test('when ignore is customized then specific level 2 events are ignored', async () => {
```
qhanam, line 111:
```suggestion
    test('when entry is ignored then level 2 navigation is not recorded', async () => {
```

src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
```diff
@@ -250,4 +251,25 @@ describe('ResourcePlugin tests', () => {
             })
         );
     });
+
+    test('when ignore() is customized then specific events are not recorded', async () => {
```
qhanam, line 255:
```suggestion
    test('when entry is ignored then resource is not recorded', async () => {
```

src/plugins/utils/performance-utils.ts
```diff
@@ -0,0 +1,2 @@
+export const defaultPerformanceIgnore = (entry: PerformanceEntry) =>
```
qhanam, line 1: issue: We need tests for this function.

- *"when entry has a non-http URL schema then entry is ignored"*
- *"when entry has an http URL schema then entry is not ignored"*
- *"when entry is not type PerformanceResourceTiming then entry is not ignored"*

src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
```diff
@@ -250,4 +251,25 @@ describe('ResourcePlugin tests', () => {
...
+        // Setup
+        mockPerformanceObjectWithResources();
+        mockPerformanceObserver();
+        const plugin = buildResourcePlugin({
+            ignore: (entry: PerformanceEntry) =>
+                getResourceFileType(entry.name) === 'image'
```
qhanam, line 261:
```suggestion
            ignore: (entry: PerformanceEntry) => true
```

src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
```diff
@@ -105,6 +108,20 @@ describe('NavigationPlugin tests', () => {
...
+        // enables plugin by default
+        const plugin: NavigationPlugin = buildNavigationPlugin({
+            ignore: (event) =>
+                (event as PerformanceNavigationTiming).type === 'navigate'
```
qhanam, line 115:
```suggestion
            ignore: (event) => true
```

src/plugins/event-plugins/NavigationPlugin.ts
williazz: I moved all of the shared performance stuff to PerformanceConfig in performance-utils. I should not that `eventLimit` should be `resourceEventLimit`, but changing that at this point would be breaking I think.

src/plugins/utils/performance-utils.ts
williazz: Added unit tests for 
* 'when entry has a non-http URL schema then resource entry is ignored'
* 'when entry has an http URL schema then resource entry is not ignored'
* 'when entry has an https URL schema then resource entry is not ignored'

Did not add unit test for:
* 'when entry is not type PerformanceResourceTiming then entry is not ignored'. PerformanceResourceTiming extends PerformanceEntry, so any PerformanceResourceTiming will be recognized as a PerformanceEntry and will not be ignored.

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -184,11 +199,24 @@ export class NavigationPlugin extends InternalPlugin {
...
+            return;
+        }
+
         const eventDataNavigationTimingLevel2: NavigationEvent = {
             version: '1.0.0',
-            initiatorType: entryData.initiatorType,
-            navigationType: entryData.type,
+            initiatorType: entryData.initiatorType as
+                | 'navigation'
+                | 'route_change',
```
williazz, line 213: question: where does the `route_change` value come from? I don't see it on the MDN docs for `PerformanceResourceTiming.initatorType`

https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/initiatorType

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -184,11 +199,24 @@ export class NavigationPlugin extends InternalPlugin {
...
+            navigationType: entryData.type as
```
williazz, line 214: question: `PerformanceNavigationTiming.type` allows for types `navigation`, `reload`, `back_forward`, and `pre-render`. Where does `reserved` come from? 

https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/type

src/plugins/event-plugins/NavigationPlugin.ts
qhanam: I copied and pasted the type that TypeScript resolved. The NavigationEvent type is created from [the corresponding JSON schema](https://github.com/aws-observability/aws-rum-web/blob/main/src/event-schemas/navigation-event.json#L12-L20).

You are correct in that it would be more accurate to omit `route_change` here, since it will never appear in the PerformanceNavigationTiming object provided by the browser.

I assume the same is true of `entryData.type`.

williazz: Oh I see. Looks like something we can address in a separate change.

src/plugins/utils/performance-utils.ts
qhanam: > any PerformanceResourceTiming will be recognized as a PerformanceEntry and will not be ignored

We need to test the path when `entry.entryType !== 'resource'`, which is currently not covered. This differentiates PerformanceResourceTiming from other subclasses of PerformanceEntry.

At a higher level, I am proposing that we ignore **resources** which do not use the **http(s)** schema. By allowlisting http(s), we ensure that we omit browser extension resources from multiple browsers (e.g., `moz-extension`, `chrome-extension`, etc.), and other unknown resources which are not relevant to the application being monitored.

However, this allowlist must only apply to PerformanceResourceTiming entries. When this function is used elsewhere (e.g., PerformanceNavigationTiming), we must not apply the allowlist.

src/plugins/event-plugins/__tests__/PerformancePlugin.test.ts
```diff
@@ -0,0 +1,26 @@
+import { defaultPerformanceIgnore } from '../../utils/performance-utils';
```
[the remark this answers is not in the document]
williazz: Moved!

src/test-utils/mock-data.ts
```diff
@@ -300,6 +300,30 @@ export const MockPerformanceTiming: PerformanceTiming = {
...
+});
+
+export const mockPerformanceNavigationTiming = (
+    partialEntry?: any
+): PerformanceNavigationTiming => ({
+    ...mockPerformanceEntry(),
+    ...partialEntry,
+    entryType: 'navigation'
+});
+
```
qhanam, line 326: nit: We can omit these. The tests for performance-utils are already minimal.

qhanam (changes requested):

ps863:
High level comment: consider updating the description of PR accounting for these new changes in the revisions.

src/plugins/utils/performance-utils.ts
```diff
@@ -1,17 +1,30 @@
...
+    ignore?: (event: PerformanceEntry) => any;
+    recordAllTypes?: ResourceType[];
+    sampleTypes?: ResourceType[];
 };
 
 export type PerformancePluginConfig = {
-    ignore: (event: PerformanceEntry) => any;
     eventLimit: number;
+    ignore: (event: PerformanceEntry) => any;
+    recordAllTypes: ResourceType[];
```
williazz, line 16: issue: we need to update the performance config documentation

src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -72,6 +54,7 @@ export class ResourcePlugin extends InternalPlugin {
         const resourceObserver = new PerformanceObserver((list) => {
             list.getEntries()
                 .filter((e) => e.entryType === RESOURCE)
```
williazz, line 56: question: is this entryType check redundant?

src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -72,6 +54,7 @@ export class ResourcePlugin extends InternalPlugin {
...
+                .filter((e) => !this.config.ignore(e))
```
williazz, line 57: question: is the stream...filter pattern for readability?

src/test-utils/mock-data.ts
williazz: deleted

src/plugins/utils/__tests__/performance-utils.test.ts
```diff
@@ -0,0 +1,83 @@
...
+} from '../../../test-utils/mock-data';
+
+describe('performance-utils', () => {
+    describe('defaultPerformanceIgnore', () => {
+        test('when resource entry is PerformanceEntry and has non-http(s) URL schema then entry is ignored', () => {
+            const mockEntry = mockPerformanceEntry({
+                name: 'chrome-extension://localhost',
+                entryType: 'resource'
+            });
+            const result = defaultPerformanceIgnore(mockEntry);
```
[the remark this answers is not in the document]
williazz: refactored!

williazz:
> High level comment: consider updating the description of PR accounting for these new changes in the revisions.

updated

ps863 (approved):

docs/configuration.md
```diff
@@ -194,3 +194,21 @@ const awsRum: AwsRum = new AwsRum(
 | Name | Type | Default | Description |
 | --- | --- | --- | --- |
 | eventLimit | Number | `10` | The maximum number of resources to record load timing. <br/><br/>There may be many similar resources on a page (e.g., images) and recording all resources may add expense without adding value. The web client records all HTML files and JavaScript files, while recording a sample of stylesheets, images and fonts. Increasing the event limit increases the maximum number of sampled resources. |
+| ignore | Function(event: PerformanceEntry) : any | `(entry: PerformanceEntry) => !/^https?:/.test(entry.name)` | A function which accepts a [PerformanceEntry](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry) and returns a value that coerces to true when the PerformanceEntry should be ignored.</br></br> By default, [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) entries with URLs that do not have http(s) schemes are ignored. This causes resources loaded by browser extensions to be ignored. |
+
+For example, the following telemetry config array causes the web client to ignore all resource entries.
```
ps863, line 199: nit: As thought exercise, any other common ways customers might use ignore? If yes, we can add examples along that.

williazz: I had some other examples, such as blocking a [hostname](https://github.com/aws-observability/aws-rum-web/pull/419/commits/e44c80346ea12132bf12b3cce6d5eed4fef28a3d#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180R149-R156) or [type casting `entry` into other subclasses](https://github.com/aws-observability/aws-rum-web/pull/419/commits/e44c80346ea12132bf12b3cce6d5eed4fef28a3d#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180R149-R156), but I was asked to remove them

src/plugins/event-plugins/ResourcePlugin.ts
qhanam: From my perspective, yes, this is about improving readability.

src/plugins/utils/performance-utils.ts
qhanam: Documentation for `eventLimit`, `recordAllTypes` and `sampleTypes` has never existed, so there's no regression or omission with respect to this PR.

Let's close out this PR and track that separately.

docs/configuration.md
```diff
@@ -194,7 +194,7 @@ const awsRum: AwsRum = new AwsRum(
...
-| ignore | Function(event: PerformanceEntry) : any | `(entry: PerformanceEntry) => !/^https?:/.test(entry.name)` | A function which accepts a [PerformanceEntry](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry) and returns a value that coerces to true when the PerformanceEntry should be ignored.</br></br> By default, [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) entries with URLs that do not have http(s) schemes are ignored. This causes resources loaded by browser extensions to be ignored. |
+| ignore | Function(event: PerformanceEntry) : any | `(entry: PerformanceEntry) => entry.entryType === 'resource' && !/^https?:/.test(entry.name)` | A function which accepts a [PerformanceEntry](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry) and returns a value that coerces to true when the PerformanceEntry should be ignored.</br></br> By default, [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) entries with URLs that do not have http(s) schemes are ignored. This causes resources loaded by browser extensions to be ignored. |
```
williazz, line 197: good catch, thank you

qhanam (approved):

docs/configuration.md
ps863: maybe we can consider in future iterations of doc updates/improvements based on customer experience with this

williazz: yes absolutely

PR #421, opened 2023-06-29
chore(release): 1.14.0

williazz:
`npm run version` and manually remove duplicate changes

---


CHANGELOG.md
```diff
@@ -2,8 +2,16 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## [1.13.0](https://github.com/aws-observability/aws-rum-web/compare/v1.12.0...v1.13.0) (2023-02-23)
+## [1.14.0](https://github.com/aws-observability/aws-rum-web/compare/v1.12.0...v1.14.0) (2023-06-29)
```
ps863, line 5: diff link is not correct

CHANGELOG.md
```diff
@@ -2,8 +2,16 @@
...
+
+### Features
+
+* Ignore resources with non-http scheme ([#419](https://github.com/aws-observability/aws-rum-web/issues/419)) ([42a2ae5](https://github.com/aws-observability/aws-rum-web/commit/42a2ae568b954190db942377df32c7fc5b89686e))
```
ps863, line 9: issue: This is the name of the commit message of the feature. However, this description describes the default behavior of the new feature/option. I think the description here should be about the feature itself (eg - Allow users to filter unwanted performance events)

See [this](https://github.com/aws-observability/aws-rum-web/commit/0150cdc7bde8637ea51140331d4e5900c92108cf) example for reference.

williazz: updated

williazz:
Changed the target branch to aws-rum-web/main

CHANGELOG.md
```diff
@@ -2,8 +2,16 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## [1.13.0](https://github.com/aws-observability/aws-rum-web/compare/v1.12.0...v1.13.0) (2023-02-23)
+## [1.14.0](https://github.com/aws-observability/aws-rum-web/compare/v1.13.0...v1.14.0) (2023-06-29)
+
+### Features
+
+* Filter unwanted performance events with custom callback. By default, ignore resources with non-http url scheme. ([#419](https://github.com/aws-observability/aws-rum-web/issues/419)) ([42a2ae5](https://github.com/aws-observability/aws-rum-web/commit/42a2ae568b954190db942377df32c7fc5b89686e))
```
qhanam, line 9:
```suggestion
* Ignore resources with non-http scheme ([#419](https://github.com/aws-observability/aws-rum-web/issues/419)) ([42a2ae5](https://github.com/aws-observability/aws-rum-web/commit/42a2ae568b954190db942377df32c7fc5b89686e))
```

ps863:
We also need to hardcode the correct web client version in `eventCache` as we do not dynamically retrieve it yet. Refer to [this](https://github.com/aws-observability/aws-rum-web/commit/ac5b2b9adb66e0dade31bc8703ccb29990451a1c) for the changes that also need to accompany this.

CHANGELOG.md
williazz: Ideally, yes. I updated the change log text after receiving [this feedback](https://github.com/aws-observability/aws-rum-web/pull/421#discussion_r1247165267), which suggests that ignore()'s customizability should also be communicated. 

As a compromise, I will put the commit text first, and then add the additional details after.

williazz: updated

CHANGELOG.md
williazz: updated

CHANGELOG.md
qhanam: Oh I see; sorry I didn't notice that.

> this description describes the default behavior of the new feature

This is by design. In my opinion, the new (default) behavior is the critical information that consumers of this package need, and the fact that it is configurable is less important.

ps863: > In my opinion, the new (default) behavior is the critical information that consumers of this package need, and the fact that it is configurable is less important.

I think if the primary objective of the change was the default behavior, then I agree. If however, the primary objective of the change also involves giving consumers this configuration option as well, we should include it so consumers are aware of this new functionality available to them.

CHANGELOG.md
```diff
@@ -2,8 +2,16 @@
...
 
-## [1.13.0](https://github.com/aws-observability/aws-rum-web/compare/v1.12.0...v1.13.0) (2023-02-23)
+## [1.14.0](https://github.com/aws-observability/aws-rum-web/compare/v1.13.7...v1.14.0) (2023-06-29)
+
+### Features
+
+* Filter unwanted performance events with custom callback. By default, ignore resources with non-http url scheme. ([#419](https://github.com/aws-observability/aws-rum-web/issues/419)) ([42a2ae5](https://github.com/aws-observability/aws-rum-web/commit/42a2ae568b954190db942377df32c7fc5b89686e))
 
+### Bug Fixes
+* Export PageAttributes type. ([#369](https://github.com/aws-observability/aws-rum-web/issues/369)) ([4b78dc1](https://github.com/aws-observability/aws-rum-web/commit/4b78dc14799151422bee07f6263c1aaa098354f6))
```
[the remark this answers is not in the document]
williazz: > question: Is this the only bug fix that isn't already in v1.13.7?

#369 was the only bug fix that I found missing in v1.13.7, but that was a manual process. I took the changes from v1.13.0->v1.14.0, and removed the ones duplicated in v1.13.7 and v1.13.0. 

> should we include those fixes in the v1.14.0 change log? I.e., is this CHANGELOG comparing v1.14.0 to v1.13.0 or v1.13.7?

My intuition is to compare 1.14.0 and 1.13.7 as well. However, implementing that comparison is manual and prone to developer mistakes. Moreover, it's unintuitive to readers because patches are not written to the main branch's change log.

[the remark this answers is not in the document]
ps863: On quick check, historically it seems we only included immediate changes in minor version bumps and that is what was done here. 

> The question is, should we include those fixes in the v1.14.0 change log? 

Since we keep our patch change log on the release branch, maybe it is a good idea to include the important patches that have come between 1.3 and 1.4 into the main change log for consumers who may not refer to the change logs in the release branch. 

In the future, we could consider adding patch related change log into the main change log to maintain one place with all changes.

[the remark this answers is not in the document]
qhanam: > My intuition is that the diff should be between v1.14.0 and v1.13.7.

I think my intuition was wrong. The diff should be between v1.14.0 and v1.13.0 for the following reasons.

1. Each branch should have a complete CHANGELOG.
2. The versioning tool should automatically update CHANGELOG.

After a minor version is released, patches are merged to both the release branch and the main branch.

Regarding (1), patches made to a release branch are reflected in the CHANGELOG of the release branch, but not in the CHANGELOG of the main branch.

Regarding (2), I believe the versioning tool  looks for the most recent tag in the current branch, and will build a complete CHANGELOG from the start of the previous release branch.

qhanam:
> We also need to hardcode the correct web client version in `eventCache` as we do not dynamically retrieve it yet. Refer to [this](https://github.com/aws-observability/aws-rum-web/commit/ac5b2b9adb66e0dade31bc8703ccb29990451a1c) for the changes that also need to accompany this.

Good catch -- we should really fix that...

williazz:
As discussed offline with @qhanam, `standard-version` uses the tag with the latest timestamp to create the diff. So to create the correct diff between v1.14.0 and v1.13.0, I deleted and recreated tag v1.13.0 before running `standard-versino`. Just identifying this tech debt for awareness.

qhanam (approved):

ps863 (approved):

PR #430, opened 2023-08-02
feat: allow customization of default attributes

jhackshaw:
Allow the customization of default attributes, excluding the `platformType`. 

Primary use case is overriding the domain reported to RUM to facilitate using the same app monitor across different top-level domains which is not natively supported in RUM configuration.

---

### DIFF
--- a/src/event-cache/__tests__/EventCache.integ.test.ts
+++ b/src/event-cache/__tests__/EventCache.integ.test.ts
@@ -75,7 +75,7 @@ describe('EventCache tests', () => {
         });
     });
 
-    test('default meta data attributes except platformType can be overriden by custom attributes', async () => {
+    test('default meta data can be overriden by custom attributes', async () => {
         // Init
         const EVENT1_SCHEMA = 'com.amazon.rum.event1';
         const config = {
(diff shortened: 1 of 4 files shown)


src/event-cache/__tests__/EventCache.integ.test.ts
```diff
@@ -75,7 +75,7 @@ describe('EventCache tests', () => {
         });
     });
 
-    test('meta data contains default attributes not overridden from custom attributes', async () => {
+    test('default meta data attributes except platformType can be overriden by custom attributes', async () => {
```
qhanam, line 78: question: Why make an exception for platform type?

qhanam:
I am generally in favor of this change, but want to call out a few risks related to overwriting default attributes:
1. This may break the CloudWatch RUM console experience, which expects some values to be members of a set.
2. This may break metrics created by CloudWatch RUM, which expects some values to be members of a set.
3. Overwritten attributes will not be ingested if they fail CloudWatch RUM's validation steps.
4. Overwriting the domain increases the risk that an app monitor will receive data from an application which it is not meant to monitor. Specifically, if the initialization script is copied and pasted from one app monitor into another.

While it is currently possible to make these mistakes, this change makes it easier to make these mistakes. The tradeoff is that it provides more flexibility for CloudWatch RUM users, especially where feature gaps exist such as for app monitors that monitor multiple top-level domains.

src/event-cache/__tests__/EventCache.integ.test.ts
jhackshaw: It was the main one that I couldn't think of a valid use case to overwrite, but if it makes sense to allow it I can update this

qhanam: (2) seems like the closest thing to a two-way-door decision, as I feel like it wouldn't be too painful to revert it with a major version bump.

Do you have important use cases for overwriting other attributes, which would require (3)?

qhanam (approved):

qhanam:
> question: where do we enforce the 10 custom attribute limit? Does changing default attributes count towards that limit?
> 
> Eg - I change all the default attributes. Can I add 10 more custom attributes?

There is no change to the 10 custom attribute limit. Default attributes do not count towards the limit even if they are overwritten.

As far as I can tell, the web client does not enforce any limits. This is all done by the service.

ps863 (approved):

PR #432, opened 2023-08-15
feat: add attributions for core web vitals: LCP, CLS, and FID

williazz:
**Revision 1**
1. Collect attributions for LCP, FID, and CLS
2. Only some fields are captured, ignoring those that are redundant or not relevant for diagnostics

**Revision 2**
- [x] Remove attribution double check 
- [x] Integ test for browser support. Checks that attributions exist for LCP and CLS. FID cannot be tested because the web vitals package ignores programmatically triggered input. 

**Will be included in another PR**
1. Link CWV attributes to their related Resource and Navigation events
2. Integ test for FID, if possible

---

### DIFF
--- a/src/event-schemas/cumulative-layout-shift-event.json
+++ b/src/event-schemas/cumulative-layout-shift-event.json
@@ -11,7 +11,36 @@
         },
         "value": {
             "type": "number",
-            "description": "Value of the cls metric"
+            "description": "Largest burst of unexpected layout shifts during a page's lifespan"
+        },
+        "attribution": {
+            "type": "object",
+            "description": "Attributions for CLS",
+            "properties": {
+                "largestShiftTarget": {
+                    "type": "string",
+                    "description": "First element in the largest layout shift contributing to CLS score"
+                },
+                "largestShiftValue": {
+                    "type": "number",
+                    "description": "Value of CLS' single largest shift"
+                },
+                "largestShiftTime": {
+                    "type": "number",
+                    "description": "DOMHighResTimeStamp of CLS' single largest shift"
+                },
+                "loadState": {
+                    "type": "string",
+                    "enum": [
+                        "loading",
+                        "dom-interactive",
+                        "dom-content-loaded",
+                        "complete"
+                    ],
+                    "description": "LoadState during CLS' single largest shift"
+                }
+            },
+            "additionalProperties": false
         }
     },
     "additionalProperties": false,
--- a/src/plugins/event-plugins/WebVitalsPlugin.ts
+++ b/src/plugins/event-plugins/WebVitalsPlugin.ts
@@ -40,55 +40,46 @@ export class WebVitalsPlugin extends InternalPlugin {
     }
 
     handleLCP(metric: LCPMetricWithAttribution | Metric) {
-        const lcpEvent: LargestContentfulPaintEvent = {
+        const a = (metric as LCPMetricWithAttribution).attribution;
+        this.context?.record(LCP_EVENT_TYPE, {
             version: '1.0.0',
-            value: metric.value
-        };
-        if ('attribution' in metric) {
-            const a = (metric as LCPMetricWithAttribution).attribution;
-            lcpEvent.attribution = {
+            value: metric.value,
+            attribution: {
                 element: a.element,
                 url: a.url,
                 timeToFirstByte: a.timeToFirstByte,
                 resourceLoadDelay: a.resourceLoadDelay,
                 resourceLoadTime: a.resourceLoadTime,
                 elementRenderDelay: a.elementRenderDelay
-            };
-        }
-        this.context?.record(LCP_EVENT_TYPE, lcpEvent);
+            }
+        } as LargestContentfulPaintEvent);
     }
 
     handleCLS(metric: CLSMetricWithAttribution | Metric) {
-        const clsEvent: CumulativeLayoutShiftEvent = {
+        const a = (metric as CLSMetricWithAttribution).attribution;
+        this.context?.record(CLS_EVENT_TYPE, {
             version: '1.0.0',
-            value: metric.value
-        };
-        if ('attribution' in metric) {
-            const a = (metric as CLSMetricWithAttribution).attribution;
-            clsEvent.attribution = {
+            value: metric.value,
+            attribution: {
                 largestShiftTarget: a.largestShiftTarget,
                 largestShiftValue: a.largestShiftValue,
                 largestShiftTime: a.largestShiftTime,
                 loadState: a.loadState
-            };
-        }
-        this.context?.record(CLS_EVENT_TYPE, clsEvent);
+            }
+        } as CumulativeLayoutShiftEvent);
     }
 
     handleFID(metric: FIDMetricWithAttribution | Metric) {
-        const fidEvent: FirstInputDelayEvent = {
+        const a = (metric as FIDMetricWithAttribution).attribution;
+        this.context?.record(FID_EVENT_TYPE, {
             version: '1.0.0',
-            value: metric.value
-        };
-        if ('attribution' in metric) {
-            const a = (metric as FIDMetricWithAttribution).attribution;
-            fidEvent.attribution = {
+            value: metric.value,
+            attribution: {
                 eventTarget: a.eventTarget,
                 eventType: a.eventType,
                 eventTime: a.eventTime,
                 loadState: a.loadState
-            };
-        }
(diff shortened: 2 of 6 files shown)


src/event-schemas/cumulative-layout-shift-event.json
```diff
@@ -11,7 +11,36 @@
         },
         "value": {
             "type": "number",
-            "description": "Value of the cls metric"
+            "description": "Largest burst of unexpected layout shifts during a page's lifespan"
+        },
+        "attribution": {
```
williazz, line 16: Note: Individual attributes are not required by design. They are sourced from the web vitals package, which does not guarantee that every attribute will be included

src/event-schemas/cumulative-layout-shift-event.json
```diff
@@ -11,7 +11,36 @@
...
+            "properties": {
+                "largestShiftTarget": {
+                    "type": "string",
+                    "description": "First element in the largest layout shift contributing to CLS score"
+                },
+                "largestShiftValue": {
+                    "type": "number",
+                    "description": "Value of CLS' single largest shift"
+                },
+                "largestShiftTime": {
```
williazz, line 28: We may want to consider removing shiftTime because loadState makes this timestamp somewhat redundant

src/event-schemas/cumulative-layout-shift-event.json
```diff
@@ -11,7 +11,36 @@
...
+        },
+        "attribution": {
+            "type": "object",
+            "description": "Attributions for CLS",
+            "properties": {
+                "largestShiftTarget": {
+                    "type": "string",
+                    "description": "First element in the largest layout shift contributing to CLS score"
+                },
+                "largestShiftValue": {
```
williazz, line 24: Largest shift value may be more useful for business logic than diagnosing high CLS

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -25,20 +33,62 @@ export class WebVitalsPlugin extends InternalPlugin {
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     configure(config: any): void {}
 
-    getWebVitalData(webVitalData: Metric, eventType: string): void {
-        const webVitalEvent:
-            | LargestContentfulPaintEvent
-            | FirstInputDelayEvent
-            | CumulativeLayoutShiftEvent = {
+    protected onload(): void {
+        onLCP((metric) => this.handleLCP(metric));
```
williazz, line 37: Requires arrow function to preserve reference of keyword `this`

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -25,20 +33,62 @@ export class WebVitalsPlugin extends InternalPlugin {
...
+        onFID((metric) => this.handleFID(metric));
+        onCLS((metric) => this.handleCLS(metric));
+    }
+
+    handleLCP(metric: LCPMetricWithAttribution | Metric) {
```
williazz, line 42: We will never expect any of these handlers to receive just `Metric` because we are using the attribution build of `web-vitals`. However, they left `Metric` in as an argument type, so we should still double check that attributions are attached

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -1,12 +1,20 @@
...
 import {
-    LCP_EVENT_TYPE,
+    CLSMetricWithAttribution,
+    FIDMetricWithAttribution,
+    LCPMetricWithAttribution,
+    Metric,
+    onCLS,
+    onFID,
+    onLCP
+} from 'web-vitals/attribution';
```
qhanam, line 13: question: By how much does using the web-vitals attribution build (+ adding our own code to record attribution) increase the total size of the compressed aws-rum-web package?

At some point we will also want to create different builds, of different sizes, for aws-rum-web.

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -25,20 +33,62 @@ export class WebVitalsPlugin extends InternalPlugin {
...
+                element: a.element,
+                url: a.url,
+                timeToFirstByte: a.timeToFirstByte,
+                resourceLoadDelay: a.resourceLoadDelay,
+                resourceLoadTime: a.resourceLoadTime,
+                elementRenderDelay: a.elementRenderDelay
+            };
+        }
+        this.context?.record(LCP_EVENT_TYPE, lcpEvent);
     }
```
qhanam, line 59:
```suggestion
    handleLCP(metric: LCPMetricWithAttribution) {
        const a = metric.attribution;
        this.context?.record(LCP_EVENT_TYPE, {
            version: '1.0.0',
            value: metric.value,
            attribution: {
                element: a.element,
                url: a.url,
                timeToFirstByte: a.timeToFirstByte,
                resourceLoadDelay: a.resourceLoadDelay,
                resourceLoadTime: a.resourceLoadTime,
                elementRenderDelay: a.elementRenderDelay
            }
        });
    }
```

src/plugins/event-plugins/WebVitalsPlugin.ts
qhanam: question: Why do we need to double check that attributions are attached?

Looking at the type definition, it should always be `LCPMetricWithAttribution`, although TypeScript does think it could be `Metric` at this point. If we know it will always be `LCPMetricWithAttribution` then we can simplify this logic.

qhanam:
issue: I think we should create browser integ tests for web vital attribution, especially considering different browsers support different features related to the collection of web vitals.

src/plugins/event-plugins/WebVitalsPlugin.ts
williazz: That was my initial thought as well, but I was spooked into double checking because because TypeScript was complaining, like you said. I'm happy to make that change though

src/plugins/event-plugins/WebVitalsPlugin.ts
williazz: I did not do an exact comparison. There was no documentation on size difference in the web vitals package for different builds, so I just did my best to keep my changes as small as possible. 

I can check though.

williazz: I used the [bundle analyzer plugin](https://github.com/webpack-contrib/webpack-bundle-analyzer) to do a size comparison. I compared parsed sizes of the `cwr.js`, `WebVitalsPlugin`, and `web-vitals` package. 

**release/2.0.x**
* total size = 212.76kb
* web vitals plugin = 1.24kb
* web-vitals package = 6.52kb

**this pr**
* total size = 215.82kb
* web vitals plugin = 9.56kb
* web-vitals package = 2.02kb

**diff**
* total size = + ~3kb
* web vitals plugins = + ~8kb
* web vitals package = -4kb?

Not sure why the attributions build is smaller than the normal build.

src/plugins/event-plugins/WebVitalsPlugin.ts
williazz: Removed double check

williazz:
> issue: I think we should create browser integ tests for web vital attribution, especially considering different browsers support different features related to the collection of web vitals.

Currently looking into this. If the scope is big enough, I may cut a different PR @qhanam

src/plugins/event-plugins/__tests__/WebVitalsPlugin.test.ts
```diff
@@ -52,7 +64,7 @@ describe('WebVitalsPlugin tests', () => {
         record.mockClear();
     });
 
-    test('When web vitals are present then events are recorded', async () => {
+    test('When web vitals are present then LCP is recorded with attributions', async () => {
```
ps863, line 67: question: Do we need unit tests to cover case where attributions could not be collected?

williazz: Yes, we need "when [web vital] does not have attribution then event is recorded without attribution". Thanks

williazz: Correction: this test is not needed. Attributions object will always exist. Some properties are optional, such as `LCP.element` and all of the CLS fields. These fields will be defined as `{element: undefined...}` and dropped entirely during `JSON.stringify()`.

qhanam (approved):

PR #433, opened 2023-08-18
fix: Add getId to enhanced authflow

qhanam:
Cognito recently onboarded CloudWatch RUM to its enhanced authflow. However, the enhanced authflow implemented in the web client is currently broken because it uses the identity pool Id, instead of the identity Id, as the argument for `getCredentialsForIdentity`.

This change modifies the enhanced authflow to fetch the identity Id before calling `getCredentialsForIdentity`.

---

### DIFF
--- a/src/dispatch/__tests__/EnhancedAuthentication.test.ts
+++ b/src/dispatch/__tests__/EnhancedAuthentication.test.ts
@@ -1,14 +1,11 @@
 import { CRED_KEY } from '../../utils/constants';
-import { Credentials } from '@aws-sdk/types';
 import { EnhancedAuthentication } from '../EnhancedAuthentication';
-import { fromCognitoIdentityPool } from '../CognitoIdentityClient';
 import { DEFAULT_CONFIG } from '../../test-utils/test-utils';
 
 const mockGetId = jest.fn();
 const getCredentials = jest.fn();
 
 jest.mock('../CognitoIdentityClient', () => ({
-    fromCognitoIdentityPool: jest.fn(),
     CognitoIdentityClient: jest.fn().mockImplementation(() => ({
         getId: mockGetId,
         getCredentialsForIdentity: getCredentials
@@ -31,18 +28,6 @@ describe('EnhancedAuthentication tests', () => {
             sessionToken: 'z',
             expiration: new Date(Date.now() + 3600 * 1000)
         });
-        (fromCognitoIdentityPool as any).mockReset();
-        (fromCognitoIdentityPool as any).mockReturnValue(
-            () =>
-                new Promise<Credentials>((resolve) =>
-                    resolve({
-                        accessKeyId: 'x',
-                        secretAccessKey: 'y',
-                        sessionToken: 'z',
-                        expiration: new Date(Date.now() + 3600 * 1000)
-                    })
-                )
-        );
         localStorage.removeItem(CRED_KEY);
     });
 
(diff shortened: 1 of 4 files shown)


williazz (changes requested):
At first glance, the unit test `when getCredentialsForIdentity called then credentials are returned` is failing. I reran to approval workflow and it failed a second time. Will take a closer look

src/dispatch/__tests__/EnhancedAuthentication.test.ts
```diff
@@ -169,29 +154,19 @@ describe('EnhancedAuthentication tests', () => {
     test('when credential is retrieved from basic auth then next credential is retrieved from localStorage', async () => {
```
ps863, line 154: I think the title for this unit test and some others in this file still reference basic auth flow. 

1. The names of the tests need to be updated. ([Example](https://github.com/aws-observability/aws-rum-web/blob/ac471363f6a323fd1a57d898eb612f30a3dee763/src/dispatch/__tests__/EnhancedAuthentication.test.ts#L84)) 
2. We should remove guest role arn from the mocked web client config when using enhanced auth flow ([Example](https://github.com/aws-observability/aws-rum-web/blob/b3f58ceb130e90f99a2d0857bb661e4b9589a02a/src/dispatch/__tests__/EnhancedAuthentication.test.ts#L75))

ps863 (changes requested):

qhanam:
> question: Should we update the documentation [here](https://github.com/aws-observability/aws-rum-web/blob/main/docs/configuration.md) (specifically for the Guest Role ARN), to reflect that Enhanced auth flow will be used if its omitted?

Yes, and probably update the examples to show the enhanced authflow instead of the basic authflow.

I think we should do make these doc changes in a separate PR, as the purpose of this change is to patch a bug.

src/dispatch/__tests__/EnhancedAuthentication.test.ts
qhanam: Good catch, however I don't think fixing these fall within the scope of this change.

qhanam:
> At first glance, the unit test `when getCredentialsForIdentity called then credentials are returned` is failing. I reran to approval workflow and it failed a second time. Will take a closer look

Oops, I didn't wait for the CI checks to finish 😬 -- fixed

ps863 (approved):
Looks good to me.
We can do future PRs with: 
- New docs and doc updates 
- Update the test names

src/dispatch/__tests__/EnhancedAuthentication.test.ts
```diff
@@ -31,18 +28,6 @@ describe('EnhancedAuthentication tests', () => {
             sessionToken: 'z',
             expiration: new Date(Date.now() + 3600 * 1000)
```
williazz, line 29: nit: is GetCredentialsForIdentity supposed to return `secretKey` instead of `secretAccessKey`?

https://docs.aws.amazon.com/cognitoidentity/latest/APIReference/API_GetCredentialsForIdentity.html

williazz (approved):
Looks good. Double checked by reviewing the enhanced auth flow and looking at the documentation for GetCredentialsForIdentity. 

Q: What was the impact of this bug? And how long has GetCredentialsForIdentity required identityId instead of identityPoolId?

qhanam:
> Q: What was the impact of this bug? And how long has GetCredentialsForIdentity required identityId instead of identityPoolId?

Cognito's enhanced authflow doesn't support RUM, so it has never been integration tested and no one is using it.

That does raise the question of *"How do we integration test unauthenicated authorization?"*. I think the answer to this question is that we test during smoke tests. We should update the smoke tests accordingly before we "release" enhanced authflow by updating docs.

src/dispatch/__tests__/EnhancedAuthentication.test.ts
qhanam: The `CognitoIdentityClient`, which is implemented by the RUM web client (recall we don't use the AWS SDK here), returns `secretAccessKey`. This is by design, because it returns a `Credential` object, which has the property `secretAccessKey`.

One could make the argument that the `CognitoIdentityClient` should have behavior which is consistent with the actual Cognito client. However, I've altered this behavior for the sake of maintainability and reducing the size of the web client bundle.

PR #436, opened 2023-08-18
fix: Save deserialized credential object to Authentication member

adebayor123:
## Details
Currently in our Authentication and EnhancedAuthentication, we save credential object as serialized string values. Then, as we read the serialized string from Storage, we parse the string into a Credential object.

However, in AnonymousStorageCredentialsProvider, we overwrite the local member variable credential with the serialized string instead of the parsed object. As a result, any situation that requires reading of the credential from storage leads to: `TypeError: this.credentials.expiration.getTime is not a function.`

This PR removes the unnecessary step to overwrite the local member `this.credential` with the serialized string.

---

### DIFF
--- a/src/dispatch/EnhancedAuthentication.ts
+++ b/src/dispatch/EnhancedAuthentication.ts
@@ -77,12 +77,12 @@ export class EnhancedAuthentication {
     private AnonymousStorageCredentialsProvider =
         async (): Promise<Credentials> => {
             return new Promise<Credentials>((resolve, reject) => {
-                let credentials;
+                let credentials: Credentials;
                 try {
                     credentials = JSON.parse(localStorage.getItem(CRED_KEY)!);
                 } catch (e) {
                     // Error decoding or parsing the cookie -- abort
-                    reject();
+                    return reject();
                 }
                 // The expiration property of Credentials has a date type. Because the date was serialized as a string,
                 // we need to convert it back into a date, otherwise the AWS SDK signing middleware
@@ -95,7 +95,7 @@ export class EnhancedAuthentication {
                     // The credentials have expired.
                     return reject();
                 }
-                resolve(credentials);
+                resolve(this.credentials);
             });
         };
 
--- a/src/dispatch/EnhancedAuthentication.ts
+++ b/src/dispatch/EnhancedAuthentication.ts
@@ -47,12 +47,11 @@ export class EnhancedAuthentication {
      *
      * Implements CredentialsProvider = Provider<Credentials>
      */
-    public ChainAnonymousCredentialsProvider =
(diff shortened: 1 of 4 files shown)


ps863 (approved):

src/dispatch/EnhancedAuthentication.ts
```diff
@@ -87,8 +87,10 @@ export class EnhancedAuthentication {
                 // The expiration property of Credentials has a date type. Because the date was serialized as a string,
                 // we need to convert it back into a date, otherwise the AWS SDK signing middleware
                 // (@aws-sdk/middleware-signing) will throw an exception and no credentials will be returned.
-                credentials.expiration = new Date(credentials.expiration);
-                this.credentials = credentials;
+                this.credentials = {
+                    ...credentials,
+                    expiration: new Date(credentials.expiration as Date)
+                };
```
qhanam, line 93: issue: We need unit tests to cover the new behavior (here and in Authentication.ts).

I think this change introduces a bug. Cover the following cases:
1. When credentials are read from storage.
2. When credentials are read from storage, and then read from the member variable.

qhanam (changes requested):

qhanam:
I've pushed a patch for EnhancedAuthentication.ts:
1. Add type to `credentials` local variable
2. Add `return` statement when rejecting
3. Fix expiration (to be date object) in resolved value
4. Added unit tests

src/dispatch/__tests__/Authentication.test.ts
```diff
@@ -336,4 +336,69 @@ describe('Authentication tests', () => {
...
+        // Run
+        const credentials = await auth.ChainAnonymousCredentialsProvider();
+
+        // Assert
+        expect(credentials.expiration!.getTime()).toEqual(
+            storageExpiration.getTime()
+        );
+    });
+
+    test('when credentials are read from storage then the member variable stores the expiration as a date object', async () => {
```
qhanam, line 372: nit: This test will pass under certain conditions when it should fail (i.e., false negative).

Specifically, when `this.credentials.expiration` is `undefined`, and the `expiration` property of the value returned by `AnonymousStorageCredentialsProvider` is a `Date` object. When this happens, control flow will reach `AnonymousCognitoCredentialsProvider`, and the test needs a way to distinguish the value returned here from the expected value.

You can fix this by mocking the response of getCredentials, like was done in [EnhancedAuthentication.test.ts here](https://github.com/aws-observability/aws-rum-web/pull/436/files#diff-94a33512cc69b461d93e786b0cfc32e12f8767a11831604e38300860eb1ffba7R348-R355).

qhanam (approved):

PR #437, opened 2023-08-19
feat: Re-use Cognito identity id

qhanam:
Cognito identity Ids, which are used to allow unauthenticated users to send RUM data, can be re-used and do not expire.

This change stores identity Ids in localStorage in a similar way to how credentials are stored. Along with using the enhanced authflow, this means in most cases there will be one auth-related request instead of three.

---

### DIFF
--- a/src/dispatch/CognitoIdentityClient.ts
+++ b/src/dispatch/CognitoIdentityClient.ts
@@ -1,6 +1,7 @@
 import { HttpHandler, HttpRequest } from '@aws-sdk/protocol-http';
 import { Credentials } from '@aws-sdk/types';
 import { responseToJson } from './utils';
+import { IDENTITY_KEY } from '../utils/constants';
 
 const METHOD = 'POST';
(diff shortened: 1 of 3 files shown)


qhanam:
I'm skeptical about the "never expires" part. There might be some unintended consequences here.

Perhaps we can fetch a new Id if `getCredentialsForIdentity` fails because the Id is not recognized.

adebayor123 (approved):
question: Will we benefit from storing the identity key as part of the `CognitoIdentityClient`'s class member, similar to `this.credentials` in the `Authentication` class?

ps863 (approved):

qhanam:
> question: Will we benefit from storing the identity key as part of the `CognitoIdentityClient`'s class member, similar to `this.credentials` in the `Authentication` class?

Yes, but nearly as much. The credential gets accessed every time events get dispatched, while the identity gets accessed only when the credential expires.

PR #440, opened 2023-08-23
chore: store resource and navigation events by PerformanceEntry

williazz:
### Purpose
We need an event store to link related RUM events. My use case is to link LCP attribution to related RUM events.
1. Resource event for the image loaded, if any
2. Navigation event for the current page

### Revision 1
1. `EventStore` maps custom keys to RumEvents. Plugins write to store with `context.record(type, event, key)` and read from store with `context.getEvent(key)`. LCP attribution can use PerformanceEntry object references as the keys.
2. `ResourcePlugin` stores image resource events by reference to the PerformanceResourceTiming entry
3. `NavigationPlugin` caches all level 2 events by reference to PerformanceNavigationTiming entry. Level 1 events do not have a related PerformanceEntry, and are therefore ignored. 

### Revision 2
- [ ] Simplify `EventStore` eviction policy to "first n" or "last n"
- [ ] Reduce the events **stored** to the bare minimum, if possible
- [ ] Include level 1 navigation event
- [ ] Investigate reducing event **recording** to only those related to web vitals

### Not included in this PR
1. Linking LCP attribution to event IDs for 2 and 3 from above. LCP attribution collected by [web-vitals package](https://github.com/GoogleChrome/web-vitals/blob/main/src/types/lcp.ts#L68-L77) includes references navigationEntry and lcpResourceEntry. These references can be used as keys to retrieve the related event ID's via `context.getEvent(key)`. 

---


qhanam:
issue: It's not obvious from the PR description what problem this change solves, or how this change solves it. Reviewing this change is difficult without this context.

williazz:
> issue: It's not obvious from the PR description what problem this change solves, or how this change solves it. Reviewing this change is difficult without this context.

Updated the description. Tl; DR: this PR creates the API necessary for linking LCP attribution to related resource and navigation events.

src/event-cache/EventCache.ts
```diff
@@ -130,8 +141,12 @@ export class EventCache {
             // Return all events.
             rumEvents = this.events;
             this.events = [];
+            this.store.clear();
         } else {
             // Dispatch the front of the array and retain the back of the array.
+            for (let i = 0; i < this.config.batchLimit; i++) {
+                this.store.evictById(this.events[i].id);
+            }
```
qhanam, line 149: issue: I agree that we need an eviction policy to limit the memory consumed by the web client. A problem with evicting records that are being dispatched is that web vitals may not have completed by the time the event is evicted.

A few other options we could explore:
1. Is there a way for us to only retain the events related to web vitals attribution?
2. Should **only** record events related to web vitals attribution?
3. Would a simple eviction policy like **first N** or **most recent N** be suitable? 

Regarding (3), maybe we can reason about which would be suitable based on the LCP algorithm.

src/event-cache/EventStore.ts
```diff
@@ -0,0 +1,56 @@
...
+    put(key: string, event: RumEvent) {
+        const dup = this.get(key);
+        if (dup) {
+            this.keys.delete(dup.id);
+        }
+        this.keys.set(event.id, key);
+        this.events.set(key, event);
+    }
+
+    evict(key: string) {
```
ps863, line 32: question: Is there a potential for a condition between the time events are evicted vs when the event is already dispatched?

williazz: There is the edge case that when LCP attribution is reported, then resource and navigation events have already been batched and evicted. My understanding is that this is rare, since dispatch interval is 5 seconds and LCP is usually reported pretty quickly. 

However, @qhanam and I were discussing using a different eviction strategy to address this edge case and reduce the manageability burden. Storing "first N" events would likely capture all the LCP related events and reduce the overall size of this PR. "Most recent N" is similar, with a bit more overhead. 

Giving this some thought and will raise another revision.

src/event-cache/EventCache.ts
williazz: 1. Some optimization on event **storing** is already done, but I am looking into how to reduce this to the bare minimum
2. I did not consider reducing event **recording** to only those related to web vitals. That would be a major behavior change and require a separate PR. 
3. As we discussed offline, yes I think a simpler eviction strategy is better. Currently troubleshooting if we can get away with retaining "first N rum events".

williazz: Regarding (1), here are the optimizations
1. Only record the first navigation event, which describes the page that contains the LCP resource. We can ignore every subsequence navigation event. 
2. Record every image resource up until LCP is reported, then find the matching URL and ignore everything else. We must record everything because any of the images could be the largest contentful paint, and we have no way of knowing until LCP is reported.

williazz:
Closing for #445

PR #443, opened 2023-08-31
chore: upgrade hardcoded version to 2.0.0

williazz:
upgrade hardcoded version to 2.0.0

---


ps863:
question: Can we automate this process? 
I think we had some issues with this (https://github.com/aws-observability/aws-rum-web/pull/381), but wondering if there is a workaround this?

ps863 (approved):

williazz:
> is a workaround this?

To be honest, I did not look into it. But we have some time to fix it. Thanks for giving me context

qhanam:
We've been doing this at the same time we bump the version in package.json.

PR #445, opened 2023-09-05
chore: dispatch all events to EventBus

williazz:
> Rewrite of #440 

### Purpose

This PR creates the structure necessary for sharing information between plugins. 

### Uses cases: 

1. Have PluginA find eventId's created by PluginB etc. Useful for LCP attribution
2. WebVitalsPlugin tracks current count of in-flight Xhr and Fetch requests

### Revisions

**Revision 1**

I have created `EventBus` with topic-based filtering. I implemented our own instead of taking on a dependency because we only need a few extremely simple features. 
* Plugins can subscribe to any topic via `context.bus.subscribe(topic, callback)`
* Plugins can publish to any topic via `context.bus.dispatch(topic, message)`
* All events are published on every record by EventCache via `bus.dispatch(eventType, message)`

**Revision 2**
* Plugins are responsible for dispatching events
* EventCache is not responsible for dispatching events
* EventCache.record returns a ParsedRumEvent so plugins can access things like eventId, metadata, details, etc. 
* Read this thread for full context https://github.com/aws-observability/aws-rum-web/pull/445#discussion_r1317539603

**Revision 3**
* EventCache dispatches every event that is successfully recorded to topic 'events'
* Resource event supplies additional key=PerformanceResourceTiming
* Navigation supplies additional key=PerformanceNavigationTiming for level 2
* Read for full context https://github.com/aws-observability/aws-rum-web/pull/445#discussion_r1317539603

**Revision 4**
* Remove key
* Target main branch
### Examples 

**Use case 1:** PluginA finds eventId's of events from PluginB
```typescript
class WebVitalsPlugin {
   imageEvents = []
   constructor() {
      this.context.bus.subscribe(PERFORMANCE_RESOURCE_EVENT, this.storeResourceEvent)
   }
   storeResourceEvent(event) {
      if (event.fileType === ResourceTypes.IMAGE) {
         this.imageEvents.push(event)
      }
   }

   ...
   onLCP(payload => {
       ...
       findMatchingLCPEvent()
       // teardown
       this.imageEvents = []
       this.context.bus.unsubscribe(...)
   })
}
```

**Use case 2:** Count in-flight fetch/xhr requests
```typescript
// FetchPlugin.ts
...
private fetch() {
   ...
   this.context.bus.dispatch('http_open_count', {payload: 1})
   return original()
   .apply()
   .then(() => {
        this.context.bus.dispatch('http_open_count', {payload: -1})
    })
   .catch(() => {
        this.context.bus.dispatch('http_open_count', {payload: -1})
    })
}

...

// WebVitalsPlugin
class WebVitalsPlugins {
    ...
    http_open_count = 0
    handleHttpCountChange({payload}) { this.http_open_count += payload }
    constructor() {
        this.context.bus.subscribe('http_open_count', this.handleHttpCountChange)
   }
   teardown() {
       this.context.bus.unsubscribe('http_open_count', this.handleHttpCountChange)
   }
   ....
}
```

---

### DIFF
--- a/src/event-bus/EventBus.ts
+++ b/src/event-bus/EventBus.ts
@@ -1,8 +1,4 @@
-export type Subscriber = (message: Message) => void;
-export interface Message {
-    key?: any;
-    payload: any;
-}
+export type Subscriber = (message: any) => void;
 export enum Topic {
     EVENT = 'event'
 }
@@ -33,7 +29,7 @@ export default class EventBus<T = Topic> {
         return false;
     }
 
-    dispatch(topic: T, message: Message): void {
+    dispatch(topic: T, message: any): void {
         const list = this.subscribers.get(topic);
         if (list) {
             for (const subscriber of list) {
--- a/src/event-bus/__tests__/EventBus.test.ts
+++ b/src/event-bus/__tests__/EventBus.test.ts
@@ -1,4 +1,4 @@
-import EventBus, { Topic } from '../EventBus';
+import EventBus from '../EventBus';
 
 export enum MockTopics {
     FOOD = 'food',
@@ -18,11 +18,11 @@ describe('EventBus tests', () => {
         eventBus.subscribe(MockTopics.FOOD, l2);
 
         // run
-        eventBus.dispatch(MockTopics.FOOD, { payload: 'burger' });
+        eventBus.dispatch(MockTopics.FOOD, 'burger');
 
         // assert
-        expect(l1).toHaveBeenCalledWith({ payload: 'burger' });
-        expect(l2).toHaveBeenCalledWith({ payload: 'burger' });
+        expect(l1).toHaveBeenCalledWith('burger');
+        expect(l2).toHaveBeenCalledWith('burger');
     });
 
     test('when subscriber is removed then it is not called', async () => {
@@ -32,10 +32,10 @@ describe('EventBus tests', () => {
         const removed = eventBus.unsubscribe(MockTopics.FOOD, l2);
 
         // run
-        eventBus.dispatch(MockTopics.FOOD, { payload: 'burger' });
+        eventBus.dispatch(MockTopics.FOOD, 'burger');
 
         // assert
-        expect(l1).toHaveBeenCalledWith({ payload: 'burger' });
+        expect(l1).toHaveBeenCalledWith('burger');
         expect(removed).toBe(true);
         expect(l2).not.toHaveBeenCalled();
     });
@@ -45,7 +45,7 @@ describe('EventBus tests', () => {
         eventBus.subscribe(MockTopics.BOOKS, l2);
 
         // run
-        eventBus.dispatch(MockTopics.FOOD, { payload: 'burger' });
+        eventBus.dispatch(MockTopics.FOOD, 'burger');
 
         // assert
         expect(l2).not.toHaveBeenCalled();
--- a/src/event-cache/EventCache.ts
+++ b/src/event-cache/EventCache.ts
@@ -88,7 +88,7 @@ export class EventCache {
      *
      * @param type The event schema.
      */
-    public recordEvent = (type: string, eventData: object, key?: any) => {
+    public recordEvent = (type: string, eventData: object) => {
         if (!this.enabled) {
             return;
         }
@@ -98,7 +98,7 @@ export class EventCache {
             this.sessionManager.incrementSessionEventCount();
 
             if (this.canRecord(session)) {
-                this.addRecordToCache(type, eventData, key);
+                this.addRecordToCache(type, eventData);
             }
         }
     };
@@ -231,7 +231,6 @@ export class EventCache {
             type
         };
         this.eventBus.dispatch(Topic.EVENT, {
-            ...(key && { key }),
             payload: {
                 ...partialEvent,
                 details: eventData,
--- a/src/event-cache/EventCache.ts
+++ b/src/event-cache/EventCache.ts
@@ -203,7 +203,7 @@ export class EventCache {
      *
      * @param type The event schema.
      */
-    private addRecordToCache = (type: string, eventData: object, key?: any) => {
+    private addRecordToCache = (type: string, eventData: object) => {
         if (!this.enabled) {
             return;
         }
@@ -231,11 +231,9 @@ export class EventCache {
             type
         };
         this.eventBus.dispatch(Topic.EVENT, {
-            payload: {
-                ...partialEvent,
-                details: eventData,
-                metadata: metaData
-            }
+            ...partialEvent,
+            details: eventData,
+            metadata: metaData
         });
         this.events.push({
             ...partialEvent,
--- a/src/orchestration/Orchestration.ts
+++ b/src/orchestration/Orchestration.ts
@@ -207,7 +207,7 @@ export class Orchestration {
     private eventCache: EventCache;
     private dispatchManager: Dispatch;
     private config: Config;
-    private eventBus = new EventBus();
+    private eventBus = new EventBus<Topic>();
 
     /**
      * Instantiate the CloudWatch RUM web client and begin monitoring the
--- a/src/plugins/event-plugins/NavigationPlugin.ts
+++ b/src/plugins/event-plugins/NavigationPlugin.ts
@@ -265,8 +265,7 @@ export class NavigationPlugin extends InternalPlugin {
         if (this.context?.record) {
             this.context.record(
                 PERFORMANCE_NAVIGATION_EVENT_TYPE,
-                eventDataNavigationTimingLevel2,
-                entryData
+                eventDataNavigationTimingLevel2
             );
         }
     };
--- a/src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
+++ b/src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
@@ -44,10 +44,6 @@ describe('NavigationPlugin tests', () => {
                 navigationType: navigationEvent.type
             })
         );
-        // expect to record with key=PerformanceNavigationEntry
-        expect(record.mock.calls[0][2]).toEqual(
-            expect.objectContaining(navigationEvent)
-        );
     });
 
     test('When navigation timing level 2 API is not present then navigation timing level 1 API is recorded', async () => {
--- a/src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
+++ b/src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
@@ -77,9 +77,6 @@ describe('NavigationPlugin tests', () => {
                 navigationTimingLevel: 1
             })
         );
-
-        // // expect to record without key
-        expect(record.mock.calls[0].length).toEqual(2);
     });
 
     test('when enabled then events are recorded', async () => {
--- a/src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
+++ b/src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
@@ -65,9 +65,6 @@ describe('ResourcePlugin tests', () => {
                 initiatorType: resourceEvent.initiatorType
             })
         );
-        expect(record.mock.calls[0][2]).toEqual(
-            expect.objectContaining(resourceEvent)
-        );
     });
 
     test('when recordResourceUrl is false then the resource name is not recorded', async () => {
--- a/src/plugins/types.ts
+++ b/src/plugins/types.ts
@@ -2,7 +2,7 @@ import { Config } from '../orchestration/Orchestration';
 import { Session } from '../sessions/SessionManager';
 import EventBus, { Topic } from '../event-bus/EventBus';
 
-export type RecordEvent = (type: string, eventData: object, key?: any) => void;
+export type RecordEvent = (type: string, eventData: object) => void;
 export type RecordPageView = (pageId: string) => void;
 export type GetSession = () => Session | undefined;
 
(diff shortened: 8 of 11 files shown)


src/event-cache/EventCache.ts
```diff
@@ -220,12 +225,20 @@ export class EventCache {
...
 
-        this.events.push({
-            details: JSON.stringify(eventData),
+        const partialEvent = {
             id: v4(),
-            metadata: JSON.stringify(metaData),
             timestamp: new Date(),
             type
+        };
+        this.bus.notify(type, {
```
williazz, line 233: Note: The "Raw" RumEvent is published so that plugins do not have to de-stringify the information they want to look at.

src/orchestration/Orchestration.ts
```diff
@@ -444,7 +446,8 @@ export class Orchestration {
             config: this.config,
             record: this.eventCache.recordEvent,
             recordPageView: this.eventCache.recordPageView,
-            getSession: this.eventCache.getSession
+            getSession: this.eventCache.getSession,
+            bus: this.bus
```
williazz, line 450: Nit: Also open to flattening the bus 
```
context = {
...
subscribe: ...,
unsubscribe: ...,
notify: ...,
}
```

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
```diff
@@ -542,17 +545,9 @@ describe('FetchPlugin tests', () => {
...
-        const xRayOnContext: PluginContext = {
-            applicationId: 'b',
-            applicationVersion: '1.0',
-            config: { ...DEFAULT_CONFIG, ...{ enableXRay: true } },
-            record,
-            recordPageView,
-            getSession
-        };
-
+        const context = Object.assign({}, xRayOnContext, { getSession });
```
williazz, line 548: fix unrelated breaking changes

src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
```diff
@@ -70,18 +70,14 @@ describe('ResourcePlugin tests', () => {
...
         mockRandom(0); // Retain order in shuffle
-        const context: PluginContext = {
-            applicationId: 'b',
-            applicationVersion: '1.0',
-            config: { ...DEFAULT_CONFIG, recordResourceUrl: false },
-            record,
-            recordPageView,
-            getSession
-        };
+
```
williazz, line 73: fix unrelated breaking changes

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
```diff
@@ -583,14 +579,7 @@ describe('XhrPlugin tests', () => {
...
         const getSession: jest.MockedFunction<GetSession> = jest.fn();
-        const context: PluginContext = {
-            applicationId: 'b',
-            applicationVersion: '1.0',
-            config: { ...DEFAULT_CONFIG, ...{ enableXRay: true } },
-            record,
-            recordPageView,
-            getSession
-        };
+        const context = { ...mockContext, getSession };
```
williazz, line 582: fix unrelated breaking changes

src/plugins/types.ts
```diff
@@ -13,4 +13,5 @@ export type PluginContext = {
     record: RecordEvent;
     recordPageView: RecordPageView;
     getSession: GetSession;
+    bus: EventBus;
```
williazz, line 16: Nit: Also open to flattening the bus 
```
context = {
...
subscribe: ...,
unsubscribe: ...,
notify: ...,
}
```

src/event-cache/EventCache.ts
qhanam: question: We are potentially publishing more than needed. Consider the following options.
1. The EventCache publishes whenever any event is recorded.
2. The Resource plugin publishes whenever it records a PerformanceResource event.

williazz: I can refactor so that record() returns a raw rum event, which includes fields like `id` and `metadata`. That way, ResourcePlugin can publish all the information we need, for instance. 

Do you have any issue with the publishing parsed rum event?

qhanam: Does the web vitals plugin need a reference to the PerformanceResource object? If so, I'm not sure (1) will work.

williazz: (1) is a simpler implementation

> Does the web vitals plugin need a reference to the PerformanceResource object?

It does not need PerformanceResourceTiming object as a key because the URL is also a good key. But the performance object is the best key

src/plugins/event-plugins/DemoPlugin.ts
```diff
@@ -19,7 +19,7 @@ export class DemoPlugin implements Plugin {
         this.configuration = {};
         this.timerId = undefined;
         // eslint-disable-next-line @typescript-eslint/no-empty-function
-        this.recordEvent = () => {};
+        this.recordEvent = () => undefined;
```
williazz, line 22: note: linter was complaining about this one

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -292,4 +282,19 @@ export class NavigationPlugin extends InternalPlugin {
...
     }
+
+    private dispatchedLoad = false;
+    /** Record event and publish if record was successful */
+    private recordEvent(event: NavigationEvent) {
+        const parsedEvent = this.context?.record(
+            PERFORMANCE_NAVIGATION_EVENT_TYPE,
+            event
+        );
+        if (parsedEvent && !this.dispatchedLoad) {
```
williazz, line 293: Note: Only first event is needed

src/event-bus/__tests__/EventBus.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+describe('EventBus tests', () => {
+    let eventBus: EventBus;
+    const l1 = jest.fn();
+    const l2 = jest.fn();
+    beforeEach(() => {
+        eventBus = new EventBus();
+        jest.clearAllMocks();
+    });
+    test('when dispatch is invoked then all listeners are called', async () => {
```
ps863, line 11: nit: consider adding test to subscribe to different topics

src/event-bus/EventBus.ts
```diff
@@ -0,0 +1,41 @@
+export type Subscriber = (message: any) => void;
+export interface Message {
+    key?: any;
+    payload: any;
+}
+
+/** A topic-based event bus to facilitate communication between plugins */
+export default class EventBus {
```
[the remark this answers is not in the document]
williazz: Sorry, I renamed it to dispatch since notify was a bad name.

Notify is usually to notify of state changes that have already happened. Dispatch is usually for sending messages. So I thought dispatch was a better name

src/event-cache/EventCache.ts
```diff
@@ -82,8 +88,12 @@ export class EventCache {
      * If the session is not being recorded, the event will not be recorded.
      *
      * @param type The event schema.
+     * @returns {ParsedRumEvent | undefined} if the event was recorded
      */
-    public recordEvent = (type: string, eventData: object) => {
+    public recordEvent = (
+        type: string,
+        eventData: object
+    ): ParsedRumEvent | undefined => {
```
williazz, line 96: see https://github.com/aws-observability/aws-rum-web/pull/445#discussion_r1316495440

ps863:
high level question: could the addition and usage of an event bus impact the memory/performance footprint of the web client on web applications?

williazz:
> high level question: could the addition and usage of an event bus impact the memory/performance footprint of the web client on web applications?

We are currently only monitoring the size and relying design principles to limit memory and cpu usage. I have also been thinking about how to objectively monitor this performance. 

But from a design perspective, I think the impact is minimal.

src/event-cache/EventCache.ts
ps863: +1. I think (1) makes more sense for the points made by @williazz

qhanam: > +1. I think (1) makes more sense for the points made by @williazz

I agree.

> If we go back to (1), we can just add an optional key parameter to record() and have EventCache dispatch both key and payload.

Could we set the topic to 'event' and let the listener decide which events to do something with?

qhanam:
My thoughts on (1) vs (2):
**If doing (1)**, this topic allows any plugin to listen for any event. Publish all events under an ‘events’ topic.
**If doing (2)**, this topic is for the specific purpose of tracking resource/nav event Ids.  Return the event id from recordEvent and publish an <event id, url> tuple.

src/event-bus/__tests__/EventBus.test.ts
williazz: Added an extra test

src/event-cache/EventCache.ts
```diff
@@ -220,12 +225,23 @@ export class EventCache {
...
-        this.events.push({
-            details: JSON.stringify(eventData),
+        const partialEvent = {
             id: v4(),
-            metadata: JSON.stringify(metaData),
             timestamp: new Date(),
             type
+        };
+        this.eventBus.dispatch(Topic.EVENTS, {
+            ...(key && { key }),
```
williazz, line 234: Note: `...(key && { key })` checks that key exists before adding property, avoiding dispatching the unnecessary `{key: undefined}`

src/event-cache/EventCache.ts
williazz: > Could we set the topic to 'event' and let the listener decide which events to do something with?

Done in revision 3

qhanam (changes requested):

src/event-bus/EventBus.ts
```diff
@@ -0,0 +1,44 @@
+export type Subscriber = (message: Message) => void;
+export interface Message {
+    key?: any;
```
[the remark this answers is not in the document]
williazz: I have removed key

src/plugins/types.ts
williazz: Renamed to eventBus

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -265,7 +265,8 @@ export class NavigationPlugin extends InternalPlugin {
         if (this.context?.record) {
             this.context.record(
                 PERFORMANCE_NAVIGATION_EVENT_TYPE,
-                eventDataNavigationTimingLevel2
+                eventDataNavigationTimingLevel2,
+                entryData
```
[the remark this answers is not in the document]
williazz: I have removed these changes

src/event-bus/EventBus.ts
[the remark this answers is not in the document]
williazz: I agree that `key` does not serve a general purpose and have removed it. 

One drawback of (1) or (2) is that it requires plugins to execute a linear search on `performance.getEntries()` to find a matching hash or tuple. Or to maintain a map. With `key`, they get the matching entry directly.

src/event-cache/EventCache.ts
```diff
@@ -220,12 +225,22 @@ export class EventCache {
...
             timestamp: new Date(),
             type
+        };
+        this.eventBus.dispatch(Topic.EVENT, {
+            payload: {
+                ...partialEvent,
+                details: eventData,
+                metadata: metaData
+            }
+        });
```
qhanam, line 239:
```suggestion
        this.eventBus.dispatch(Topic.EVENT, {
            ...partialEvent,
            details: eventData,
            metadata: metaData
        });
```

src/plugins/event-plugins/__tests__/NavigationPlugin.test.ts
```diff
@@ -77,6 +77,9 @@ describe('NavigationPlugin tests', () => {
                 navigationTimingLevel: 1
             })
         );
+
+        // // expect to record without key
+        expect(record.mock.calls[0].length).toEqual(2);
```
qhanam, line 82:
```suggestion
```

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
qhanam: What was the problem here?

src/event-cache/EventCache.ts
```diff
@@ -198,7 +203,7 @@ export class EventCache {
      *
      * @param type The event schema.
      */
-    private addRecordToCache = (type: string, eventData: object) => {
+    private addRecordToCache = (type: string, eventData: object, key?: any) => {
```
qhanam, line 206:
```suggestion
    private addRecordToCache = (type: string, eventData: object) => {
```

qhanam (approved):

src/event-bus/EventBus.ts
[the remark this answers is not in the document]
williazz: Issue: 

(1) we cannot reliably create a `id` from <name, entryType, startTime, duration> because `recordResourceURL` is configurable. If turned off, we cannot match `PerformanceEntry.name` to `ResourceEvent.targetUrl` because the latter does not exist. If we do (1), it would have to use id=<entryType, startTime, duration>, which is still a decent key. 

(2) is the same thing as dispatching `{key: performanceEntry, payload:...}`, since the object hash is implied in javascript. Manually creating the hash would require [object-hash](https://github.com/puleos/object-hash) (~16M weekly downloads) but has size=~60kb. Doing our own implementation will be smaller but not very performant. 

In my opinion, if the goal is for key(performanceEntry) === key(resourceEvent), then **the most serviceable option is option (1) with key=<entryType, startTime, duration>**

PR #447, opened 2023-09-10
feat: Add trace id to http events

qhanam:
There is currently no direct link between HTTP requests and their respective X-Ray traces. This makes it difficult to find the correct X-Ray trace when debugging client-side issues related to HTTP requests.

This change adds the trace Id and segment Id to http request events when tracing is enabled.

---

### DIFF
--- a/src/event-schemas/http-event.json
+++ b/src/event-schemas/http-event.json
@@ -9,6 +9,16 @@
             "type": "string",
             "description": "Schema version."
         },
+        "trace_id": {
+            "type": "string",
+            "minLength": 35,
+            "maxLength": 35
+        },
+        "segment_id": {
+            "type": "string",
+            "minLength": 16,
+            "maxLength": 16
+        },
         "request": {
             "type": "object",
             "properties": {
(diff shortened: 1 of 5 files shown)


limhjgrace (approved):

williazz:
Question: at a high level, why are we using the subsegmentId as segmentId?

ps863 (approved):

ps863:
question: do smoke tests need an update? maybe in scope for different PR

qhanam:
> Question: at a high level, why are we using the subsegmentId as segmentId?

We could probably use either, but I believe the subsegment represents the actual HTTP request. In theory we could add other segments to the root segment.

williazz:
> We could probably use either, but I believe the subsegment represents the actual HTTP request. 

* A [segment](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-segments) provides the resource's name, details about the request, and details about the work done
* [Subsegments](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-subsegments) provide more granular timing information and details about downstream calls that your application made to fulfill the original request.

I'm still onboarding to x-ray concepts, but I am not sure that these two ideas are interchangeable. Not a blocking question though

PR #448, opened 2023-09-11
feat: link lcp attribution to image resource and navigation page load

williazz:
### Summary
LCP can be attributed to other RumEvents.
1. If LCP resource is an image, then a site owners want to know how that image was loaded
2. Site owners want to know if LCP is blocked by the page load sequence

### Revision 1
1. Link LCP Attribution to ResourceEvent eventId for image load, if any
2. Link LCP Attribution to NavigationEntry eventId of page load, if any

**Two callouts in revision 1**
1. Caught a bug where EventBus was instantiated twice
2. Populate resourceEvent with startTime

---


src/utils/common-utils.ts
```diff
@@ -7,6 +7,21 @@ export enum ResourceType {
...
+export interface HasLatency {
+    startTime: DOMHighResTimeStamp;
+    duration: DOMHighResTimeStamp;
+}
+
+/**
+ * Creates key to link a RumEvent to the PerformanceEntry that it is sourced from
+ * e.g. performanceKey(ResourceEvent) === performanceKey(PerformanceResourceTiming)
+ */
+export const performanceKey = (details: HasLatency) =>
```
williazz, line 22: Note: the general purpose of `performanceKey()` deserves some discussion.

All RumEvents that are sourced from PerformanceEntries share (or will share) fields duration and startTime. These fields are highly unique because they are [DOMHighResTimestamps](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp). 

This pattern can also used to link monkey patched HttpEvents to PerformanceResourceTiming entries for with `initatorType = Fetch || Xhr`

src/orchestration/Orchestration.ts
```diff
@@ -402,7 +402,8 @@ export class Orchestration {
                 id: applicationId,
                 version: applicationVersion
             },
-            this.config
+            this.config,
+            this.eventBus
```
williazz, line 406: Callout: this was a bug that was included in #445. Currently implementing robust tests to prevent this in the futrue

src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -120,6 +120,7 @@ export class ResourcePlugin extends InternalPlugin {
             const eventData: ResourceEvent = {
                 version: '1.0.0',
                 initiatorType: entryData.initiatorType,
+                startTime: entryData.startTime,
```
williazz, line 123: (2) Callout: `startTime` needs to be captured in order for `ResourceEvent` to be linked to its source `PerformanceEvent`

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -14,15 +14,28 @@ import {
...
 
 export const WEB_VITAL_EVENT_PLUGIN_ID = 'web-vitals';
 
 export class WebVitalsPlugin extends InternalPlugin {
     constructor() {
         super(WEB_VITAL_EVENT_PLUGIN_ID);
+        this.lcpHelper = this.lcpHelper.bind(this);
     }
+    private resourceEventIds = new Map<string, string>();
+    private navigationEventsIds: string[] = [];
```
qhanam, line 38: nit: Since we only ever use the first, this doesn't need to be an array.

williazz: agreed, changed to a private string var

ps863:
comment: I think the tests [here](https://github.com/aws-observability/aws-rum-web/blob/main/src/plugins/event-plugins/__tests__/WebVitalsPlugin.test.ts) need an update based on these changes

williazz:
> comment: I think the tests [here](https://github.com/aws-observability/aws-rum-web/blob/main/src/plugins/event-plugins/__tests__/WebVitalsPlugin.test.ts) need an update based on these changes

Unit tests have been added. Mocking these dependencies was tedious. We may want to introduce dependency injection or some other solution in the future

ps863 (approved):

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -34,28 +47,66 @@ export class WebVitalsPlugin extends InternalPlugin {
...
     protected onload(): void {
+        this.context.eventBus.subscribe(Topic.EVENT, this.lcpHelper); // eslint-disable-line
         onLCP((metric) => this.handleLCP(metric));
         onFID((metric) => this.handleFID(metric));
         onCLS((metric) => this.handleCLS(metric));
     }
 
-    handleLCP(metric: LCPMetricWithAttribution | Metric) {
+    private lcpHelper(msg: any) {
+        const event = msg as ParsedRumEvent;
```
qhanam, line 57:
```suggestion
    private lcpHelper(event: ParsedRumEvent) {
```

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -34,28 +47,66 @@ export class WebVitalsPlugin extends InternalPlugin {
...
+                    this.resourceEventIds.set(
+                        performanceKey(event.details as HasLatency),
+                        event.id
+                    );
+                }
+                break;
+            case PERFORMANCE_NAVIGATION_EVENT_TYPE:
+                if (!this.navigationEventId) {
+                    this.navigationEventId = event.id;
+                }
```
qhanam, line 75:
```suggestion
                this.navigationEventId = event.id;
```

williazz: I wanted to make this less brittle in case multiple navigation events are captured in the future, but I guess that's not a thing that will happen.

src/plugins/event-plugins/WebVitalsPlugin.ts
williazz: I remember typescript was complaining when I tried this initially. Double checking to see if this suggestion works

src/plugins/event-plugins/__tests__/WebVitalsPlugin.test.ts
```diff
@@ -47,11 +52,41 @@ const mockCLSData = {
...
 jest.mock('web-vitals/attribution', () => {
     return {
-        onLCP: jest
-            .fn()
-            .mockImplementation((callback) => callback(mockLCPData)),
+        onLCP: jest.fn().mockImplementation((callback) => {
+            context.eventBus.dispatch(Topic.EVENT, imageResourceRumEvent);
+            context.eventBus.dispatch(Topic.EVENT, navigationRumEvent);
+            callback(mockLCPDataWithImage);
+        }),
```
qhanam, line 89: question: Should we test any integrations? I'd like some evidence that everything works together as intended.

E.g.,
1. Between nav/resource and web vitals plugin.
2. Between browser and web vitals plugin.
3. Between web vitals lib and web vitals plugin.

src/event-bus/__tests__/EventBus.test.ts
```diff
@@ -1,9 +1,34 @@
...
+    constructor() {
+        super(MockPluginId);
+        this.subscriber = this.subscriber.bind(this);
+    }
+
+    enable(): void {} // eslint-disable-line
+    disable(): void {} // eslint-disable-line
+
+    protected onload(): void {
+        this.context.eventBus.subscribe(Topic.EVENT, this.subscriber); // eslint-disable-line
```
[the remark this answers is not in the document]
williazz: @typescript-eslint/unbound-method

src/plugins/event-plugins/__tests__/WebVitalsPlugin.test.ts
williazz: Good idea. FWIW manual testing had no problems so far

src/event-bus/__tests__/EventBus.test.ts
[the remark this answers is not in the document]
qhanam: oh nvm this is test code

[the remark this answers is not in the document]
williazz: i'll add the specific rule in my next revision where necessary

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -34,28 +47,63 @@ export class WebVitalsPlugin extends InternalPlugin {
     configure(config: any): void {}
 
     protected onload(): void {
+        this.context.eventBus.subscribe(Topic.EVENT, this.handleEvent); // eslint-disable-line
```
qhanam, line 50: nit: Instead of disabling es-lint here, you can fix this warning by (A) using an arrow function to declare `handleEvent`, or (B) annotating the function with `this: WebVitalsPlugin`.

williazz:
I couldn't get the LCP resource linking integration test to work. It works in manual testing every time though. 

My setup was 
```
// fs
app/
    ...
    web_vital_event.html
    assets/
        lcp.jpg
```

```html
// web_vital_event.html
body {
...
<img src="assets/lcp.jpg"/>
}
```

```typescript
// loader-web-vitals.js
loader('cwr', 'abc123', '1.0', 'us-west-2', './rum_javascript_telemetry.js', {
    dispatchInterval: 0,
    metaDataPluginsToLoad: [],
    eventPluginsToLoad: [
        new ResourcePlugin(),
        new NavigationPlugin(),
        new WebVitalsPlugin()
    ],
    telemetries: [],
    clientBuilder: showRequestClientBuilder
});
...
```

```typescript
// integ
// WebVitalPlug.test.ts
test('when lcp image resource is recorded then it is attributed to lcp', async (t: TestController) => {
    const browser = t.browser.name;
    if (browser === 'Safari' || browser === 'Firefox') {
        return 'Test is skipped';
    }
    await t.wait(300);

    await t
        // Interact with page to trigger lcp event
        .click(testButton)
        .click(makePageHidden)
        .expect(RESPONSE_STATUS.textContent)
        .eql(STATUS_202.toString())
        .expect(REQUEST_BODY.textContent)
        .contains('BatchId');

    const events = JSON.parse(await REQUEST_BODY.textContent).RumEvents;
    // console.log(events);
    const lcp = events.filter(
        (x: { type: string }) => x.type === LCP_EVENT_TYPE
    )[0];
    const resource = events.filter(
        (x: { details: string; type: string }) =>
            x.type === PERFORMANCE_RESOURCE_EVENT_TYPE &&
            x.details.includes('lcp.jpg')
    )[0];
    await t.expect(lcp.details).contains(`"lcpResourceEntry":"${resource.id}"`);
});
```

### Problem
TestCafe could never locate the resource event for the image, and therefore could never . However, during manual testing with the same setup, the

williazz:
There is something about TestCafe that hides the ResourceEvent representing the LCP resource. This prevents me from testing the event linking for LCPAttribution.lcpResourceEntry. 

My setup:

```
// fs
    app/
       ...
       web_vital_event.html
       assets/
           lcp.jpg
```

```typescript
// web_vital_event.html
... body {
    ...
    <img src="assets/lcp.jpg" />
}...
```

```typescript
// loader-web-vital.js
...
loader('cwr', 'abc123', '1.0', 'us-west-2', './rum_javascript_telemetry.js', {
    dispatchInterval: 0,
    metaDataPluginsToLoad: [],
    eventPluginsToLoad: [
        new ResourcePlugin(),
        new NavigationPlugin(),
        new WebVitalsPlugin()
    ],
    telemetries: [],
    clientBuilder: showRequestClientBuilder
});
...
```

```typescript
// WebVitalsPlugin.test.ts (integ)
test('when lcp image resource is recorded then it is attributed to lcp', async (t: TestController) => {
    const browser = t.browser.name;
    if (browser === 'Safari' || browser === 'Firefox') {
        return 'Test is skipped';
    }
    await t.wait(300);

    await t
        // Interact with page to trigger lcp event
        .click(testButton)
        .click(makePageHidden)
        .expect(RESPONSE_STATUS.textContent)
        .eql(STATUS_202.toString())
        .expect(REQUEST_BODY.textContent)
        .contains('BatchId');

    const events = JSON.parse(await REQUEST_BODY.textContent).RumEvents;
    // console.log(events);
    const lcp = events.filter(
        (x: { type: string }) => x.type === LCP_EVENT_TYPE
    )[0];
    const resource = events.filter(
        (x: { details: string; type: string }) =>
            x.type === PERFORMANCE_RESOURCE_EVENT_TYPE &&
            x.details.includes('fastly.picsum.photos')
    )[0];
    await t.expect(lcp.details).contains(`"lcpResourceEntry":"${resource.id}"`);
});
```

### Problem
The ResourceEvent simply does not exist when ran by TestCafe, but is successfully linked every time when I try manually with `npm run server` with the same setup.

williazz:
> There is something about TestCafe

This was fixed by upgrading test cafe XD

src/plugins/event-plugins/__integ__/WebVitalsPlugin.test.ts
```diff
@@ -61,3 +62,58 @@ test('WebVitalEvent records lcp and cls events on chrome', async (t: TestControl
         .expect(clsEventDetails.attribution)
         .typeOf('object');
 });
+
+test('when lcp image resource is recorded then it is attributed to lcp', async (t: TestController) => {
```
williazz, line 66: I believe this test is correct but there is something wrong with either TestCafe or ResourcePlugin that is causing this to fail.

williazz:
Need to investigate these failing integ tests on windows. They seem very unrelated
1. When custom attribute set at init, custom attribute recorded in event metadata
2. When a browser is not supported then the command function is a no-op
3. When cookie is disabled, sessionManager records events using member variables

williazz:
For commit "fix: switch order", the problem is that there are no resource events for resources that are already cached. So the resource event test had to be put first.

In the future,

1. TestCafe tests might be able to to disable caching https://testcafe.io/documentation/402736/reference/test-api/test/disablepagecaching
2. ResourcePlugin deserves a discussion on whether events should be made for cached resources, since it might be useful for site owners to know that assets are being cached

src/plugins/event-plugins/WebVitalsPlugin.ts
```diff
@@ -34,28 +47,63 @@ export class WebVitalsPlugin extends InternalPlugin {
...
 
-    handleLCP(metric: LCPMetricWithAttribution | Metric) {
+    private handleEvent = (event: ParsedRumEvent) => {
+        switch (event.type) {
+            // lcp resource is either image or text
+            case PERFORMANCE_RESOURCE_EVENT_TYPE:
+                const details = event.details as ResourceEvent;
+                if (
+                    details.fileType === ResourceType.IMAGE ||
+                    details.initiatorType === 'img'
```
qhanam, line 63: suggestion: Use the shared `getResourceFileType` function (from common-utils.ts) here to help ensure file type decisions are consistent.

williazz: fileType is a RUM concept not included in the PerformanceResourceTiming schema, so checking fileType is a check on `getResourceFileType`.

williazz: I removed the check to `initiatorType`

williazz:
This is an anti-pattern, but I have included #451 in this PR because it is a dependency. Please merge #451 before merging #448

src/utils/common-utils.ts
qhanam: Good discussion -- consider adding it to the method's JavaDoc comment for posterity.

> startTime or duration is zero... can happen when resources are cached

Does this happen in practice? Do we need some logic in place to prevent conflicts?

qhanam (approved):

src/utils/common-utils.ts
williazz: Updated the comment. It was my initial worry that duration might be zero when resources are cached. But when attempting to produce this, I have not seen duration actually be zero. Also, startTime will still be extremely unique.

williazz:
issue: WebVitals plugin needs to know if `onLCP` is supported before caching image resources to avoid memory leak

williazz:
> issue: WebVitals plugin needs to know if onLCP is supported before caching image resources to avoid memory leak

There are two solutions to this:
1. Check browser support at runtime. 
2. After 10 seconds, cleanup cache and stop listening for LCP. 

(1) is explicitly not recommended by by the owners of `web-vitals` https://github.com/GoogleChrome/web-vitals/issues/37#issuecomment-628916043. 

If we implemented (1) anyways, we would check support for PerformanceObserver + PerformancePaintTiming. Some additional analysis is needed to see what other dependencies the web-vitals requires. In the worst case, we never cleanup `resourceEventIds` and cause a small memory leak. If web-vitals package implements a polyfill for LCP in the future, this would have to be updated to avoid data loss.

(2) looks janky but is the simplest solution and my recommendation. The worst case is that we waste some memory for 10 seconds and then clean it up. The benefit is that we do not have to worry about any edge cases or do any maintenance work.

ps863:
> > issue: WebVitals plugin needs to know if onLCP is supported before caching image resources to avoid memory leak
> 
> There are two solutions to this:
> 
> 1. Check browser support at runtime.
> 2. After 10 seconds, cleanup cache and stop listening for LCP resource candidates.
> 
> (1) is explicitly not recommended by by the owners of `web-vitals` [GoogleChrome/web-vitals#37 (comment)](https://github.com/GoogleChrome/web-vitals/issues/37#issuecomment-628916043).
> 
> If we implemented (1) anyways, we would check support for PerformanceObserver + PerformancePaintTiming. Some additional analysis is needed to see what other dependencies the web-vitals requires. In the worst case, we never cleanup `resourceEventIds` and cause a small memory leak. If web-vitals package implements a polyfill for LCP in the future, this would have to be updated to avoid data loss.
> 
> (2) looks janky but is the simplest solution and my recommendation. The benefit is that we do not have to worry about many edge cases or do any maintenance work. The risk is that we are setting a hard limit--if it takes longer than ~10 seconds to report LCP, then we do not attribute lcpResourceEntry


I think approach 1 has other issues as well: 
- Browser support can change anytime (with new releases of browser versions, etc). This approach would require far more maintenance potentially if you are doing an explicit check. 


Question on approach 2: 
- Why 10 seconds? Is this based on a recommendation? What if the time window were smaller?

williazz:
Follow up on offline conversation...

> Why 10 seconds? Is this based on a recommendation? What if the time window were smaller?
10 sec is an arbitrary duration where LCP has probably been recorded already, if supported. We could do another time.

We ultimately opted for (1), and should keep an eye on updates to [LargestContentfulPaint](https://developer.mozilla.org/en-US/docs/Web/API/LargestContentfulPaint) browser support and `web-vitals`.

ps863 (approved):

PR #449, opened 2023-09-12
fix: Avoid overwriting existing trace header

limhjgrace:
When a CW Synthetics canary has `ActiveTracing` enabled and generates activity on a web application monitored by a CW RUM app monitor that also has `ActiveTracing` enabled, users experience a broken X-Ray trace experience where the segment from the canary to the web application is disconnected from the downstream segments. This is because the RUM Web Client overwrites the X-Ray trace header in the HTTP requests with a new trace while the RUM DataPlane drops activity generated by CW Synthetics canaries (and thus the trace is never created/sent to X-Ray for ingestion). 

This PR will update the Web Client tracing behavior to (1) not trace HTTP requests if the activity is generated by a CW Synthetics canary and (2) not overwrite existing trace headers.

Unit tests should be enough to cover this change, so no integrations were written.

---

### DIFF
--- a/src/plugins/event-plugins/FetchPlugin.ts
+++ b/src/plugins/event-plugins/FetchPlugin.ts
@@ -19,7 +19,8 @@ import {
     resourceToUrlString,
     is429,
     is4xx,
-    is5xx
+    is5xx,
+    getTraceHeader
 } from '../utils/http-utils';
 import { HTTP_EVENT_TYPE, XRAY_TRACE_EVENT_TYPE } from '../utils/constant';
 import {
@@ -271,8 +272,12 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
         if (!isUrlAllowed(resourceToUrlString(input), this.config)) {
             return original.apply(thisArg, argsArray as any);
         }
+        const traceHeader = getTraceHeader((input as Request).headers);
 
-        if (this.isTracingEnabled() && this.isSessionRecorded()) {
+        if (traceHeader.traceId && traceHeader.segmentId) {
+            httpEvent.trace_id = traceHeader.traceId;
+            httpEvent.segment_id = traceHeader.segmentId;
+        } else if (this.isTracingEnabled() && this.isSessionRecorded()) {
             trace = this.beginTrace(input, init, argsArray);
             httpEvent.trace_id = trace.trace_id;
             httpEvent.segment_id = trace.subsegments![0].id;
--- a/src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
+++ b/src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
@@ -32,6 +32,10 @@ const URL = 'https://aws.amazon.com';
 const TRACE_ID =
     'Root=1-0-000000000000000000000000;Parent=0000000000000000;Sampled=1';
 
+const existingTraceId = '1-0-000000000000000000000001';
+const existingSegmentId = '0000000000000001';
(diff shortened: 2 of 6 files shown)


williazz:
Question: why does Xhr not need to test for the following required for Fetch?

1. "when fetch is called and request has existing trace header then the plugin records a trace with existing trace data"
2. "when fetch is called and request has existing trace header then existing trace data is added to the http event"

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
```diff
@@ -934,4 +942,162 @@ describe('FetchPlugin tests', () => {
...
+                        },
+                        response: { status: 200, content_length: 125 }
+                    }
+                }
+            ]
+        });
+    });
+    test('when fetch is called and request has existing trace header then existing trace data is added to the http event', async () => {
+        // Init
+        console.log(navigator.userAgent);
```
williazz, line 1073:
```suggestion

```

limhjgrace:
> Question: why does Xhr not need to test for the following required for Fetch?
> 
> 1. "when fetch is called and request has existing trace header then the plugin records a trace with existing trace data"
> 2. "when fetch is called and request has existing trace header then existing trace data is added to the http event"

For XHR requests, we're unable to access/read the request headers and thus, can't detect if there is an existing trace header. This results in the changes for the XHR plugin to be a subset of the changes for the Fetch plugin and so, some of the tests added for the Fetch plugin don't apply for the XHR plugin changes.

src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -85,11 +87,26 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
...
+        let segmentId;
+
+        const headers = (input as Request).headers;
+        if (headers) {
+            const headerComponents = headers.get(X_AMZN_TRACE_ID)?.split(';');
+            if (headerComponents?.length === 3) {
+                traceId = headerComponents[0].split('Root=')[1];
+                segmentId = headerComponents[1].split('Parent=')[1];
+            }
+        }
```
qhanam, line 100: nit: Stick this in a helper function, and add unit tests. http-utils might be a good place.

src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -272,7 +289,11 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
             return original.apply(thisArg, argsArray as any);
         }
 
-        if (this.isTracingEnabled() && this.isSessionRecorded()) {
+        if (
+            !isSyntheticsUA() &&
```
qhanam, line 293: suggestion: Look for an existing trace header before we reach this point. If there is a header, then simply set `httpEvent.trace_id` and `httpEvent.segment_id`. If there is no header, then `beginTrace(...)`

I don't think we need to use `isSyntheticsUA` here, since we know whether or not there is a trace header.

src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -85,11 +87,26 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
...
+
         const startTime = epochTime();
         const http: Http = createXRayTraceEventHttp(input, init, true);
         const xRayTraceEvent: XRayTraceEvent = createXRayTraceEvent(
             this.config.logicalServiceName,
-            startTime
+            startTime,
+            undefined,
+            traceId,
+            segmentId
```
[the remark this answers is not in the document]
limhjgrace: I agree that we shouldn't create a new segment. Updated accordingly.

qhanam (approved):

PR #450, opened 2023-09-15
fix: Record resource timing after load event

qhanam:
The web client does not currently record PerformanceResource entries after the  *load* event fires. This means that slow, asynchronous resources, that contribute to poor web vitals like LCP or CLS, will not be recorded.

This change modifies the PerformanceResource plugin to continue listening to PerformanceResource entries after the load event.

To accomplish this, we modify our resource prioritization in the following ways:
1. We always record entries for document, scripts, stylesheets and fonts (was previously capped).
2. We record entries for everything else (including images) up to the configured limit.

---


src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -132,6 +114,12 @@ export class ResourcePlugin extends InternalPlugin {
     };
 
     protected onload(): void {
-        window.addEventListener(LOAD, this.resourceEventListener);
+        // We need to set `buffered: true`, so the observer also records past
```
williazz, line 117: Question: are will still concerned about using `performance.geEntries()` as backup for browsers that do not support `PerformanceObserver`?

williazz:
```when resource loads after window.load then the resource is recorded
      Browser: Firefox 117.0 / Ubuntu 20.04

         37 |        (e) =>
         38 |            e.type === PERFORMANCE_RESOURCE_EVENT_TYPE &&
         39 |            JSON.parse(e.details).targetUrl.includes('blank.png')
         40 |    );
         41 |
       > 42 |    await t.expect(events.length).eql(1);
         43 |});
         44 |
```
This was the same issue that I ran into in #448

src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -15,90 +15,72 @@ import {
...
-                .forEach((event) => {
-                    const type: ResourceType = getResourceFileType(event.name);
-                    if (this.config.recordAllTypes.includes(type)) {
-                        recordAll.push(event);
-                    } else if (this.config.sampleTypes.includes(type)) {
-                        sample.push(event);
-                    }
-                });
-        }
+        list.filter((e) => e.entryType === RESOURCE)
```
williazz, line 63: question: aren't these already filtered?

src/plugins/event-plugins/__integ__/ResourcePlugin.test.ts
```diff
@@ -18,25 +18,25 @@ test('when resource loads before the plugin then the resource is recorded', asyn
...
-                'rum_javascript_telemetry.js'
-            )
+            JSON.parse(e.details).targetUrl.includes('blank.png')
     );
 
     await t.expect(events.length).eql(1);
 });
 
-test('when resource loads after window.load then the resource is recorded', async (t: TestController) => {
+test('when resource loads before the plugin then the resource is recorded', async (t: TestController) => {
```
williazz, line 27: Question: When does window.load occur during TestCafe test? Why did reversing these fix the problem?

williazz:
I agree with the implementation. Only blocking thing only blocking thing for me to understand [how you fixed](https://github.com/aws-observability/aws-rum-web/pull/450#discussion_r1327049760) that resource plugin integration test

src/plugins/event-plugins/__integ__/ResourcePlugin.test.ts
qhanam: Firefox seems caching the image between runs, so during the second run it doesn't make a request to the server for it.

I assume there is some Firefox (or TestCafe) configuration that prevents this caching.

williazz (approved):

src/plugins/event-plugins/__integ__/ResourcePlugin.test.ts
williazz: Switched the tests in #448 like you did here

ps863 (approved):

src/plugins/utils/performance-utils.ts
```diff
@@ -20,11 +20,11 @@ export type PerformancePluginConfig = {
...
+        ResourceType.DOCUMENT,
+        ResourceType.SCRIPT,
         ResourceType.STYLESHEET,
-        ResourceType.IMAGE,
-        ResourceType.FONT,
-        ResourceType.OTHER
-    ]
+        ResourceType.FONT
+    ],
+    sampleTypes: [ResourceType.IMAGE, ResourceType.OTHER]
```
williazz, line 29: Question: Why are images not included? This may impact lcp attribution. I remember you considered recording all types

williazz: nevermind

src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -15,90 +15,72 @@ import {
...
+            .filter((e) => !this.config.ignore(e))
+            .forEach((event) => {
+                const type: ResourceType = getResourceFileType(event.name);
```
williazz, line 66: I have found that sometimes event.name does not include the correct file type extension. Could we add logic to use `initatorType` as a backup method to verify this information?

williazz: https://github.com/aws-observability/aws-rum-web/pull/448/files#diff-7f9eb043b568ee17c6e1abb8ce54754a6f6a1571604e94509f756b5a751a6e1cR62-R63

qhanam: I agree using `initiatorType` sounds like a good idea to get more accurate resource types. That is, instead of using only the file extension to identify the resource type,`getResourceFileType` would also use `initiatorType`.

I believe this is a new feature though. In #448, could you also use `getResourceFileType`? Then we can make a separate PR to improve `getResourceFileType` accuracy by considering `initatorType`.

PR #451, opened 2023-09-20
feat: getResourceType() uses initiatorType when file extension missing

williazz:
This is a follow up from an issue spotted in #450 and #448. See https://github.com/aws-observability/aws-rum-web/pull/450#discussion_r1330455383

The ResourcePlugin mislabels `fileType`. Currently, `fileType` maps to buckets of file extensions from `PerformanceResourceTiming.name` (e.g. ".jpg" and ".png" map to "image"). However, file extensions cannot be trusted to exist, such as during redirects. When extensions are missing, events are mislabelled as "other". 

I have considered two solutions:
1. Use [initiatorType](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/initiatorType) to infer fileType when file extension is missing. 
2. Remove the idea of "fileType" altogether and only use initiatorType. 

**My recommendation is (1), which this PR implements.** This is the least obtrusive method and we can always go to (2) later. If we were implementing this from scratch, I would instead recommend (2) because fileType is NOT a W3C schema found in the [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) schema. 

To help me review, please also read [initiatorType](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/initiatorType) documentation to see if I missed any mappings from `initiatorType` -> `fileType`. 

---

### DIFF
--- a/src/plugins/event-plugins/ResourcePlugin.ts
+++ b/src/plugins/event-plugins/ResourcePlugin.ts
@@ -1,9 +1,5 @@
 import { InternalPlugin } from '../InternalPlugin';
-import {
-    getResourceFileType,
-    ResourceType,
-    shuffle
-} from '../../utils/common-utils';
+import { getResourceFileType, shuffle } from '../../utils/common-utils';
 import { ResourceEvent } from '../../events/resource-event';
 import { PERFORMANCE_RESOURCE_EVENT_TYPE } from '../utils/constant';
 import {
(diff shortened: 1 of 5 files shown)


qhanam:
> My recommendation is (1), which this PR implements... If we were implementing this from scratch, I would instead recommend (2) because fileType is NOT a W3C schema found in the [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) schema.

I agree -- love this analysis. I think we should consider this for the next major version.

qhanam (approved):

PR #453, opened 2023-09-22
feat: Allow override of domain metadata attribute

qhanam:
In #430, we propose allowing all metadata attributes to be overridden. The primary motivation for this change is to work around the fact that CloudWatch RUM does not currently support associating multiple domains with an app monitor. Specifically, by allowing applications to overwrite the domain read from `window.location`, the application can spoof the domain and pass the domain check in CloudWatch RUM.

However, allowing **all** metadata attributes to be overridden **could** make it more difficult to provide backwards compatibility in the future, or cause users pain if they accidentally override a built-in metadata value.

For example, if the user agent is parsed by the service, instead of here, the service may choose overwrite the metadata values provided in the request. A counter argument is that allowing metadata attributes to be overridden makes it easier for applications to set values in cases where the built-in metadata values are incorrect.

Instead, this change allows **only** the domain attribute to be overridden. 

---


williazz:
Question: so just so I fully understand, Cx can do something like this to use multiple apps to the same app monitor. 

```typescript
cwr('addSessionAttributes', {
    domain: 'override', // spoof
    trueDomain: 'example.com'}) // manually insert actual domain for tracking
```

qhanam:
> just so I fully understand, Cx can do something like this to use multiple apps to the same app monitor?

Correct. Alternatively, a RUM user could differentiate user flows across domains by adding the domain to the page Id.

ps863 (approved):

williazz (approved):
I'm good with this change, though I think we should think of a no-configuration solution in the long run.

PR #455, opened 2023-09-22
chore: restore XhrPlugin var names

williazz:
Follow up to https://github.com/aws-observability/aws-rum-web/pull/454#discussion_r1335028410

1. Restores XhrPlugin names, as requested
2. Remove cacheSize() getter, which was only used for testing

---

### DIFF
--- a/src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
+++ b/src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
@@ -65,7 +65,6 @@ describe('XhrPlugin tests', () => {
                 statusText: 'OK'
             }
         });
-        expect((plugin as any).xhrMap.size).toEqual(0);
     });
 
     test('when XHR is called then the plugin records a trace', async () => {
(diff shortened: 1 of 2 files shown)


src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
```diff
@@ -65,7 +65,7 @@ describe('XhrPlugin tests', () => {
                 statusText: 'OK'
             }
         });
-        expect(plugin.cacheSize).toEqual(0);
+        expect((plugin as any).xhrMap.size).toEqual(0);
```
williazz, line 68: Note: type casting to `any` is how you access a private method/property in typescript

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
```diff
@@ -120,7 +120,7 @@ describe('XhrPlugin tests', () => {
                 }
             ]
         });
-        expect(plugin.cacheSize).toEqual(0);
+        expect((plugin as any).xhrMap.size).toEqual(0);
```
qhanam, line 123: issue: Each unit test should verify exactly one property of the program. In this case, we want to verify that `xhrMap` is empty after the XHR event was recorded. To accomplish this, we can a new unit test which verifies that property -- something like *"when there are no more active xhrs then the cache is empty"*.

Are there any other cases that we should test? Do we have reasonable path coverage (e.g., do we cover successful and error cases)?

williazz: > we can add a new unit test which verifies that property.

Created tests for Xhr success, error, abort, and timeout

> are there any other cases that we should test?

Maybe the DOMEventPlugin, I'd have to read through everything first.

PR #463, opened 2023-10-27
chore: Unit test urlsToInclude with tracing enabled

### LINKED ISSUE #429: [Feature Request]: Only add XRayTraceIdHeader to provided domains
### Please provide your feature request

We are trying to make use of the XRayTraceIdHeader but some of our requests to external servers are being blocked because of the header, which is to be expected.

Would be great if we could provide a list via the HttpPluginConfig with domains to which the XRayTraceIdHeader should be added, so it only adds the header when requests are being made to endpoints (our api's) which are known to allow the header.

We've worked around it for now by not adding 'http' to the telemetries and creating a custom HttpPlugin and XhrPlugin with this functionality, but for maintainability it would be great if the rum library would support such functionality.

qhanam:
When tracing is enabled, the web client currently adds the `X-Amzn-Trace-Id` header to **all** fetch and XHR requests. This causes these HTTP requests to fail (during CORS preflight) when the callee does not explicitly permit the request using the Access-Control-Allow-Origin response header. This prevents applications from enabling tracing, because the application owner may not want to, or be able to, modify the CORS policy of some endpoints.

This change solves this problem by adding the `urlsToTrace` configuration option. When an application uses this configuration option, the web client will only trace requests that match one of the regular expressions defined in `urlsToTrace`. By default, all requests are traced, making this change backwards compatible.

Resolves #429

---

### DIFF
--- a/src/plugins/event-plugins/FetchPlugin.ts
+++ b/src/plugins/event-plugins/FetchPlugin.ts
@@ -73,8 +73,12 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
         this.enable();
     }
 
-    private isTracingEnabled = () => {
-        return this.context.config.enableXRay;
+    private isTracingEnabled = (url: string) => {
+        return (
+            this.context.config.enableXRay &&
+            this.config.urlsToTrace.length &&
+            this.config.urlsToTrace.every((re) => re.test(url))
+        );
     };
 
     private isSessionRecorded = () => {
@@ -268,8 +272,9 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
     ): Promise<Response> => {
         const httpEvent: HttpEvent = this.createHttpEvent(input, init);
         let trace: XRayTraceEvent | undefined;
+        const url: string = resourceToUrlString(input);
 
-        if (!isUrlAllowed(resourceToUrlString(input), this.config)) {
+        if (!isUrlAllowed(url, this.config)) {
             return original.apply(thisArg, argsArray as any);
         }
         const traceHeader = getTraceHeader((input as Request).headers);
@@ -277,7 +282,7 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
         if (traceHeader.traceId && traceHeader.segmentId) {
             httpEvent.trace_id = traceHeader.traceId;
             httpEvent.segment_id = traceHeader.segmentId;
-        } else if (this.isTracingEnabled() && this.isSessionRecorded()) {
+        } else if (this.isTracingEnabled(url) && this.isSessionRecorded()) {
             trace = this.beginTrace(input, init, argsArray);
             httpEvent.trace_id = trace.trace_id;
             httpEvent.segment_id = trace.subsegments![0].id;
--- a/src/plugins/event-plugins/FetchPlugin.ts
+++ b/src/plugins/event-plugins/FetchPlugin.ts
@@ -73,12 +73,8 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
         this.enable();
     }
 
-    private isTracingEnabled = (url: string) => {
-        return (
-            this.context.config.enableXRay &&
-            this.config.urlsToTrace.length &&
-            this.config.urlsToTrace.every((re) => re.test(url))
-        );
+    private isTracingEnabled = () => {
+        return this.context.config.enableXRay;
     };
 
     private isSessionRecorded = () => {
@@ -272,9 +268,8 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
     ): Promise<Response> => {
         const httpEvent: HttpEvent = this.createHttpEvent(input, init);
         let trace: XRayTraceEvent | undefined;
-        const url: string = resourceToUrlString(input);
 
-        if (!isUrlAllowed(url, this.config)) {
+        if (!isUrlAllowed(resourceToUrlString(input), this.config)) {
             return original.apply(thisArg, argsArray as any);
         }
         const traceHeader = getTraceHeader((input as Request).headers);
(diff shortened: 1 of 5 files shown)


src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -73,8 +73,12 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
         this.enable();
     }
 
-    private isTracingEnabled = () => {
-        return this.context.config.enableXRay;
+    private isTracingEnabled = (url: string) => {
+        return (
+            this.context.config.enableXRay &&
+            this.config.urlsToTrace.length &&
+            this.config.urlsToTrace.every((re) => re.test(url))
```
williazz, line 80: nit: I think `.every()`` evaluates to true when the list is empty

williazz, line 80:
```suggestion
            this.context.config.enableXRay &&
            this.config.urlsToTrace.every((re) => re.test(url))
```

nit: I think `.every()` evaluates to true when the list is empty

williazz: Sorry ignore this, I didn't see your intended behavior for empty list

williazz:
```suggestion
            this.context.config.enableXRay &&
            this.config.urlsToTrace.length &&
            this.config.urlsToTrace.some((re) => re.test(url))
```
> the web client will only trace requests that match one of the regular expressions defined in urlsToTrace.

nit: did you mean to use some instead of every?

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
```diff
@@ -969,4 +966,58 @@ describe('FetchPlugin tests', () => {
...
+
+        // Run
+        await fetch(URL);
+        plugin.disable();
+
+        // Assert
+        expect(record).toHaveBeenCalledTimes(0);
+    });
+
+    test('when the url does not match urlsToTrace then the plugin does not record a trace', async () => {
```
williazz, line 987: question: what about the test case 'when url matches some but not all regex in urlsToTrace'?

src/plugins/event-plugins/XhrPlugin.ts
```diff
@@ -130,8 +130,12 @@ export class XhrPlugin extends MonkeyPatched<XMLHttpRequest, 'send' | 'open'> {
         return this.config.addXRayTraceIdHeader;
     };
 
-    private isTracingEnabled = () => {
-        return this.context.config.enableXRay;
+    private isTracingEnabled = (url: string) => {
+        return (
+            this.context.config.enableXRay &&
+            this.config.urlsToTrace.length &&
+            this.config.urlsToTrace.every((re) => re.test(url))
```
williazz, line 137: question: same as the other

src/plugins/event-plugins/__tests__/XhrPlugin.test.ts
```diff
@@ -1122,4 +1121,91 @@ describe('XhrPlugin tests', () => {
             }
         });
     });
+
+    test('when urlsToTrace is empty then the plugin does not record a trace', async () => {
```
williazz, line 1125: question: same as the other

williazz:
> We used 'urlsToTraceWithXRayHeader'. Using 'urlsToTrace' would also be fine but sounds a little misleading as tracing would still work even if a url is not included in the 'urlsToTrace' config option, it would just not add the xray header id to the request.

This PR disables traces that are not allow listed, but the customer only requests not adding the xray header. Am I understanding correctly?

qhanam:
> This PR disables traces that are not allow listed, but the customer is only asking to add the xray header depending on the allow list.

Yes, we need to decide which of the following behaviors we want:
1. The configuration option controls whether or not the request is traced.
2. The configuration option controls whether or not the X-Amzn-Trace-Id is added to the request.

The current implementation is (1).

Alternatively, we could use `urlsToInclude` for tracing in the same way we use it for recording http requests. Two options:  
3. We trace a request if it matches `urlsToInclude` and does not match `urlsToExclude`.
4. We trace a requests if and only if the http request is recorded.

src/plugins/utils/http-utils.ts
```diff
@@ -26,6 +27,7 @@ export type HttpPluginConfig = {
     logicalServiceName: string;
     urlsToInclude: RegExp[];
     urlsToExclude: RegExp[];
+    urlsToTrace: RegExp[];
```
qhanam, line 30: suggestion: Instead of creating a new config option, extend `addXRayTraceIdHeader` to be either a boolean (current) or a regex array (proposed extension). When it is a regex array, it would only add headers to matching URLs.

src/plugins/event-plugins/FetchPlugin.ts
ps863: thought: we should make this clear in documentation. some customers might think if a url fails any one of the regex tests, it would not be included, which is not the case here.

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
```diff
@@ -969,4 +966,58 @@ describe('FetchPlugin tests', () => {
...
+        const plugin: FetchPlugin = new FetchPlugin(config);
+        plugin.load(xRayOnContext);
+
+        // Run
+        await fetch(URL);
+        plugin.disable();
+
+        // Assert
+        expect(record).toHaveBeenCalledTimes(1);
+        expect(record.mock.calls[0][0]).toEqual(XRAY_TRACE_EVENT_TYPE);
```
ps863, line 1021: question: what about test case for if the regex expression is invalid?

src/plugins/utils/http-utils.ts
ps863: +1 agree, lot cleaner.

qhanam:
I think we've reached consensus that we want the following:
1. The `addXRayTraceIdHeader` option will be extended to accept an allowlist of URLs.
2. The `urlsToInclude` and `urlsToExclude` options already prevents tracing.

Regarding (1), we will create a separate PR.

Regarding (2), I've reverted the changes which added the `urlsToTrace` config option, and amended the unit tests to verify `urlsToInclude` and `urlsToExclude` also cover tracing.

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
qhanam: In general the web client doesn't validate config options, which means the behavior of the web client is undefined when a config option is invalid.

Perhaps we should consider doing validation for options on critical paths. The downside is that it makes the package larger and more complex.

williazz: I think an invalid reg exp would throw an error here at the very least https://github.com/aws-observability/aws-rum-web/blob/main/src/plugins/utils/http-utils.ts#L78

williazz (approved):

williazz:
We should probably remove "Resolves #429" but otherwise looks good!

PR #465, opened 2023-10-30
Time to interactive 

ps863:
TTI (Time to Interactive) is a measurement that can tell users how their web application took for them to ready for interaction. 

TTI can be measurement using several different algorithms (mainly [Google](https://developer.chrome.com/en/docs/lighthouse/performance/interactive/) and [Boomerang](https://akamai.github.io/boomerang/oss/BOOMR.plugins.Continuity.html#toc11__anchor)). 

Since TTI is gradually being phased out, supported for the library implementations of either is limited/on path to deprecation. Therefore, in this PR, we implement the Boomerang TTI algorithm ourselves for better maintainability and avoid taking dependency on packages that are no longer actively maintained. 

#### How can users get TTI telemetry for their web applications? 

Documentation changes will come in separate PR. 
Since TTI is used by limited users and incurs additional telemetry expenses (for additional events), users will have to import the TTI plugin and then configure their web client configuration to use it. 

```
import { TTIPlugin } from "aws-rum-web";

```
```
const config = {
    sessionSampleRate: 1,
    guestRoleArn:
      "",
    identityPoolId: "",
    endpoint: "https://dataplane.rum.us-west-2.amazonaws.com",
    telemetries: ["performance", "errors", "http"],
    eventPluginsToLoad: [new TTIPlugin()],
    allowCookies: true,
    enableXRay: false,
  };
```

#### How is TTI using algorithm measured? 

We implement TTI Boomerang algorithm with some modifications

Steps to TTI: 

1) Find visually ready timestamp (highest of domcontentLoadedEnd(if available), FCP(if available), LCP(if available))
2) Starting from the Visually Ready timestamp, find a 500ms quiet window. 

A quiet window has the following characteristics: 
a) No Long Tasks 
b) FPS is above 20

Boomerang TTI's measurement interval is every 100ms. Each check is performed and bucketed into a 100ms interval. 

3) TTI is recorded as Visually ready timestamp + time from visually ready to the start of the quiet window

Note: 
We do not record PageBusy metrics for now as there were some performance issues observed with applications getting blocked. 
Without PageBusy metric, the algorithm requires that Long Tasks are recorded to compute a TTI. As such TTI can only be computed for now for browsers that support Long Tasks. 

---


src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+        return new Promise<number>((resolve, reject) => {
+            this.checkForVisualReady().then((result) => {
+                this.computeTTI(result).then((result) => {
+                    resolve(result);
+                });
+            });
+        });
+    }
+
+    checkForVisualReady(): Promise<number> {
```
qhanam, line 64: nit: By convention, this package uses async functions instead of promises. Make this an async function.

Async functions are just promises with syntactic sugar.

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+                }
+                if (this.lcpSupported) {
+                    if (!this.lcpTime) {
+                        visuallyReady = false;
+                    }
+                }
+
+                if (!this.domContentLoadedEventEnd) {
+                    visuallyReady = false;
+                }
```
qhanam, line 85: question: Do we need these checks? Instead, could we check that they are supported before computing TTI, as we already do for long task support? 
```suggestion
                let visualReady = this.fcpTime && this.lcpTime && this.domContentLoadedEventEnd;
```

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+
+                if (
+                    visuallyReady ||
+                    timeIntervals * this.CHECK_PERIOD >
+                        this.VISUALLY_READY_RESOLVE_TIMEOUT
+                ) {
```
qhanam, line 91: issue: There's a lot of nesting here, which makes it difficult to follow the logic. Could you flatten this somehow? E.g., by extracting methods where needed.

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+                    // If timed out, check if any visually ready signals are present, else can't find tti
+                    if (
+                        timeIntervals * this.CHECK_PERIOD >
+                        this.VISUALLY_READY_RESOLVE_TIMEOUT
+                    ) {
+                        if (
+                            !this.lcpTime &&
+                            !this.domContentLoadedEventEnd &&
+                            !this.fcpTime
```
qhanam, line 100: question: Isn't this always false, since `visualReady` must be true here? Or am I not following the logic correctly?

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+            );
+        }
+        if (metric.name === LCP) {
+            metric.entries.forEach(
+                (entry) => (this.lcpTime = entry.startTime + entry.duration)
+            );
+        }
+    }
+
+    computeTTI(visuallyReadyTimestamp: number): Promise<number> {
```
qhanam, line 140: nit: Make this an async function.

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+                    totalIntervals * this.CHECK_PERIOD >
+                    this.TTI_RESOLVE_TIMEOUT
+                ) {
+                    // TTI did not resolve in timeout period, so TTI can't be computed
+                    this.ttiResolved = true;
+                    this.ttiTracker = {};
+                    reject('TTI computation timed out');
+                }
+                const endBucket = this.computeTimeWindow();
+                for (let bucket = currBucket; bucket <= endBucket; bucket++) {
```
qhanam, line 161: suggestion: This has a lot of nesting, and cyclomatic complexity. Could you clean it up? Helper functions would make this easier to read.

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+        );
+        eventObserver.observe({
+            entryTypes: [LONG_TASK, NAVIGATION]
+        });
+    };
+
+    /*
+    Track frames per second 
+    */
+    framesPerSecondListener = () => {
```
qhanam, line 319: question: What does this do, and is it necessary? It seems like it might impact application performance.

src/plugins/event-plugins/TTIPlugin.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    disable(): void {}
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    configure(config: any): void {}
+
+    load(context: PluginContext): void {
+        this.context = context;
+        const ttiBoomerang: TTIBoomerang = new TTIBoomerang();
```
qhanam, line 28: suggestion: *"Boomerang"* is another project and doesn't help describe the behavior. Instead, name this class *"TimeToInteractive"*.

src/plugins/event-plugins/__tests__/TTIPlugin.test.ts
```diff
@@ -0,0 +1,124 @@
+import { TTIBoomerang } from '../../../time-to-interactive/TTIBoomerang';
+import { mockLongTaskPerformanceObserver } from '../../../test-utils/mock-data';
+import { TTIPlugin } from '../TTIPlugin';
+import { context, record } from '../../../test-utils/test-utils';
+import { TIME_TO_INTERACTIVE_EVENT_TYPE } from '../../../plugins/utils/constant';
+
+describe('Time to Interactive - Plugin Tests', () => {
```
qhanam, line 7: issue: I don't see a test that verifies the TTI value is correct. We could add one here, or in the integ test.

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+
+3) TTI is recorded as Visually ready timestamp + time from visually ready to the start of the quiet window
+
+
+We do not record PageBusy metrics for now as there were some performance issues observed. 
+As such TTI can only be computed for now for browsers that support Long Tasks. 
+*/
+
+export class TTIBoomerang {
+    private fcpTime: number | null = null;
```
williazz, line 32:
```suggestion
    private fcpTime?: number;
```

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+    checkForVisualReady(): Promise<number> {
+        // Check if visually ready
+        return new Promise<number>((resolve, reject) => {
+            let timeIntervals = 0;
+            const visuallyReadyInterval: any = setInterval(() => {
+                // Check if all Visually ready timestamps are available
+                let visuallyReady = true;
+
+                if (this.fcpSupported) {
+                    if (!this.fcpTime) {
```
williazz, line 73: Suggestion: might want to do a manual null check here. It would never happen, but fcpTime === 0 would coerce to true

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+                            reject(
+                                'Insufficient visually ready timestamps to compute TTI'
+                            );
+                        }
+                    }
+                    // Visually ready so start checking for TTI
+                    clearInterval(visuallyReadyInterval);
+
+                    this.visuallyReadyTimestamp = Math.max(
+                        this.fcpTime ? this.fcpTime : 0,
```
williazz, line 114:
```suggestion
                        this.fcpTime ?? 0,
```

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+    private TTI_RESOLVE_TIMEOUT = 10000;
+
+    private FPS_THRESHOLD = 20 / (1000 / this.COLLECTION_PERIOD);
+    private LONG_TASK_THRESHOLD = 0;
+
+    // Needs to return a Promise
+    computeTimeToInteractive(): Promise<number> {
+        this.initListeners();
+        return new Promise<number>((resolve, reject) => {
+            this.checkForVisualReady().then((result) => {
```
williazz, line 56: For readability

```suggestion
            this.checkForVisualReady().then((visuallyReadTimestamp) => {
```

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+        });
+    }
+
+    checkForVisualReady(): Promise<number> {
+        // Check if visually ready
+        return new Promise<number>((resolve, reject) => {
+            let timeIntervals = 0;
+            const visuallyReadyInterval: any = setInterval(() => {
+                // Check if all Visually ready timestamps are available
+                let visuallyReady = true;
```
williazz, line 70: Question: If none of fcp, lcp, and domContentLoadedEventEnd is supported, then the visuallyReadyTimestamp is zero. Is that what TTI expects?

williazz: Oh I see, you reject this edge case below

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+                    currBucket = endBucket;
+                }
+                totalIntervals += 1;
+            }, this.CHECK_PERIOD);
+        });
+    }
+
+    initListeners(): void {
+        // record domcontentloaded if its found already
+        if (window.performance) {
```
williazz, line 223: Question: how are you capturing `domContentLoadedEventEnd` if it fires after this check?

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        if (!this.ttiTracker[type][bucket]) {
+            this.ttiTracker[type][bucket] = 0;
+        }
+
+        this.ttiTracker[type][bucket] += value;
+    }
+
+    eventListener = () => {
```
williazz, line 275: Question: where are we cleaning up this listener?

src/plugins/event-plugins/TTIPlugin.ts
williazz: I believe Boomerang does describe the behavior because it is one of many definitions of TTI (though this class is only a partial implementation of Boomerang)

src/plugins/event-plugins/TTIPlugin.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { TTIBoomerang } from '../../time-to-interactive/TTIBoomerang';
+import { TimeToInteractiveEvent } from '../../events/time-to-interactive-event';
+import { TIME_TO_INTERACTIVE_EVENT_TYPE } from '../utils/constant';
+import { PluginContext } from 'plugins/types';
+import { isLongTaskSupported } from '../../utils/common-utils';
+
+export const TTI_EVENT_PLUGIN_ID = 'time-to-interactive';
+
+export class TTIPlugin implements Plugin {
```
williazz, line 11: Question: by convention, should we be using InternalPlugin?

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+        // Use web vitals library for LCP and FCP, if supported
+        if (this.fcpSupported) {
+            onFCP((metric) => this.handleWebVitals(metric));
+        }
+        if (this.lcpSupported) {
+            onLCP((metric) => this.handleWebVitals(metric));
+        }
+
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        if (window.requestAnimationFrame(() => {})) {
```
williazz, line 248: Question: what is the purpose this this call and when would it return a falsy value (0)? Documentation states `This is a non-zero value, but you may not make any other assumptions about its value.`

https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame#return_value

src/time-to-interactive/TTIBoomerang.ts
williazz: This listener tracks FPS per 100 ms time window by incrementing a time bucket on every animation frame. We need this to calculate the quiet window, which is defined by no long tasks and FPS > 20.

It looks like requestAnimationFrame is a standard way of calculating FPS at runtime, not sure what alternate methods are more performant.

williazz: To improve performance, we could do a sample of FPS instead of calculating an exact value. But the current collection period is already fairly low at 100ms, so I would not recommend that

src/time-to-interactive/TTIBoomerang.ts
williazz: I see, if we are just checking for support for requestAnimationFrame, then should this just check that the function exists? 
```suggestion
        if (window.requestAnimationFrame) {
```

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+            // Init FPS listener if supported
+            this.fpsSupported = true;
+            this.framesPerSecondListener();
+        }
+    }
+
+    private computeTimeWindow(currTime?: number) {
+        if (typeof currTime === 'undefined') {
```
williazz, line 256: nit
```suggestion
        if (currTime === undefined) {
```

src/time-to-interactive/TTIBoomerang.ts
```diff
@@ -0,0 +1,331 @@
...
+            if (window.performance.getEntriesByType(NAVIGATION)) {
+                this.domContentLoadedEventEnd =
+                    window.performance.getEntriesByType(
+                        NAVIGATION
+                    )[0]?.domContentLoadedEventEnd;
+            }
+        }
+
+        // Use performance observer to record long tasks and domcontentloaded
+        this.eventListener();
```
williazz, line 233: Suggestion: for improved naming and separation of concerns, it would be nice if we had LongTaskObserver and NavigationObserver. This way, we can also remove the manual check above with navigationObserver.observe({type: 'navigation', buffered: true})

src/time-to-interactive/TTIBoomerang.ts
williazz: Another performance improvement is that we could wait until after visuallyReadyTimestamp is calculated before tracking FPS

williazz:
I agree with the overall design. We could definitely cleanup the logic a little and teardown the listeners, but the main blocker for me is seeing if there's any way we can improve the performance of tracking FPS.

src/plugins/event-plugins/__tests__/TTIPlugin.test.ts
williazz: Is it possible to validate the TTI value other than a assertNotNull? I'm not sure how we would could reproduce the quiet window

src/time-to-interactive/TTIBoomerang.ts
ps863: we require these checks as the algorithm requires us to pick the highest of the **supported** visual ready timestamps. if the timestamp is not supported, then we need to ignore it. 

I refactored above to make it similar to above + support the ability to check all timestamps that are supported

src/time-to-interactive/TTIBoomerang.ts
ps863: refactored with methods

src/plugins/event-plugins/TTIPlugin.ts
ps863: changed naming as behavior has been modified from other project

src/time-to-interactive/TTIBoomerang.ts
ps863: agree. changed

src/time-to-interactive/TTIBoomerang.ts
ps863: cleaned up in refactor

src/time-to-interactive/TTIBoomerang.ts
ps863: @williazz summed it up accurately. It's one component that is part of the criteria to TTI. Right now this is the most performant way to gather this data. There is some [experimental](https://udn.realityripple.com/docs/Web/API/PerformanceFrameTiming) proposals around adding this as something that can be tracked by the PerformanceObserver, but there is no ETA on the feature. 

> Another performance improvement is that we could wait until after visuallyReadyTimestamp is calculated before tracking FPS

Experimented with this a bit before. Often times, if TTI is soon after VR (visually ready), the tracker does not have enough FPS datapoints leading to the TTI being delayed more than what it should be. I believe the tracker is in practice not instantaneous in initialization, and this leads to this problem. 

In most cases, it should have little impact on the [application](https://akamai.github.io/boomerang/oss/BOOMR.plugins.Continuity.html#toc5__anchor), but I can see cases where it might. 


One option is we can remove FPS from TTI computation entirely for now. In essence a quiet window with no long tasks should have an ideal frame rate so perhaps we are more or less covering this aspect of the measurement through the long task measurements already.

src/plugins/event-plugins/__tests__/TTIPlugin.test.ts
ps863: in integ test will be hard to predict the value due to too many variables.
I believe the approach would be to mock TimeToInteractive's instance variables and then verify the functions provide the right values.

src/time-to-interactive/TTIBoomerang.ts
ps863: good catch. fixed

src/time-to-interactive/TTIBoomerang.ts
ps863: yep, there is not enough VR markers to define start of TTI measurement window

src/time-to-interactive/TTIBoomerang.ts
ps863: Added logic to stop and disconnect listeners upon TTI resolution or rejection

src/plugins/event-plugins/TTIPlugin.ts
ps863: I used plugin as we are planning not to add this into the web bundle itself and rather is just an additional plugin consumers can consume optionally. Therefore, I reasoned it did not meet the criteria of an InternalPlugin. I am open to changing this either way

src/time-to-interactive/TTIBoomerang.ts
qhanam: Got it, thanks! There's a lot of good information here that should be persisted. Please add this info to the source file, including (1) a natural language description of the behavior, and (2) expected performance impact.

> One option is we can remove FPS from TTI computation entirely for now

We could make it optional and off by default. If (A) this is not critical to measuring TTI, and (B) we are unsure about the performance, this might be the right decision.

src/plugins/event-plugins/TTIPlugin.ts
qhanam: I agree -- this class should extend InteralPlugin.

Because it lives in the aws-rum-web package, it should have the internal prefix to avoid naming conflicts.

src/plugins/event-plugins/__tests__/TTIPlugin.test.ts
qhanam: > I believe the approach would be to mock TimeToInteractive's instance variables

Yes, something like this.

The unit test coverage of *TTIPlugin.ts* looks good to me. However, we've implemented the TTI logic in *TimeToInteractive.ts*, which does not have unit tests. Unit tests for *TimeToInteractive.ts* should go in `src/time-to-interactive/__tests__/TimeToInteractive.test.ts`.

src/time-to-interactive/TimeToInteractive.ts
```diff
@@ -0,0 +1,382 @@
...
+        return new Promise<number>((resolve, reject) => {
+            const startBucket = Math.max(
+                this.computeTimeWindow(visuallyReadyTimestamp),
+                0
+            );
+            let currBucket = startBucket;
+            let acceptedIntervals = 0;
+            let totalIntervals = 0;
+
+            const tti: any = setInterval(() => {
```
williazz, line 108: Nonblocking issue: setInterval should not be trusted to exactly match the expected time interval. For precision, we should be using performance.now() whenever possible

williazz: Nevermind, it's accurate within few ms

src/time-to-interactive/TimeToInteractive.ts
```diff
@@ -0,0 +1,382 @@
...
+        visuallyReadyTimestamp: number
+    ): number {
+        /*
+        TTI is the time to the start of the TTI quiet window. 
+        Therefore, it is the sum of the time to visually ready + time from VR to the start of the quiet window
+        */
+        const timeToQuietPeriodFromVisuallyReady: number =
+            (currBucket - this.REQUIRED_ACCEPTED_INTERVALS - startBucket + 1) *
+            this.COLLECTION_PERIOD;
+
```
williazz, line 236:
```suggestion
    this.ttiTracker = {}
```

src/plugins/event-plugins/TTIPlugin.ts
ps863: changed to internal plugin

src/plugins/event-plugins/__tests__/TTIPlugin.test.ts
ps863: added some tests to exercise behavior of _TimeToInteractive.ts_

src/time-to-interactive/TTIBoomerang.ts
ps863: added description and link to this discussion at top of `TimeToInteractive.ts`

src/time-to-interactive/TimeToInteractive.ts
```diff
@@ -0,0 +1,406 @@
...
+class TimeToInteractive {
+    private onReport: TTIReportCallback;
+
+    private fcpTime?: number;
+    private lcpTime?: number;
+    private domContentLoadedEventEnd?: number;
+    private visuallyReadyTimestamp = 0;
+    private timeIntervals = 0;
+    private visualReadyInterval: NodeJS.Timeout;
+    private ttiTracker: any = {};
```
qhanam, line 70: nit: Use explicit types for the FPS and LongTask trackers.

src/time-to-interactive/TimeToInteractive.ts
```diff
@@ -0,0 +1,406 @@
...
+ * 1) Find visually ready timestamp (highest of domcontentLoadedEnd, FCP or LCP).
+ * 2) Starting from the visually ready timestamp, find a 500ms quiet window. A
+ *    quiet window has the following characteristics:
+ *      a) No Long Tasks
+ *      b) FPS is above 20 (if enabled)
+ * 3) TTI is recorded as visually ready timestamp + time from visually ready to the start of the quiet window.
+ * <p>
+ * This class measures TTI only when running in a browser that supports Long Tasks.
+ */
+class TimeToInteractive {
```
qhanam, line 61: suggestion: We are keeping track of a lot of different state in this file. Consider splitting this up into multiple files. Specifically, there seems to be two distinct phases that we could use separate classes to track: (1) waiting for visual ready, and (2) waiting for quiet period.

qhanam (approved):

PR #466, opened 2023-11-07
feat: extend addXRayTraceIdHeader to accept allow list

### LINKED ISSUE #429: [Feature Request]: Only add XRayTraceIdHeader to provided domains

williazz:
Extends addXRayTraceIdHeader to also accept an array of regular expressions. Closes #429 

---


src/plugins/event-plugins/XhrPlugin.ts
```diff
@@ -142,9 +146,9 @@ export class XhrPlugin extends MonkeyPatched<XMLHttpRequest, 'send' | 'open'> {
         const xhr: XMLHttpRequest = e.target as XMLHttpRequest;
         const xhrDetails: XhrDetails = this.xhrMap.get(xhr) as XhrDetails;
         if (xhrDetails) {
-            const endTimee = epochTime();
-            xhrDetails.trace!.end_time = endTimee;
-            xhrDetails.trace!.subsegments![0].end_time = endTimee;
+            const endTime = epochTime();
```
williazz, line 149: unrelated but this was bugging me

src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -99,13 +100,21 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
...
 
-        if (this.config.addXRayTraceIdHeader) {
+        if (this.shouldAddXrayTraceIdHeader(input)) {
             this.addXRayTraceIdHeader(input, init, argsArray, xRayTraceEvent);
         }
 
         return xRayTraceEvent;
     };
 
+    private shouldAddXrayTraceIdHeader(input: RequestInfo | URL | string) {
```
qhanam, line 110: suggestion: Rename `isXRayTraceIdHeaderEnabled` or `isTraceIdHeaderEnabled`.

docs/configuration.md
```diff
@@ -136,7 +136,7 @@ telemetries: [
...
+| addXRayTraceIdHeader | boolean or RegExp[] | `false`  | By default, the `X-Amzn-Trace-Id` header will not be added to the HTTP request. This means that the client-side trace and server-side trace will **not be linked** in X-Ray or the ServiceLens graph.<br/><br/> When this field is `true`, the `X-Amzn-Trace-Id` header will be added to HTTP requests (`XMLHttpRequest` or `fetch`). **Adding the header is dangerous and you must test your application before setting this field to `true` in a production environment.** The header could cause CORS to fail or invalidate the request's signature if the request is signed with sigv4.<br/><br/> When this field has type RegExp[], it behaves as `true` when the url matches some of the regular expressions, and behaves as `false` when there is no match.
```
qhanam, line 139:
```suggestion
| addXRayTraceIdHeader | boolean or RegExp[] | `false` | By default, the `X-Amzn-Trace-Id` header will not be added to the HTTP request. This means that the client-side trace and server-side trace will **not be linked** in X-Ray or the ServiceLens graph.<br/><br/> When this field is `true`, the `X-Amzn-Trace-Id` header will be added to HTTP requests (`XMLHttpRequest` or `fetch`).<br/><br/> When this field is an array of regular expressions (RegExp[]), the `X-Amzn-Trace-Id` header will be added only to HTTP requests that contain an URL matching one or more of the regular expressions.<br/><br/>**Adding the `X-Amzn-Trace-Id` header can cause CORS failures. Test your application before enabling this feature in a production environment.**
```

src/plugins/utils/http-utils.ts
```diff
@@ -36,7 +36,17 @@ export type HttpPluginConfig = {
     // Applications that wish to link their client and server traces by enabling
     // the X-Amzn-Trace-Id header should test their applications before enabling
     // it in a production environment.
-    addXRayTraceIdHeader: boolean;
+    addXRayTraceIdHeader: boolean | RegExp[];
+};
+
+export const shouldAddXrayTraceIdHeader = (
```
qhanam, line 42: suggestion: Rename `isXRayTraceIdHeaderEnabled` or `isTraceIdHeaderEnabled`.

src/plugins/event-plugins/XhrPlugin.ts
```diff
@@ -126,8 +127,11 @@ export class XhrPlugin extends MonkeyPatched<XMLHttpRequest, 'send' | 'open'> {
         ];
     }
 
-    private addXRayTraceIdHeader = () => {
-        return this.config.addXRayTraceIdHeader;
+    private addXRayTraceIdHeader = (url: string) => {
```
qhanam, line 130: suggestion: Rename `isXRayTraceIdHeaderEnabled` or `isTraceIdHeaderEnabled`.

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
```diff
@@ -516,6 +516,119 @@ describe('FetchPlugin tests', () => {
...
+
+        // Run
+        await fetch(URL, init);
+        plugin.disable();
+
+        // Assert
+        expect(init.headers).toEqual(undefined);
+    });
+
+    test('when addXRayTraceIdHeader is false then RequestInit is not added to the HTTP request', async () => {
```
qhanam, line 538: suggestion: I see where you're going with these "RequestInit" tests, but I don't think they are verifying quite the right properties. Consider refactoring as follows:

1. when header is disabled then header not added (cases: false, regex empty, regex no match)
2. when header is enabled then header added (cases: true, regex match)
3. when RequestInit is supplied, overwrites header
4. when RequestInit is not supplied, creates new RequestInit

You could also use the `mockFetch` methodology to verify (1) and (2) instead of looking at the RequestInit obect.

src/plugins/event-plugins/__tests__/FetchPlugin.test.ts
```diff
@@ -516,6 +516,119 @@ describe('FetchPlugin tests', () => {
...
+
+        // Run
+        await fetch(URL, init);
+        plugin.disable();
+
+        // Assert
+        expect(init.headers).toEqual(undefined);
+    });
+
+    test('when url matches no regex in addXRayTraceIdHeader then RequestInit is not added to the HTTP request', async () => {
```
qhanam, line 615: nit: Add a test for the case where the array is empty.

src/plugins/utils/http-utils.ts
qhanam: suggestion: Remove some redundancy between the XhrPlugin tests and FetchPlugin tests by unit testing this function.

qhanam (approved):

PR #482, opened 2023-12-21
chore: add check for package size regression

williazz:
Analyzes the package size diff of every pull request, broken down by file. For example, https://github.com/williazz/aws-rum-web/pull/1#issuecomment-1743927215. Moving forward, this will help us keep aws-rum-web as small as possible.  

The GitHub Action depends on third party package https://github.com/pkg-size/action

---

### DIFF
--- a/.github/workflows/package-size-report.yaml
+++ b/.github/workflows/package-size-report.yaml
@@ -1,0 +1,28 @@
+name: Package Size Report
+
+on:
+    pull_request:
+        types: [opened, ready_for_review]
+        branches: [main, release/*]
+
+permissions:
+    pull-requests: write
+
+jobs:
+    pkg-size-report:
+        name: Package Size Report
+        runs-on: ubuntu-latest
+
+        steps:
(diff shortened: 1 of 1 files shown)


williazz:
Issue: I think the default permissions for `GITHUB_TOKEN` are limited at the AWS level. Have to change when we have the chance

qhanam:
I like the idea behind this. That is, to gives us visibility into the size of the bundles we are releasing. This will be particularly helpful if and when we create multiple builds (e.g., without sigv4 or web vitals attribution).

However, I think this tool solves the wrong problem. This tool shows us the package size, when we are interested in the size of the bundle (i.e., the file _"cwr.js"_). The package doesn't account for things like dependencies (which are the bulk of the bundle size), minimization, or compression.

For example, I think we want something like the following table:

| File   |      Size      |  Compressed Size |
|----------|:-------------:|------:|
| cwr.js | 218KB | 54KB |
| cwr_min.js | 50KB | 10KB |

PR #487, opened 2024-01-02
chore: Only set credentials for sampled sessions

limhjgrace:
Currently, the RUM web client retrieves credentials for all sessions even if the `sessionSampleRate` is < 1. In order to reduce the number of unnecessary requests to Cognito and STS, we will only set/retrieve credentials for sessions that are sampled.

To do this, we will call `getSession()` to get the session and verify if it's recorded or not. Currently, a new session is created when we get a session and it doesn't exist. This PR introduces no change to this default behavior as the new boolean variable `createIfNull`, that is checked before creating a new session, is set to `true` by default. We will only set it to `false` when retrieving the session information to determine whether the web client should retrieve credentials.

I also considered checking if the eventCache had events instead of checking if the session was recorded. However, this approach was rejected as these two scenarios could diverge. We will only check if the session is recorded to keep a single source of truth.

---

### DIFF
--- a/src/event-cache/EventCache.ts
+++ b/src/event-cache/EventCache.ts
@@ -81,6 +81,13 @@ export class EventCache {
     };
 
     /**
+     * Returns true if the session is sampled, false otherwise.
+     */
+    public isSessionSampled(): boolean {
+        return this.sessionManager.isSampled();
+    }
+
+    /**
      * Add an event to the cache and reset the session timer.
      *
      * If the session is being recorded, the event will be recorded.
--- a/src/orchestration/Orchestration.ts
+++ b/src/orchestration/Orchestration.ts
@@ -415,6 +415,13 @@ export class Orchestration {
             this.config
         );
 
+        // Only retrieves and sets credentials if the session is sampled.
+        // The nil session created during initialization will have the same sampling decision as
+        // the new session created when the first event is recorded.
+        if (!this.eventCache.isSessionSampled()) {
+            return dispatch;
+        }
+
         if (this.config.identityPoolId && this.config.guestRoleArn) {
             dispatch.setAwsCredentials(
                 new Authentication(this.config)
--- a/src/sessions/SessionManager.ts
+++ b/src/sessions/SessionManager.ts
@@ -102,11 +102,20 @@ export class SessionManager {
     }
 
     /**
+     * Returns true if the session is sampled, false otherwise.
+     */
+    public isSampled(): boolean {
+        return this.session.record;
+    }
+
+    /**
      * Returns the session ID. If no session ID exists, one will be created.
      */
     public getSession(): Session {
         if (this.session.sessionId === NIL_UUID) {
             // The session does not exist. Create a new one.
+            // If it is created before the page view is recorded, the session start event metadata will
+            // not have page attributes as the page does not exist yet.
             this.createSession();
         } else if (
             this.session.sessionId !== NIL_UUID &&
@@ -219,9 +228,15 @@ export class SessionManager {
     }
 
     private createSession() {
+        // The semantics of the nil session (created during initialization) are that there is no session.
+        // We ensure the nil session and new session created right after initialization have the same sampling decision.
+        // Otherwise, we will always reevaluate the sample decision.
         this.session = {
             sessionId: v4(),
-            record: this.sample(),
+            record:
+                this.session.sessionId === NIL_UUID
+                    ? this.session.record
+                    : this.sample(),
             eventCount: 0
         };
         this.session.page = this.pageManager.getPage();
--- a/src/sessions/__tests__/SessionManager.test.ts
+++ b/src/sessions/__tests__/SessionManager.test.ts
@@ -22,6 +22,7 @@ import {
     DEFAULT_CONFIG,
     mockFetch
 } from '../../test-utils/test-utils';
+import { advanceTo } from 'jest-date-mock';
 
 global.fetch = mockFetch;
 const NAVIGATION = 'navigation';
@@ -469,6 +470,7 @@ describe('SessionManager tests', () => {
 
         // Assert
         expect(session.record).toBeTruthy();
+        expect(sessionManager.isSampled()).toBeTruthy();
     });
 
     test('when sessionSampleRate is zero then session.record is false', async () => {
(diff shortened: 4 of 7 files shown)


src/sessions/SessionManager.ts
```diff
@@ -104,8 +104,8 @@ export class SessionManager {
     /**
      * Returns the session ID. If no session ID exists, one will be created.
      */
-    public getSession(): Session {
-        if (this.session.sessionId === NIL_UUID) {
+    public getSession(createIfNull = true): Session {
+        if (createIfNull && this.session.sessionId === NIL_UUID) {
```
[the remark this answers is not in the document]
limhjgrace: Currently, the page is created before a session is created , ensuring that there are page attributes when the session is created and the session start event is recorded. In the case that `getSession()` is called when initializing the web client and it is a _nil_ session, the page hasn't been created so the session start event metadata is missing the page related attributes. This bug was detected by one of the integ tests that was verifying that a session start event was recorded with the `pageId` attribute in the metadata. 

This approach was to keep the current web client behavior of when we create sessions so that we do not introduce the new scenario of prematurely creating a new session during initialization. I'd prefer to not increase the risk of data loss that comes with (1) and (2), especially if we will stop emitting the session start event and we can clean up this variable later. Or is the concern about introducing `createIfNull` when it could become obsolete later?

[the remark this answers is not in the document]
adebayor123: +1 to Quinn's point, we will always be arbitrarily creating the sample rate (thus fetching credentials arbitrarily)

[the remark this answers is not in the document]
limhjgrace: Will refactor to make the sampling decision before creating the session in next revision!

src/sessions/SessionManager.ts
```diff
@@ -104,8 +107,8 @@ export class SessionManager {
     /**
      * Returns the session ID. If no session ID exists, one will be created.
      */
-    public getSession(): Session {
-        if (this.session.sessionId === NIL_UUID) {
+    public getSession(createIfNull = true): Session {
```
qhanam, line 110: issue: This flag variable is confusing. One alternative is to create a new method like `isSampled`.

src/sessions/SessionManager.ts
```diff
@@ -81,10 +82,12 @@ export class SessionManager {
         this.record = record;
         this.pageManager = pageManager;
 
+        this.isSampled = this.sample();
```
qhanam, line 85: issue: This changes the behavior so that the sampling decision persists for all sessions. The correct behavior is that every session is sampled randomly.

I'm a little surprised that we don't have a unit test to cover this behavior.

src/sessions/__tests__/SessionManager.test.ts
```diff
@@ -773,4 +773,23 @@ describe('SessionManager tests', () => {
...
         expect(actualSessionAttributes.domain).toEqual('overridden');
     });
+
+    test('when session is created then it will use global sampling decision', async () => {
+        const sessionManager = defaultSessionManager({
+            ...DEFAULT_CONFIG,
+            sessionSampleRate: 0.5
+        });
+
+        // This test is not foolproof. It will give a false positive 1% of the time.
```
qhanam, line 783: issue: If needed after the next revision, mock the random value generator to get rid of this source of flakiness.

src/orchestration/Orchestration.ts
```diff
@@ -415,16 +415,18 @@ export class Orchestration {
...
-            dispatch.setAwsCredentials(
-                new Authentication(this.config)
-                    .ChainAnonymousCredentialsProvider
-            );
-        } else if (this.config.identityPoolId) {
-            dispatch.setAwsCredentials(
-                new EnhancedAuthentication(this.config)
-                    .ChainAnonymousCredentialsProvider
-            );
+        if (this.eventCache.getSession(false)?.record) {
```
adebayor123, line 418: suggestion: Please add comments that explain this portion

src/sessions/SessionManager.ts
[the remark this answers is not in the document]
limhjgrace: Refactored so that sampling decision is always evaluated when creating a session. Only exception is to if the existing session is the nil session that hasn't expired - in this case, we'll use existing sampling decision.

src/sessions/SessionManager.ts
limhjgrace: Refactored to add public method.

src/sessions/SessionManager.ts
limhjgrace: Refactored and added unit test.

src/sessions/__tests__/SessionManager.test.ts
limhjgrace: Thanks for pointing this out! Used this to remove flakiness.

src/orchestration/Orchestration.ts
limhjgrace: Added comments!

src/orchestration/Orchestration.ts
```diff
@@ -415,16 +415,21 @@ export class Orchestration {
...
-            );
-        } else if (this.config.identityPoolId) {
-            dispatch.setAwsCredentials(
-                new EnhancedAuthentication(this.config)
-                    .ChainAnonymousCredentialsProvider
-            );
+        // Only retrieves and sets credentials if the session is sampled.
+        // The first nil session will have the same sampling decision as the
+        // new session created when the first event is recorded.
+        if (this.eventCache.isSessionSampled()) {
```
adebayor123, line 421: nit: we can perhaps just do `return dispatch;` if the session isn't sampled to avoid nested if else

src/sessions/SessionManager.ts
```diff
@@ -218,16 +230,27 @@ export class SessionManager {
...
     private createSession() {
+        this.sessionExpiry = new Date(
+            new Date().getTime() + this.config.sessionLengthSeconds * 1000
+        );
+
         this.session = {
             sessionId: v4(),
-            record: this.sample(),
+            record:
+                this.session.sessionId === NIL_UUID &&
```
adebayor123, line 247: question: shouldn't this be `!==` rather than `===`? what if the current session (non nil) ends, shouldn't we have a new sample rate?

src/orchestration/Orchestration.ts
limhjgrace: Refactored!

adebayor123 (approved):

src/event-cache/EventCache.ts
```diff
@@ -80,6 +80,15 @@ export class EventCache {
         }
     };
 
+    /**
+     * Returns the sampling decision of a session.
+     *
+     * @returns boolean
```
qhanam, line 86: nit: Either omit the `@returns` tag, or describe the return value.

By convention we don't typically use the `@returns` tag when the return value is already specified in the title. I believe this convention conforms with [Google's Java code guidelines](https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line) -- which we try to use in this repo -- but violates [Oracle's JavaDoc guidelines](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#@return).

src/sessions/SessionManager.ts
```diff
@@ -101,8 +101,20 @@ export class SessionManager {
...
+     */
+    public isSampled(): boolean {
+        return this.session.record;
+    }
+
     /**
      * Returns the session ID. If no session ID exists, one will be created.
+     * If a new session is created, it will record the session start event.
+     * If it is created before the page view event is recorded, the page does not exist and the session start event metadata
+     * will not have page attributes.
```
qhanam, line 117: nit(blocking): This JavaDoc doesn't adhere to convention. Begin the comment with a [summary fragment](https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment). If additional paragraphs are needed, [place a blank line](https://google.github.io/styleguide/javaguide.html#s7.1.2-javadoc-paragraphs) between the summary and the next paragraph.

Note that the purpose of the the JavaDoc is to explain the function to the caller. Any implementation details intended to provide context for a maintainer should be placed in the method itself.

src/sessions/SessionManager.ts
```diff
@@ -218,16 +230,27 @@ export class SessionManager {
...
+        this.sessionExpiry = new Date(
+            new Date().getTime() + this.config.sessionLengthSeconds * 1000
+        );
+
         this.session = {
             sessionId: v4(),
-            record: this.sample(),
+            record:
+                this.session.sessionId === NIL_UUID &&
+                new Date() < this.sessionExpiry
```
qhanam, line 248: issue: Isn't this always true, since we've updated `this.sessionExpiry` at line 240? Why do we care if the nil session has expired? The semantics of the nil session are that there is no session.

limhjgrace: > Why do we care if the nil session has expired? The semantics of the nil session are that there is no session.
I had considered a nil session to still be a valid session. Refactored as this was understanding was incorrect. 

> Isn't this always true, since we've updated this.sessionExpiry at line 240?
Removed check as this is now obsolete.

williazz (approved):
This is super useful, and we should use `isSessionSampled` as a check in `addRecordToCache` in another PR.

qhanam (approved):
Thanks; looks good!

PR #489, opened 2024-01-09
feat: capture all W3C fields in ResourceEvents

williazz:
Update the ResourceEvent schema to exactly match [PerformanceResourceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming) in aws-rum-web 2.0. I also cleaned up some tech debt, as highlighted in the comments. 

1. All durations are converted into start and end timestamps (excluding overall duration)
2. No experimental fields are included
3. Non-W3C fields such as fileType and targetUrl are dropped

In addition:
1. PerformanceResourceTimingEvents are sampled by first N FIFO, instead of random sampling. This also removes a bug in the random sampling logic when eventLimit is large.
2. Derive fileType from the initiatorType only, not the file extension.

PR #494, opened 2024-01-12
feat: capture all W3C fields in NavigationEvents

williazz:
Update the Navigation schema to exactly match [PerformanceNavigationTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming) in aws-rum-web 2.0. I have also cleaned up some tech debt, as highlighted in the comments. 

Some callouts:
1. Remove check for `load` event because the PerformanceObserver checks the buffer. I kept the tests in case there's a regression.
2. Remove support for Level1 because the level2 support is very good and very few users use Safari <v15
3. Decision to keep non-W3C initiatorType=route_change, at least for this PR. 

---


src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -8,288 +11,100 @@ import {
...
+     * Callback to record PerformanceNavigationTiming as RUM NavigationEvent
      * https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming
-     *
-     * Only the current document resource is included in the performance timeline;
-     * there is only one PerformanceNavigationTiming object in the performance timeline.
-     * https://www.w3.org/TR/navigation-timing-2/
      */
-    eventListener = () => {
-        if (performance.getEntriesByType(NAVIGATION).length === 0) {
-            this.performanceNavigationEventHandlerTimingLevel1();
```
williazz, line 77: Remove support for Level1 timing, since browser compatibility for Level2 is very strong

https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming#browser_compatibility

The only previous concern was Safari version older than v15, but customer use is extremely low for those versions.

https://www.w3schools.com/browsers/browsers_safari.asp

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -8,288 +11,100 @@ import {
...
 
-    /**
-     * loadEventEnd is populated as 0 if the web page has not loaded completely, even though LOAD has been fired.
-     * As a result, if loadEventEnd is populated, we can ignore eventListener and record the data directly.
-     * On the other hand, if not, we have to use eventListener to initializes PerformanceObserver.
-     * PerformanceObserver will act as a second check for the final load timings.
-     */
     protected onload(): void {
-        if (this.enabled) {
-            if (this.hasTheWindowLoadEventFired()) {
```
williazz, line 281: Checking for load event is not necessary anymore, since the `po` checks the buffer.

src/sessions/VirtualPageLoadTimer.ts
```diff
@@ -240,13 +240,11 @@ export class VirtualPageLoadTimer extends MonkeyPatched<
...
 
     private recordRouteChangeNavigationEvent(page: Page) {
-        const virtualPageNavigationEvent: NavigationEvent = {
-            version: '1.0.0',
-            initiatorType: 'route_change',
-            navigationType: 'navigate',
-            startTime: page.start,
+        const virtualPageNavigationEvent = {
+            initiatorType: 'navigation',
+            type: 'navigate',
```
williazz, line 245: Open question: PageManager overwrites this field with the non-W3C field 'route_change'. I recommend we avoid overwriting and copy PerformanceAPI exactly, but up for discussion

src/plugins/event-plugins/NavigationPlugin.ts
williazz: However, I have left the load event tests in case there's a regression.

src/sessions/VirtualPageLoadTimer.ts
limhjgrace: We currently don't use this field in the navigation event for any queries so there aren't any breaking changes to the system. But I would image there was a use case for this when we were implementing support for SPAs. Probably a good idea to get Reno to review this change. 

On the other hand, assuming customer experience breaking changes are okay for v2.x version bump, we could start with strictly W3C compliant fields and come back to add/overwrite this later down the road if necessary.

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -8,288 +8,99 @@ import {
...
+                loadEventEnd: e.loadEventEnd,
+                loadEventStart: e.loadEventStart,
+                redirectCount: e.redirectCount,
+                type: e.type,
+                unloadEventEnd: e.unloadEventEnd,
+                unloadEventStart: e.unloadEventStart
+            } as PerformanceNavigationTimingEvent);
+
+            // Teardown
+            this.po.disconnect();
```
limhjgrace, line 92: question: why do we have to disconnect here?

williazz: There's only one navigation event, so there's no use for the observer after after the first entry.

src/plugins/event-plugins/NavigationPlugin.ts
qhanam: What happens when the user has a browser that doesn't support PerformanceNavigationTiming?

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -8,288 +8,99 @@ import {
...
-                this.context.record(
-                    PERFORMANCE_NAVIGATION_EVENT_TYPE,
-                    eventDataNavigationTimingLevel1
-                );
+    performanceEntryHandler: PerformanceObserverCallback = (
+        list: PerformanceObserverEntryList
+    ) => {
+        list.getEntries().forEach((entry) => {
+            // Ignore
+            if (!this.enabled || this.config.ignore(entry)) {
```
qhanam, line 48:
```suggestion
            if (!this.enabled || this.config.ignore(entry)) {
```

qhanam (approved):

src/plugins/event-plugins/NavigationPlugin.ts
williazz: If navigation is not supported, then our performance observer will simply never detect and PerformanceNavigationTiming entries. I have added a runtime check for support so that if the browser is incompatible, then we do not initiate an observer.

williazz:
Final changes:
1. Require field `initiatorType` because we are always manually setting the value `navigation` or `route_change`
2. Remove value requirement `startTime`=0 because the route changes created by VirtualPageLoadTimer can have startTimes >= 0.\
3. Restore enum for field `type`
4. Add runtime check to prevent initializing an observer if PerformanceNavigationTiming is not supported

Issue(non-blocking): currently, data validation will throw errors if the PerformanceAPI ever changes the values of the fields we are expecting. For enums, data ingestion will break if any new accepted values are added. In a future change, we need to add logic so that any new unexpected values will be converted to a generic type 'other'.

williazz:
Final changes:
1. Require field `initiatorType` because we are always manually setting the value `navigation` or `route_change`
2. Add documentation for (1)
3. Remove value requirement `startTime`=0 because the route changes created by VirtualPageLoadTimer can have startTimes >= 0.
4. Restore enum for field `type`
5. Add runtime check to prevent initializing an observer if PerformanceNavigationTiming is not supported

Issue(non-blocking): currently, RUM will break if PerformanceAPI adds new accepted values the  field `type` because it has a strict enum. To prevent data loss, a future PR should add a type 'other', either here or via enrichment during ingestion. 

It's necessary to keep the enums for data validation to keep cardinality low. Also, the same problem exists with `initiatorType` for PerformanceResourceTiming.

limhjgrace (approved):

src/plugins/event-plugins/NavigationPlugin.ts
```diff
@@ -1,295 +1,108 @@
...
+                encodedBodySize: e.encodedBodySize,
+                decodedBodySize: e.decodedBodySize,
+                domComplete: e.domComplete,
+                domContentLoadedEventEnd: e.domContentLoadedEventEnd,
+                domContentLoadedEventStart: e.domContentLoadedEventStart,
+                domInteractive: e.domInteractive,
+                loadEventEnd: e.loadEventEnd,
+                loadEventStart: e.loadEventStart,
+                redirectCount: e.redirectCount,
+                type: e.type,
```
limhjgrace, line 88: question: similar question as https://github.com/aws-observability/aws-rum-web/pull/489#discussion_r1456644095 for the [`type`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/type) field

williazz: +1 https://github.com/aws-observability/aws-rum-web/pull/494#issuecomment-1896993094

PR #496, opened 2024-01-22
fix: Record 0 for headerSize if transferSize is 0

limhjgrace:
The navigation event schema validates that the `transferSize` field is non-negative. The `transferSize` field is currently calculated by: `headerSize - encodedBodySize`. However, [`headerSize` is 0 if it the resource was instantaneously retrieved from a cache](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/transferSize). This results in a negative `headerSize` and thus, an invalid event that is dropped by CWR DP. 

This PR adds a check to ensure `transferSize` is greater than 0 when calculating the `headerSize`. 

---

### DIFF
--- a/src/plugins/event-plugins/NavigationPlugin.ts
+++ b/src/plugins/event-plugins/NavigationPlugin.ts
@@ -253,7 +253,10 @@ export class NavigationPlugin extends InternalPlugin {
 
             duration: entryData.duration,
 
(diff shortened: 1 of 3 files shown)


williazz (approved):
Great catch!

qhanam (approved):

PR #498, opened 2024-01-30
fix: Invalidate cognito identity and re-try

### LINKED ISSUE #497: [Bug]: CWR: Failed to retrieve credentials from STS: TypeError: Cannot read properties of undefined (reading 'split')
### Which web client version did you detect this bug with?

v1.6.1

### What environment (build systems, module system, and framework) did you detect this bug with?

browserName: Chrome
browserVersion: 119.0.0.0
osName: Mac OS
osVersion: 10.15.7
deviceType: desktop

TypeScript v5.3.2, ECMAScript modules (ESM)

### Is your web application a single page application (SPA) or multi page application (MPA)?

MPA

### Please provide your web client configuration

 {
      sessionSampleRate: 1,
      endpoint: "https://dataplane.rum.us-east-1.amazonaws.com",
      telemetries: [
        "http",
        [
          "errors",
          {
            ignore: (event: ErrorEvent | PromiseRejectionEvent) =>
              "message" in event &&
              // ResizeObserver loop completed with undelivered notifications
              // ResizeObserver loop limit exceeded
              event.message.startsWith("ResizeObserver loop"),
          },
        ],
        "performance",
      ],
      allowCookies: true,
      cookieAttributes: {
        sameSite: "None",
      },
      enableXRay: false,
      identityPoolId: rumConfig.identityPoolId,
      guestRoleArn: rumConfig.guestRoleArn,
      sessionAttributes: {build: __APP_VERSION__, ...scope},
}

### Please describe the bug/issue

I'm seeing this error in production with a few clients and it is sporadic. RUM seems to be working fine as I'm getting the error reports and stats but is occuring daily in the logs.
Stacktrace:
```
/aws-rum-web/dist/es/dispatch/StsClient.js:98:30
/aws-rum-web/dist/es/dispatch/StsClient.js:43:22
/aws-rum-web/dist/es/dispatch/StsClient.js:24:52
/aws-rum-web/dist/es/dispatch/StsClient.js:15:57
```

I had a look at #345, but it seems unrelated.

qhanam:
When the Cognito identity id (`cwr_i`) is present in localStorage, `GetOpenIdToken` (basic authflow) can fail in a way that is not recoverable (see #497). For example, this can happen when a ResourceNotFound error is returned by Cognito.

This change (1) invalidates the identity Id when Cognito throws an error during basic authflow, (2) extends error handling to unknown errors for both authflows, and (3) adds a re-try mechanism to both authflows.

Resolves #497.

---

### DIFF
--- a/src/dispatch/BasicAuthentication.ts
+++ b/src/dispatch/BasicAuthentication.ts
@@ -27,21 +27,27 @@ export class BasicAuthentication extends Authentication {
      */
     protected AnonymousCognitoCredentialsProvider =
         async (): Promise<AwsCredentialIdentity> => {
-            return this.cognitoIdentityClient
-                .getId({
-                    IdentityPoolId: this.config.identityPoolId as string
-                })
-                .then((getIdResponse) =>
-                    this.cognitoIdentityClient.getOpenIdToken(getIdResponse)
-                )
-                .then((getOpenIdTokenResponse) =>
-                    this.stsClient.assumeRoleWithWebIdentity({
-                        RoleArn: this.config.guestRoleArn as string,
-                        RoleSessionName: 'cwr',
-                        WebIdentityToken: getOpenIdTokenResponse.Token
-                    })
-                )
-                .then((credentials: AwsCredentialIdentity) => {
+            let retries = 1;
+
+            while (true) {
+                try {
+                    const getIdResponse =
+                        await this.cognitoIdentityClient.getId({
+                            IdentityPoolId: this.config.identityPoolId as string
+                        });
+
+                    const getOpenIdTokenResponse =
+                        await this.cognitoIdentityClient.getOpenIdToken(
+                            getIdResponse
+                        );
+
+                    const credentials =
+                        await this.stsClient.assumeRoleWithWebIdentity({
+                            RoleArn: this.config.guestRoleArn as string,
+                            RoleSessionName: 'cwr',
+                            WebIdentityToken: getOpenIdTokenResponse.Token
+                        });
+
                     this.credentials = credentials;
                     try {
                         localStorage.setItem(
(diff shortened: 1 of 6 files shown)


williazz (approved):
Q: is there a retry mechanism for when credentials retrieved from localStorage are invalid, and cause a 403 to PutRumEvents?

src/dispatch/BasicAuthentication.ts
```diff
@@ -27,21 +27,27 @@ export class BasicAuthentication extends Authentication {
...
-                )
-                .then((getOpenIdTokenResponse) =>
-                    this.stsClient.assumeRoleWithWebIdentity({
-                        RoleArn: this.config.guestRoleArn as string,
-                        RoleSessionName: 'cwr',
-                        WebIdentityToken: getOpenIdTokenResponse.Token
-                    })
-                )
-                .then((credentials: AwsCredentialIdentity) => {
+            let retries = 1;
```
ps863, line 30: question: why do we allow only one retry?

ps863:
> Q: is there a retry mechanism for when credentials retrieved from localStorage are invalid, and cause a 403 to PutRumEvents?

i dont think there is. maybe in scope of another PR?

ps863 (approved):

src/dispatch/BasicAuthentication.ts
qhanam: The goal of the retry is to address the case when the identity Id is invalid. In this case, we only need on re-try after the bad identity Id has been removed from browser storage.

At a higher level, we want to minimize the impact of the RUM instrumentation on the performance of the web application. If the situation is not recoverable, then we don't want to keep re-trying.

williazz: Retries should also be extended to PutRumEvents. A future PR can consolidate the retry logic, and maybe add configuration for it.

PR #501, opened 2024-02-01
feat: retry with exponential backoff

williazz:
As a best practice, we should do exponential backoff https://docs.aws.amazon.com/prescriptive-guidance/latest/cloud-design-patterns/retry-backoff.html

---

### DIFF
--- a/src/dispatch/RetryHttpHandler.ts
+++ b/src/dispatch/RetryHttpHandler.ts
@@ -16,7 +16,7 @@ export class RetryHttpHandler implements HttpHandler {
     public constructor(
         handler: HttpHandler,
         retries: number,
-        backoff: BackoffFunction = (n) => n * 2000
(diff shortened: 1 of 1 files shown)


ps863 (approved):

qhanam:
Currently, PutRumEvents requests (i.e., those that are sent via `fetch`, and not `sendBeacon`) retry every two seconds. The maximum number of re-tries is [currently 2](https://github.com/aws-observability/aws-rum-web/blob/634467b7a0825724a21cb722b1d44600c80e3208/src/orchestration/Orchestration.ts#L87). Once both re-tries fail, the data is dropped.

With this change, PutRumEvents requests retry with exponential backoff, starting with two seconds and doubling every re-try thereafter. This means that the second re-try will happen after 4 seconds instead of after 2 seconds.

qhanam (approved):

PR #504, opened 2024-02-08
chore: migrate EventCache.canRecord to SessionManager.shouldSample

williazz:
### Purpose

Makes the session event counting logic more extensible for #505 

### Problem

SessionManager should decide if the sessionEventCount is under the sessionEventLimit, not EventCache. This PR gives that responsibility to SessionManager by removing `EventCache.canRecord` and implementing `SessionManager.shouldSample`, making it extensible to check for sessionEventLimitOverride(s) (#480). 

---

Engineering plan for #480 summarized here https://github.com/aws-observability/aws-rum-web/issues/480#issuecomment-1933212088

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #480, opened 2023-12-18
[General Inquiry]: Should errors be limited by 'sessionEventLimit'?


mitchmcleod-ew:
### What is your general inquiry?

I wasn't expecting the behaviour of errors not being recorded when a sessionEventLimit is not, got a bit caught out by it. Is there a reason for blanket preventing all events based on the limit, including errors?

Would be handy if this was configurable. Being able to still collect errors beyond when you may care about other events in the session.

williazz:
The design was to limit cost to customers, but I agree that it's inconvenient. 

Here is what I propose. Do you have any feedback? 

```typescript
const config: AwsRumConfig = {
    ...
    sessionEventLimitOverride: [
       // list of RUM event types   
       ['js-error', 0] // where zero removes the limit
       ['other-event', 5] // count against override limit before counting against overall sessionEventLimit
    ]
}
```

mitchmcleod-ew:
Yeah ok, that makes sense.

Just thinking about it, it would be a huge win for us to be able to limit all of the individual event types, just as you have there. So that looks pretty much perfect to me.

williazz:
Perfect, I'll see what I can do. 

Out of curiosity, which other events were you thinking about using this configuration for?

mitchmcleod-ew:
I would probably do something like this once this configuration is available.

```TS
const config: AwsRumConfig = {
    ...
    sessionEventLimitOverride: [
       ['js-error', 0],
       ['fetch', 0], // Assuming this is correct to capture all HTTP
       ['pageview', 20], // Allow viewing some of the user journey
       // Maybe other limits for performance events
    ]
}
```

williazz:
There are several requirements for this to be closed out

- [ ] Cleanup session sampling tech debt #504 
- [ ] Add sessionLimitOverride support to always record certain types
- [ ] Move session event counts from cookies to local storage
- [ ] Add sessionLimitOverride support to set limits on specific types

PR #505, opened 2024-02-08
feat: add sessionLimitOverride always record by type

williazz:
### Purpose

If merged, RUM users can use the below example to bypass `sessionEventLimit` and record whatever events they need. However, this PR is limited to "always record" configuration and does not support specific limits, which would need us to migrate event counts from cookies to localStorage to avoid pollution. 

```typescript
    ...
    sessionEventLimitOverride: {
        'com.amazonaws.rum.http_event': 0, // always record
        'custom_type': 1 // non-zero limits are not supported in this PR, and will be handled by sessionEventLimit
    }
}
```

Documentation updates will be addressed in a separate PR. 

### Depends on

#504 

---

Engineering plan for #480 summarized here https://github.com/aws-observability/aws-rum-web/issues/480#issuecomment-1933212088

---

### DIFF
--- a/src/event-cache/EventCache.ts
+++ b/src/event-cache/EventCache.ts
@@ -101,7 +101,7 @@ export class EventCache {
         }
         this.sessionManager.getSession(); // refresh
         if (this.isCurrentUrlAllowed()) {
-            this.sessionManager.incrementSessionEventCount();
+            this.sessionManager.countEvent();
             if (this.sessionManager.shouldSample()) {
                 this.addRecordToCache(type, eventData);
             }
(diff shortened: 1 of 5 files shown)


williazz:
Depends on #504

williazz:
Issue: I forgot about our prefix. Currently rum event types look like `com.amazonaws.rum.http_event`. Custom events will have types such as `my_custom_event` from [our own documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-RUM-custom-events.html). It doesn't make sense for customers to supply the entire prefix. This probably needs another refactoring PR to resolve the underlying issue.

Issue #493, opened 2024-01-12
[Bug]: Cannot override title attribute in metadata


stangerjm:
### Which web client version did you detect this bug with?

v1.13.3

### What environment (build systems, module system, and framework) did you detect this bug with?

TypeScript v5.0.2, Webpack v4.46.0, ECMAScript modules (ESM) and React v18.2.0

### Is your web application a single page application (SPA) or multi page application (MPA)?

None

### Please provide your web client configuration

```
{
     sessionSampleRate: 1,
     endpoint: 'https://dataplane.rum.us-west-2.amazonaws.com',
     telemetries: ['errors', 'performance', 'http'],
     allowCookies: true,
     disableAutoPageView: true,
     sessionAttributes: { title: 'override' },
}
```

### Please describe the bug/issue

When `aws-rum-web` emits metrics from the browser, it will emit the title of the current page.

In my use-case, the page's title can contain sensitive customer information that I don't want to log. Trying to override this does not work and the client will not allow me to remove that metadata value or override it.

Currently the client does allow you to override certain metadata attributes such as `domain`. See https://github.com/aws-observability/aws-rum-web/pull/453
This is not possible with other attributes. We should be able to configure what data the client sends to RUM or override attributes to obfuscate them.

ps863:
Hi @stangerjm, 
There are few concerns we had in past with opening this up to other default attributes. Will discuss with the team and get back to you

qhanam:
Thanks for raising this issue. I believe adopting #430 will resolve this. The concern was that users could get an app monitor into bad states by overriding default metadata. However, this risk seems acceptable if it solves critical use cases.

ps863:
#430 is part of v1.17.0. This should allow you to override the title. Closing issue

PR #508, opened 2024-02-09
fix: Allow title override in page attributes

qhanam:
For every page view, the web client updates the `title` attribute with the value in `window.document.title`. This means that  the `title` attribute still cannot be overridden (i.e., #430 did not actually solve #493).

This change allows the `title` attribute to be overridden when `recordPageView` is used.

**Before**
```
document.title = "sensitive";
pageManager.recordPageView({
            pageId: '/home',
            pageAttributes: {
                title: 'override'
            }
        });
// attributes: { pageId: '/home', title: 'sensitive' }
```

**After**
```
document.title = "sensitive";
pageManager.recordPageView({
            pageId: '/home',
            pageAttributes: {
                title: 'override'
            }
        });
// attributes: { pageId: '/home', title: 'override' }
```

### Discussion

Options for fixing this include the following.
1. Stop recording the page title.
1. Record title at the start of the session only.
1. Session attributes cannot be overwritten.
1. Allow page attributes to be overwritten.

All of these options have problems.
- Regarding (1), we stop giving users a common piece of information.
- Regarding (2), if the title changes during route changes, the new title will not be recorded.
- Regarding (3), this would create special classes of attributes.
- Regarding (4), requires the application to explicitly recording page views, and overwrite the title for each page view.

I am recommending option (4), because it has the least risk with respect to backwards compatibility. 

---

### DIFF
--- a/src/sessions/PageManager.ts
+++ b/src/sessions/PageManager.ts
@@ -14,7 +14,6 @@ export type Page = {
 };
 
 export type Attributes = {
-    title: string;
     pageId: string;
     parentPageId?: string;
     interaction?: number;
@@ -187,7 +186,9 @@ export class PageManager {
         customPageAttributes?: PageAttributes
     ) {
         this.attributes = {
-            title: document.title,
+            title: customPageAttributes?.pageAttributes?.title
+                ? customPageAttributes.pageAttributes.title
+                : document.title,
             pageId: page.pageId
         };
 
--- a/src/sessions/__tests__/PageManager.test.ts
+++ b/src/sessions/__tests__/PageManager.test.ts
@@ -390,6 +390,63 @@ describe('PageManager tests', () => {
         );
     });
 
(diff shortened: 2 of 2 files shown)


williazz (approved):

williazz:
Question: can the override logic be implemented by swapping these two lines instead? https://github.com/aws-observability/aws-rum-web/pull/508/files#diff-93123c2ab6bd111166aa9a50cf26e70a2ee8db05649d6f23e8a4c819a6c3a56fR208-R209

ps863 (approved):

qhanam:
> can the override logic be implemented by swapping these two lines instead?

The only other attributes that could be overridden are `pageId` and `pageTags`. These are passed as arguments to `recordPageView` along with the custom attributes, and so in the original spec we decided that the arguments would take precedence over the attributes.

PR #514, opened 2024-02-26
chore(release): 1.17.1

limhjgrace:
See README

---


ps863:
comment: We also need to update the hardcoded web client version in eventCache and/or automate getting the web client version. 
[Ref](https://github.com/aws-observability/aws-rum-web/pull/502/files)

ps863:
question: Shouldn't we be merging this onto the 1.17 branch instead of mainline since its a patch version?

PR #518, opened 2024-03-04
chore: optionally record HttpEvent.statusText

### LINKED ISSUE #516: fix: optionally write HttpEvent.statusText
Issue: `HttpEvent.statusText` should by optionally written because it's an optional field. Currently, it is always written and bloats the payload to PutRumEvents.

williazz:
closes #516

### DIFF
--- a/src/plugins/event-plugins/FetchPlugin.ts
+++ b/src/plugins/event-plugins/FetchPlugin.ts
@@ -242,9 +242,10 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
         response: Response
     ) => {
         if (this.config.recordAllRequests || !response.ok) {
+            const { status, statusText } = response;
             httpEvent.response = {
-                status: response.status,
-                statusText: response.statusText
+                status,
+                ...(statusText && { statusText })
             };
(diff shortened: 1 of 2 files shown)


qhanam:
Please add a description. Include (1) the problem being solved, and (B) how this change solves the problem.

src/plugins/event-plugins/FetchPlugin.ts
```diff
@@ -242,9 +242,10 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
         response: Response
     ) => {
         if (this.config.recordAllRequests || !response.ok) {
+            const { status, statusText } = response;
             httpEvent.response = {
-                status: response.status,
-                statusText: response.statusText
```
qhanam, line 247: question: When `statusText` is undefined, does the key appear in the PutRumEvent payload? I believe `JSON.stringify` removes keys where the value is `undefined`.

williazz: Yeah you're right, closing. A customer raised this issue on the backend, which is a separate domain.

Issue #521, opened 2024-03-11
[Feature Request]: offline support


williazz:
### Please provide your feature request

When network connection is offline, RUM should cache events and dispatch them to the collector when the connection recovers. There are a number of design requirements need to be explored and we can use this thread to consolidate them.

williazz:
One immediate issue is that RUM disables itself when the number of retries has been exhausted. This might be a bad assumption to make because the network connection might recover, and RUM will become needlessly disabled.

gmayankcse15:
Please help to prioritize the fix where RUM disables itself when the number of retries has been exhausted due to network failure or other error codes like 401, 307 etc.  RUM should not disable in such scenario and should continue to emit metrics.

williazz:
We do not want to remove this feature entirely because RUM should disable when we know that retries will never succeed, such as when credentials are invalid. 

To solve the immediate problem, we can add configuration such as `disableOnFailure: boolean`. And we'd want to make sure that this config is extensible for when provide full offline support.

gmayankcse15:
sure as an immediate solution the above config will work. Thanks.

williazz:
I've merged in the feature to keep RUM alive when dispatch fails. To close out on this issue, we need to figure out an offline strategy that customers can opt into.

1. Offline support should be opt-in because web applications by default are not configured to work offline.
2. When offline, events should be written to the [IndexDB API](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API). LocalStorage is not a good solution because they only support strings, which is not a good solution for appending structured data. Putting events back into the event cache is also not good because that is stored in memory.
3. When RUM dispatches, it should send events stored in IndexDB and the event cache. 

Open question:
1. When dispatch fails, what should happen to the payload? I recommend we put the events back in the offline cache so long as we did not receive a validation error (400).

Issue #519, opened 2024-03-05
[Feature Request]: collect metadata for connection type


williazz:
### Please provide your feature request

Connection type is an important metadata for site operators for their diagnostic workflow. For instance, a slow page load can easily be explained by mobile phone on a 2G connection. 

https://developer.mozilla.org/en-US/docs/Glossary/Effective_connection_type

PR #524, opened 2024-03-15
feat: keep alive when dispatch fails

williazz:
Addresses an immediate problem in #521, highlighted in https://github.com/aws-observability/aws-rum-web/issues/521#issuecomment-1993613490

To summarize, RUM currently disables itself when the attempt to PutRumEvents (PRT) fails and all retries have been exhausted. The intention was to avoid giving the web application needless errors when we know that RUM the network connection is disabled anyways. 

The problem is that sometimes the network connection recovers, and RUM will experience data loss because it is needlessly disabled. This PR solves that problem by adding configuration for `disableOnFail` (default true). When site operator decides to set this to false, then RUM will not disable on PRT failure, and be able to resume sending events when the network connection recovers.

This PR should be shipped out with #500 and #501 to limit the number of unnecessary retries and errors. 

In the long run, RUM should listen for `online` and `offline` DOM events ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine)). When offline, we should write events to localStorage instead of the PRT endpoint. When online, `Dispatch` should attempt to combine the current batch with events recorded offline. 

---


gmayankcse15 (commented):
Looks Good to me

qhanam:
We want to be careful not to add unnecessary configuration options. In this case, it seems like we should build a solution that just works for all users. What would be a permanent solution that doesn't involve adding a configuration option?

qhanam:
The commit message "disableOnFail" doesn't really describe what the change does. Could you re-name the commit message to be more descriptive? It sounds like we are trying to solve the problem of recovering from network problems.

williazz:
> We want to be careful not to add unnecessary configuration options. In this case, my intuition is that we can build a solution that just works for all users. What would be a permanent solution that doesn't involve adding a configuration option?

For zero configuration, we should just implement offline support summarized in this paragraph and forget about `disableOnFail`. 

> In the long run, RUM should listen for online and offline DOM events ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine)). When offline, we should write events to localStorage instead of the PRT endpoint. When online, Dispatch should attempt to combine the current batch with events recorded offline.

However, customer may still want to disable offline support for many reasons, such as limiting costs or maybe their web applications are not designed for offline use. So I'd still recommend some configuration such as `disableWhenOffline`. 

In addition, we would need to collect additional metadata such as #519 and online/offline to help site operators troubleshoot offline RUM events. And I'd still recommend #500 and #501.

gmayankcse15:
To add we should not disable RUM for 401 (Unauthorized error) , I had the usecase where due to invalid/expired token , fetch is not able to redirect to token provider website and refresh the token.

williazz:
> To add we should not disable RUM for 401 (Unauthorized error) , I had the usecase where due to invalid/expired token , fetch is not able to redirect to token provider website and refresh the token.

Looks like our docs may have to be updated. Another definition could be: disable on everything that isn't 429 or 5xx.

williazz:
Revision: I have updated the PR so that RUM only disables when Dispatch fails with 403 or 404. Those are the only status codes in which we know RUM will never succeed. In all other cases, RUM should stay alive.

qhanam:
So now, if my understanding is correct, the behavior is to:
1. When a dispatch fails (it gets three attempts), then the payload is dropped.
1. If the status code is 403 (Forbidden) or 404 (Not Found), the web client disables itself.
1. Otherwise, the web client will continue as normal, which is to attempt dispatch every 5 seconds.

I have two questions.
1. When would we receive 401 (Unauthorized) vs 403 (Forbidden)? Why is one recoverable, but not the other?
1. Long-term, is dropping payloads the behavior we want? We could add them back to the cache. Saving to localStorage is another option. Although both of those might just be adding unnecessary complexity.

Those questions aren't blocking. I think this is a good change and can be merged.

qhanam (approved):

williazz:
> When a dispatch fails (it gets three attempts), then the payload is dropped.

Almost, RUM only retries on 429 and 5xx. Then the payload is dropped. 

> When would we receive 401 (Unauthorized) vs 403 (Forbidden)? Why is one recoverable, but not the other?

RUM is not sending 401 status codes, so I didn't include it. But I can include that in case someone's proxy uses that code.

> Long-term, is dropping payloads the behavior we want? 

Dropping payloads is not the behavior we want long term, but I haven't implemented it because there are still open questions:

A. How should dropped events be reinserted into the queue and what priority should they be given against the limit? 
B. Should customers be allowed to choose an offline strategy? e.g. Reinsert into the queue OR Write to localStorage. 
C. What additional metadata can we provide to site operators to root cause "offline" events which will likely have more errors? (This can probably be resolved by providing the `Navigator.onLine` attribute)

williazz:
> To add we should not disable RUM for 401 (Unauthorized error) , I had the usecase where due to invalid/expired token , fetch is not able to redirect to token provider website and refresh the token.

> When would we receive 401 (Unauthorized) vs 403 (Forbidden)? Why is one recoverable, but not the other?

Added 401!

mayankcse15:
Looks like I missed this, although I already pointed out why we are still disabling the RUM for 401 code, 401 is a valid usecase where the proxy server might return with (Unauthorized Token) and browser might retry later with a valid token. Please help to remove 401.

PR #532, opened 2024-03-29
fix: Ignore URL construction error from invalid performance resource event

shubhsheth:
Based on this PR: https://github.com/aws-observability/aws-rum-web/pull/527

This PR fixes an error originating from recordResourceEvent() method when it tries to process an event that has an invalid name prop. While trying to construct a URL from the string, it fails with Failed to construct 'URL': Invalid URL. This causes the failure from RUM library to be logged as a RUM error event.

Adding a try/catch around the URL construction to silently ignore errors originating from unintended string formats.

---


src/plugins/event-plugins/ResourcePlugin.ts
```diff
@@ -90,8 +90,17 @@ export class ResourcePlugin extends InternalPlugin {
...
             /.*\/application\/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\/events/;
-        const entryUrl = new URL(name);
+        let entryUrl: URL;
+        try {
+            entryUrl = new URL(name);
+        } catch {
+            // Ignore failures where an invalid string causes the URL
+            // construction to fail. Otherwise RUM failures are logged as
+            // an error event.
+            return;
```
qhanam, line 100: issue: If `name` is an invalid URL, we still want to record the event.
```suggestion
```

src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
```diff
@@ -235,4 +235,28 @@ describe('ResourcePlugin tests', () => {
 
         expect(record).not.toHaveBeenCalled();
     });
+
+    test('when entry name is an invalid url then resource is ignored', async () => {
```
qhanam, line 239: issue: We want to test that when the entry name is an invalid URL, then a PerformanceResourceTiming event is recorded.

src/plugins/event-plugins/ResourcePlugin.ts
shubhsheth: @qhanam I was going off the suggestions in the previous version of changes:
> suggestion: Since this is the only place we use entryUrl, we could (1) modify this condition to only run when entryUrl is truthy, and (2) catch and ignore the error thrown by new URL(name).

Can you please clarify by providing the try/catch block you intended?

qhanam: In the current revision, we catch the error, but instead of ignoring it we exit the function without having recorded an event. The suggested change yields the intended behavior.

src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
qhanam: Is there a reason why we can't record the event if the URL is invalid?

The only reason we parse the URL is to check to see if the resource is a (PutRumEvents) call to the RUM endpoint, in which case we don't want to record the resource. Without this check, we wouldn't know or care if the URL is invalid.

src/plugins/event-plugins/ResourcePlugin.ts
williazz: I think see what Quinn is saying. Raised a PR

https://github.com/shubhsheth/aws-rum-web/pull/2

src/plugins/event-plugins/__tests__/ResourcePlugin.test.ts
williazz: https://github.com/shubhsheth/aws-rum-web/pull/2

williazz: I kinda get what what Quinn is saying. Please review https://github.com/shubhsheth/aws-rum-web/pull/3

shubhsheth: Thanks @williazz! I merged in your changes and updated unit tests.

qhanam:
Thanks @williazz and @shubhsheth ! LGTM -- just the one unit test in the approval workflow.

src/utils/__tests__/common-utils.test.ts
```diff
@@ -107,21 +107,21 @@ describe('Common utils tests', () => {
...
-        const endpointHost = 'https://dataplane.rum.us-west-2.amazonaws.com';
+        const endpointHost = 'dataplane.rum.us-west-2.amazonaws.com';
         const resourceUrl =
             'https://dataplane.rum.us-west-2.amazonaws.com/user';
         expect(utils.isPutRumEventsCall(resourceUrl, endpointHost)).toBe(false);
     });
 
     test('when url has wrong host and wrong path then it is not a PutRumEvents call', async () => {
-        const endpointHost = 'https://example.com';
+        const endpointHost = 'example.com';
```
williazz, line 124: Nice catch! 😅

williazz (approved):
Thanks

shubhsheth:
Thanks @qhanam and @williazz for working with me on this! I wanted to ask if it's okay if I create a release `1.7.2` up to this change?

qhanam:
Yes, we can do a patch release for this.

PR #541, opened 2024-04-12
chore: Update smoke tests to use enhanced auth flow

qiah:
Remove `guestRoleArn` from rum config to use enhanced auth flow in smoke tests.


qhanam:
Looks good to me. The CDN authflow is still using the basic authflow, which means the smoke tests still cover that as well.

It might make sense to add integ tests for authorization at some point, but that is out of scope here.

qhanam (approved):

PR #551, opened 2024-05-07
fix: keep alive when dispatch fails with 401

williazz:
When dispatch fails with 401, then RUM should keep alive.

RUM backend does not respond with 401 so this is change is not relevant for standard setups. However, some users have setup proxy servers with valid use cases.

For example https://github.com/aws-observability/aws-rum-web/pull/524#issuecomment-2097987765

---

### DIFF
--- a/src/dispatch/__tests__/Dispatch.test.ts
+++ b/src/dispatch/__tests__/Dispatch.test.ts
@@ -479,36 +479,6 @@ describe('Dispatch tests', () => {
         expect((dispatch as unknown as any).enabled).toBe(true);
     });
 
-    test('when a fetch request is rejected with 401 then dispatch is disabled', async () => {
-        // Init
-        (DataPlaneClient as any).mockImplementationOnce(() => ({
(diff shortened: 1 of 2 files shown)


src/dispatch/__tests__/Dispatch.test.ts
```diff
@@ -479,36 +479,6 @@ describe('Dispatch tests', () => {
         expect((dispatch as unknown as any).enabled).toBe(true);
     });
 
-    test('when a fetch request is rejected with 401 then dispatch is disabled', async () => {
```
ps863, line 482: nit: do we need reverse of this unit test?

ps863 (approved):

ps863:
LGTM

src/dispatch/__tests__/Dispatch.test.ts
williazz: It's not scalable for every status code to have a separate unit test. In the future, we should consolidate them into either "keep alive" or "disable".

Issue #557, opened 2024-05-31
[Feature Request]: Update CookieAttributes.unique behavior to control cwr_c name


ChaseFranz:
### Problem
When utilizing multiple rum clients targeting app monitors in different AWS accounts (requiring a different set of AWS credentials for signatures) we receive 403 error responses on PUT requests.

### Context
We currently have a MFE architecture where multiple MFE's utilize distinct rum clients. These clients each need to push metrics into distinct app monitors existing in different AWS accounts. The current configuration support of the client does not support this behavior without utilizing a proxy. All clients store credentials in the same cwr_c key. When a rum client signs a PUT request it is accessing credentials from different rum client session and causing 403 errors due to the invalid signature.

### Request
We require the CookieAttributes.unique property to control the name used for cwr_c similarly to how it controls cwr_s and cwr_u. If this property is set to false the key used in local storage will continue to be cwr_c and if it's set to true use the pattern cwr_c_[AppMonitor Id]. This would enable rum clients requiring distinct sets of credentials to sign PUT requests properly.

qhanam:
I think this would be a good feature addition, as it solves a common use case and should be straightforward to implement.

The [`unique` cookie attribute](https://github.com/aws-observability/aws-rum-web/blob/main/docs/configuration.md#cookieattributes) makes the `cwr_s` key unique by appending the app monitor Id. This configuration option could be extended to make the `cwr_c` key unique as well. Specifically, the app monitor Id could be added to the localStorage key written within [EnhancedAuthentication credential provider](https://github.com/aws-observability/aws-rum-web/blob/main/src/dispatch/EnhancedAuthentication.ts#L37).

williazz:
This was deployed in v1.19.0

Please reopen if you have another issue

PR #560, opened 2024-06-06
feat: support unique credential cookie names

limhjgrace:
Multiple AppMonitors from different AWS accounts cannot currently be used on the same page. Because the credential cookie name `cwr_c` is not unique to an AppMonitor, multiple web clients running on the same page attempt to read and store credential info to the same cookie, which would produce incorrect data.

This change uses the existing `cookieAttributes.unique` configuration to add the AppMonitorID as a postfix to the credential cookie `cwr_c`. When used, the credential cookie name will have the format `cwr_c_[AppMonitor Id]`, and multiple web clients sending data to app monitors in different accounts can run on the same page. 

Related FR: https://github.com/aws-observability/aws-rum-web/issues/557

---

### DIFF
--- a/src/dispatch/Authentication.ts
+++ b/src/dispatch/Authentication.ts
@@ -3,19 +3,22 @@ import { Config } from '../orchestration/Orchestration';
 import { AwsCredentialIdentity } from '@aws-sdk/types';
 import { FetchHttpHandler } from '@aws-sdk/fetch-http-handler';
 import { CRED_KEY, CRED_RENEW_MS } from '../utils/constants';
+import { getCookieName } from '../utils/cookies-utils';
 
 export abstract class Authentication {
+    protected applicationId: string;
     protected cognitoIdentityClient: CognitoIdentityClient;
     protected config: Config;
     protected credentials: AwsCredentialIdentity | undefined;
 
-    constructor(config: Config) {
+    constructor(config: Config, applicationId: string) {
         const region: string = config.identityPoolId!.split(':')[0];
         this.config = config;
         this.cognitoIdentityClient = new CognitoIdentityClient({
             fetchRequestHandler: new FetchHttpHandler(),
             region
         });
+        this.applicationId = applicationId;
     }
 
     /**
@@ -80,7 +83,15 @@ export abstract class Authentication {
             return new Promise<AwsCredentialIdentity>((resolve, reject) => {
                 let credentials: AwsCredentialIdentity;
                 try {
-                    credentials = JSON.parse(localStorage.getItem(CRED_KEY)!);
+                    credentials = JSON.parse(
+                        localStorage.getItem(
+                            getCookieName(
+                                this.config.cookieAttributes.unique,
+                                CRED_KEY,
+                                this.applicationId
+                            )
+                        )!
+                    );
                 } catch (e) {
                     // Error retrieving, decoding or parsing the cred string -- abort
                     return reject();
--- a/src/dispatch/Authentication.ts
+++ b/src/dispatch/Authentication.ts
@@ -13,10 +13,14 @@ export abstract class Authentication {
     constructor(config: Config, applicationId: string) {
         const region: string = config.identityPoolId!.split(':')[0];
         this.config = config;
-        this.cognitoIdentityClient = new CognitoIdentityClient({
-            fetchRequestHandler: new FetchHttpHandler(),
-            region
-        });
+        this.cognitoIdentityClient = new CognitoIdentityClient(
+            {
+                fetchRequestHandler: new FetchHttpHandler(),
+                region,
+                uniqueCookies: this.config.cookieAttributes.unique
+            },
+            applicationId
+        );
         this.credentialStorageKey = this.config.cookieAttributes.unique
             ? `${CRED_KEY}_${applicationId}`
             : CRED_KEY;
--- a/src/dispatch/Authentication.ts
+++ b/src/dispatch/Authentication.ts
@@ -87,11 +87,7 @@ export abstract class Authentication {
                 try {
                     credentials = JSON.parse(
                         localStorage.getItem(
-                            getCookieName(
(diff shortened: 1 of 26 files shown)


ps863:
question: will we update existing doc/provide examples in separate PR?

ps863 (approved):

ps863:
question: if there are multiple web clients running for different app monitors, user would have to make sure the unique configuration is enabled in each one for this to work correctly right?

limhjgrace:
> question: will we update existing doc/provide examples in separate PR?

Thanks for the reminder, updated existing doc in latest commit.

limhjgrace:
> question: if there are multiple web clients running for different app monitors, user would have to make sure the unique configuration is enabled in each one for this to work correctly right?

Technically if they had 2 app monitors, only one would need to configure unique cookies because they would read from different cookies. But as best practice, users should configure each app monitor, if using multiple, to configure using unique cookies.

src/dispatch/Authentication.ts
```diff
@@ -3,19 +3,22 @@ import { Config } from '../orchestration/Orchestration';
...
+    protected applicationId: string;
     protected cognitoIdentityClient: CognitoIdentityClient;
     protected config: Config;
     protected credentials: AwsCredentialIdentity | undefined;
 
-    constructor(config: Config) {
+    constructor(config: Config, applicationId: string) {
         const region: string = config.identityPoolId!.split(':')[0];
         this.config = config;
         this.cognitoIdentityClient = new CognitoIdentityClient({
```
qhanam, line 17: issue: We need unique keys for the Cognito identity as well.

src/dispatch/Authentication.ts
```diff
@@ -3,19 +3,22 @@ import { Config } from '../orchestration/Orchestration';
...
             fetchRequestHandler: new FetchHttpHandler(),
             region
         });
+        this.applicationId = applicationId;
```
qhanam, line 21:
```suggestion
        this.credentialStorageKey = this.config.cookieAttributes.unique
            ? `${CRED_KEY}_${this.applicationId}`
            : CRED_KEY;
```

src/dispatch/Authentication.ts
```diff
@@ -80,7 +83,15 @@ export abstract class Authentication {
...
                 try {
-                    credentials = JSON.parse(localStorage.getItem(CRED_KEY)!);
+                    credentials = JSON.parse(
+                        localStorage.getItem(
+                            getCookieName(
+                                this.config.cookieAttributes.unique,
+                                CRED_KEY,
+                                this.applicationId
+                            )
+                        )!
```
qhanam, line 93:
```suggestion
                            this.credentialStorageKey
```

src/sessions/SessionManager.ts
```diff
@@ -176,7 +176,11 @@ export class SessionManager {
     private createOrRenewSessionCookie(session: Session, expires: Date) {
         if (btoa) {
             storeCookie(
-                this.sessionCookieName(),
+                getCookieName(
+                    this.config.cookieAttributes.unique,
+                    SESSION_COOKIE_NAME,
+                    this.appMonitorDetails.id!
+                ),
```
qhanam, line 183: suggestion: Set the cookie name once in the constructor.
```
this.sessionCookieName = this.config.cookieAttributes.unique
            ? `${SESSION_COOKIE_NAME}_${this.appMonitorDetails.id}`
            : SESSION_COOKIE_NAME;
```
Then we can re-use the value.
```suggestion
                this.sessionCookieName,
```

qhanam:
I think this change needs to be integration tested somehow. That is, when unique keys are enabled, and multiple web clients are used concurrently, then authentication succeeds.

For example:
1. When unique keys are enabled, does authentication succeed?
2. When there are multiple app monitor instances, does authentication succeed?

I don't know if mocking the requests to Cognito and STS makes sense here. Mocking might be OK if we at least manually verify authentication succeeds for case (2). Otherwise, these might need to be run in the smoke test environment, which makes remote requests.

limhjgrace:
> I think this change needs to be integration tested somehow. That is, when unique keys are enabled, and multiple web clients are used concurrently, then authentication succeeds.
> 
> For example:
> 
> 1. When unique keys are enabled, does authentication succeed?
> 2. When there are multiple app monitor instances, does authentication succeed?
> 
> I don't know if mocking the requests to Cognito and STS makes sense here. Mocking might be OK if we at least manually verify authentication succeeds for case (2). Otherwise, these might need to be run in the smoke test environment, which makes remote requests.

Added smoke tests for applications using npm. The second web client is created after a 10 second wait from the first one to ensure credentials are retrieved and loaded separately. During smoke test dev, we discovered that we currently don't support multiple web clients for applications using CDN.

limhjgrace:
> > I think this change needs to be integration tested somehow. That is, when unique keys are enabled, and multiple web clients are used concurrently, then authentication succeeds.
> > For example:
> > 
> > 1. When unique keys are enabled, does authentication succeed?
> > 2. When there are multiple app monitor instances, does authentication succeed?
> > 
> > I don't know if mocking the requests to Cognito and STS makes sense here. Mocking might be OK if we at least manually verify authentication succeeds for case (2). Otherwise, these might need to be run in the smoke test environment, which makes remote requests.
> 
> Added smoke tests for applications using npm. The second web client is created after a 10 second wait from the first one to ensure credentials are retrieved and loaded separately. During smoke test dev, we discovered that we currently don't support multiple web clients for applications using CDN.

Smoke tests successfully pass: https://github.com/limhjgrace/aws-rum-web/actions/runs/9487042028/job/26142730870

williazz (approved):

qhanam (approved):

limhjgrace:
Smoke tests passed: https://github.com/limhjgrace/aws-rum-web/actions/runs/9847231882/job/27186743456. Need to update secret keys for main repo.

PR #565, opened 2024-06-25
[231] Add Source Map support

andy-root:
---

### DIFF
--- a/src/orchestration/Orchestration.ts
+++ b/src/orchestration/Orchestration.ts
@@ -107,6 +107,11 @@ export type CookieAttributes = {
 
 export type PartialCookieAttributes = Partial<CookieAttributes>;
 
+export type SourceMapsFetchFunction = (
+    input: RequestInfo,
+    init?: RequestInit
+) => Promise<string>;
+
 export interface Config {
     allowCookies: boolean;
     batchLimit: number;
@@ -159,6 +164,15 @@ export interface Config {
     telemetries: Telemetry[];
     useBeacon: boolean;
     userIdRetentionDays: number;
+    /**
+     * Apply source maps to error stacks. Enabling this feature will make fetch calls
+     * for javascript source and map files. If these files are cross domain then CORS
+     * headers must be included in their responses. If credentials or other headers are
+     * required then provide a configured fetch function that returns the response as
+     * a string promise.
+     */
+    sourceMapsEnabled?: boolean;
+    sourceMapsFetchFunction?: SourceMapsFetchFunction;
 }
 
 export interface PartialConfig
--- a/src/plugins/event-plugins/__tests__/JsErrorPlugin.test.ts
+++ b/src/plugins/event-plugins/__tests__/JsErrorPlugin.test.ts
@@ -1,5 +1,14 @@
+import * as StackTrace from 'stacktrace-js';
 import { JsErrorPlugin } from '../JsErrorPlugin';
-import { context, getSession, record } from '../../../test-utils/test-utils';
+import {
+    context,
+    getSession,
+    record,
+    sourceMapsOnContext,
+    sourceMapsFetchFunctionContext,
+    sourceMapsFetchFunction,
+    waitForTick
+} from '../../../test-utils/test-utils';
 import { JS_ERROR_EVENT_TYPE } from '../../utils/constant';
 
 declare global {
@@ -28,10 +37,21 @@ expect.extend({
     }
 });
 
+jest.mock('stacktrace-js', () => ({
+    fromError: jest.fn(() =>
+        Promise.resolve([
+            'at onclick (/path/to/feedback.js:6:6)',
+            'at invokeTheCallbackFunction (/path/to/EventHandlerNonNull.js:14:28)',
+            'at anonymous (/path/to/create-event-accessor.js:35:32)'
+        ])
+    )
+}));
+
 describe('JsErrorPlugin tests', () => {
     beforeEach(() => {
         record.mockClear();
         getSession.mockClear();
+        jest.clearAllMocks();
     });
 
     test('when a TypeError is thrown then the plugin records the name, message and stack', async () => {
@@ -46,6 +66,7 @@ describe('JsErrorPlugin tests', () => {
         plugin.disable();
 
         // Assert
+        await waitForTick();
         expect(record).toHaveBeenCalledTimes(1);
         expect(record.mock.calls[0][0]).toEqual(JS_ERROR_EVENT_TYPE);
         expect(record.mock.calls[0][1]).toMatchObject(
@@ -70,6 +91,7 @@ describe('JsErrorPlugin tests', () => {
         plugin.disable();
 
         // Assert
+        await waitForTick();
         expect(record).toHaveBeenCalledTimes(1);
         expect(record.mock.calls[0][0]).toEqual(JS_ERROR_EVENT_TYPE);
         expect(record.mock.calls[0][1]).toMatchObject(
@@ -106,6 +128,7 @@ describe('JsErrorPlugin tests', () => {
         plugin.disable();
 
         // Assert
+        await waitForTick();
         expect(record).toHaveBeenCalledTimes(1);
         expect(record.mock.calls[0][0]).toEqual(JS_ERROR_EVENT_TYPE);
         expect(record.mock.calls[0][1]).toMatchObject(
@@ -131,6 +154,7 @@ describe('JsErrorPlugin tests', () => {
         plugin.disable();
 
         // Assert
+        await waitForTick();
         expect(record).toHaveBeenCalledTimes(1);
         expect(record.mock.calls[0][0]).toEqual(JS_ERROR_EVENT_TYPE);
         expect((record.mock.calls[0][1] as any).stack).toEqual(undefined);
@@ -150,6 +174,7 @@ describe('JsErrorPlugin tests', () => {
         plugin.disable();
 
         // Assert
+        await waitForTick();
         expect(record).toHaveBeenCalledTimes(1);
         expect(record.mock.calls[0][0]).toEqual(JS_ERROR_EVENT_TYPE);
         expect(record.mock.calls[0][1]).toMatchObject({
@@ -177,6 +202,7 @@ describe('JsErrorPlugin tests', () => {
         plugin.disable();
 
         // Assert
+        await waitForTick();
         expect(record.mock.calls[0][0]).toEqual(JS_ERROR_EVENT_TYPE);
         expect(record.mock.calls[0][1]).toMatchObject({
             version: '1.0.0',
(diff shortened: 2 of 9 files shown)


andy-root:
RUM Source Maps

andy-root:
Apply source maps to error stacks. Enabling this feature will make fetch calls for javascript source and map files. If these files are cross domain then CORS headers must be included in their responses. If credentials or other headers are required then an optional fetch function is included.

Error events will get handled by plugins that handle errors, these include Fetch, Error and Xhr. Before each plugin records an event it will convert the Error Event to a JS Error Event. This error processing is were the Rum Event is created and the error stack is truncated. This makes for a good location to process the error stack. If the RUM client config has sourceMapsEnabled then the error's stack will be parsed and replaced with a stack written from source maps.

williazz:
Update: I have updated the configuration to target the default browser targets. This reduces the build size to the following:

Screenshot 2024-06-25 at 6 27 38 PM

As a follow up, I am working on improving support for older targets by adding polyfills via corejs; however that is still a WIP. While the polyfills work on all browsers on unix/macos, none of them work on windows yet. Those polyfills will cost us 5-20 kb, so up for discussion as well once I raise another PR.

src/plugins/event-plugins/__tests__/JsErrorPlugin.test.ts
```diff
@@ -674,6 +719,69 @@ describe('JsErrorPlugin tests', () => {
         plugin.disable();
 
         // Assert
+        await waitForTick();
```
williazz, line 722: Question: thanks for adding these. Were these causing test flakiness?

src/plugins/utils/js-error-utils.ts
```diff
@@ -56,11 +58,39 @@ const appendErrorPrimitiveDetails = (
...
+    error: globalThis.Error,
+    fetchFunction?: SourceMapsFetchFunction
+) => {
+    if (error && error.stack) {
+        try {
+            const stackFrames = await StackTrace.fromError(error, {
+                ajax: fetchFunction
+            } as StackTrace.StackTraceOptions);
+            error.stack = stackFrames.join(' \n ');
+            console.log('appendErrorSourceMaps done');
```
williazz, line 71: Nit: please delete the console.logs

src/plugins/utils/js-error-utils.ts
```diff
@@ -56,11 +58,39 @@ const appendErrorPrimitiveDetails = (
...
 
-const appendErrorObjectDetails = (
+const appendErrorSourceMaps = async (
+    error: globalThis.Error,
+    fetchFunction?: SourceMapsFetchFunction
+) => {
+    if (error && error.stack) {
+        try {
+            const stackFrames = await StackTrace.fromError(error, {
+                ajax: fetchFunction
```
williazz, line 68: Issue: What does this ajax option do? I don't see it in documentation. 

```
    export interface StackTraceOptions {
        filter?: (stackFrame: StackFrame) => boolean;
        sourceCache?: SourceCache;
        offline?: boolean;
    }
```

https://github.com/stacktracejs/stacktrace.js/blob/710bba1118d396466ee342a30b3dfd19ecbda8b5/stacktrace-js.d.ts#L12-L16

src/plugins/utils/js-error-utils.ts
```diff
@@ -56,11 +58,39 @@ const appendErrorPrimitiveDetails = (
     rumEvent.message = error.toString();
 };
 
-const appendErrorObjectDetails = (
+const appendErrorSourceMaps = async (
```
williazz, line 61: Question: why does this need to be async? Similar to the `ajax` question.

src/plugins/utils/js-error-utils.ts
```diff
@@ -56,11 +58,39 @@ const appendErrorPrimitiveDetails = (
...
+    error: globalThis.Error,
+    fetchFunction?: SourceMapsFetchFunction
+) => {
+    if (error && error.stack) {
+        try {
+            const stackFrames = await StackTrace.fromError(error, {
```
williazz, line 67: Issue: does this need its own try catch logic to capture validation errors?

From documentation: https://github.com/stacktracejs/stacktrace.js?tab=readme-ov-file#get-stack-trace-from-an-error

```
var error = new Error('BOOM!');

StackTrace.fromError(error).then(callback).catch(errback);
//===> Promise(Array[StackFrame], Error)
``

File: src/orchestration/Orchestration.ts [hunk 6]
```diff
@@ -159,6 +164,15 @@ export interface Config {
...
     useBeacon: boolean;
     userIdRetentionDays: number;
+    /**
+     * Apply source maps to error stacks. Enabling this feature will make fetch calls
+     * for javascript source and map files. If these files are cross domain then CORS
+     * headers must be included in their responses. If credentials or other headers are
+     * required then provide a configured fetch function that returns the response as
+     * a string promise.
+     */
+    sourceMapsEnabled?: boolean;
```
williazz, line 174: Issue: to clarify, does this PR actually do two things? 

1. Enhance stack traces of all JSErrors by taking on dependency `stacktrace-js`
2. Optionally enable source maps if `sourceMapsEnabled` and `sourceMapsFetchFunction` are supplied. 

Question: How are the source maps retrieved if `sourceMapsFetchFunction` is undefined?  If supplying `sourceMapsFetchFunction` is required to enable this feature, then `sourceMapsEnabled` may be redundant?

src/plugins/utils/js-error-utils.ts
williazz: Question: assuming `ajax` option works as intended, does that mean an http request is made on every single JSError? If so, is the result cached?

src/test-utils/test-utils.ts
```diff
@@ -110,6 +111,9 @@ export const getSession: jest.MockedFunction<GetSession> = jest.fn(() => ({
     eventCount: 0
 }));
 
+export const sourceMapsFetchFunction: jest.MockedFunction<SourceMapsFetchFunction> =
```
williazz, line 114: Backlog item: this needs an integration test to validate that source maps are being applied correctly

williazz:
Thanks so much for the awesome PR! I left a few high level questions and can definitely help close the gaps.

src/plugins/utils/js-error-utils.ts
williazz: Question: how does `stacktrace-js` identify local source maps?

williazz:
Of the three fatures: 
1. Enhance all JSErrors with `stacktrace-js` - this seems very close to mergable. We need better documentation on the expected diff and value add. Also we need to validate things such as package size increase; and how large JSErrors will become with the additional info since there are hard backend limits. 
2. Apply local source maps - I don't understand if this is supported out of the box. If it is, then I have the same feedback as (1). 
3. Fetch remote source maps - This needs more design based on the above comments, and probably should be a separate PR

src/plugins/event-plugins/__tests__/JsErrorPlugin.test.ts
andy-root: This was needed because the tests were finishing before the record method has called resulting in failed tests. This is a result of changing recordHttpEventWithError() to an async method. This was the recommend way I found on jest to wait for async methods to finish. https://jestjs.io/docs/tutorial-async

src/plugins/utils/js-error-utils.ts
andy-root: missed one, ok

src/plugins/utils/js-error-utils.ts
andy-root: stacktrace-js uses stacktrace-gps to make the fetch calls. The options object is passed directly to `new StackTraceGPS()`, https://github.com/stacktracejs/stacktrace.js/blob/master/stacktrace.js#L104 stacktrace-gps takes `ajax` as an option, https://github.com/stacktracejs/stacktrace-gps/blob/master/README.md

From the stacktrace-gps documentation, 
>ajax: Function (String URL => Promise(responseText)) - Function to be used for making X-Domain requests

This is an optional fetch function that can be configured to send credentials or other headers when making requests for the js and map files.

When an error is processed fetch calls for js files are made for all the urls listed in the error stack, at the end of the file there should be a link to a map file, this is also fetched. I did not notice refetching of these files again by the browser if the same error occurs.

src/plugins/utils/js-error-utils.ts
andy-root: I added this in the case that if an error does occur from the stacktrace lib that it gets recorded.

src/orchestration/Orchestration.ts
andy-root: Setting `sourceMapsEnabled` to `true` is all that is needed unless additional headers or credentials are needed when requesting the js and map file. `sourceMapsFetchFunction` is optional, users would only provide a fetch method if they need to provide additional headers or credential when the stacktrace lib requests the js and map files.

src/test-utils/test-utils.ts
andy-root: Yes, I stoped development and shared the PR so that RUM could decide if this client solution is something they want to move forward with vs. a service side solution or maybe this is a temporary solution? If we want to move forward I will work on the integration tests.

andy-root:
Comments on the Three features

For questions about local vs remote source maps, do you mean linked vs. inline? Minified js files will usually have source maps either linked at the end of the js file or the source map content will be included at the end of the js file. You either have 2 files, file.js and file.js.map or 1 larger js file that contains both the source and maps.  In either case the change in size of the error stack will be the same.

src/plugins/utils/js-error-utils.ts
williazz: Suggestion: if there are validation errors from `stacktrace-js`, then the JSErrors look like they will be dropped. We may want to record the original error instead as a backup.

andy-root: This try catch is here as a precaution is case StackTrace.fromError() fails, we would want to know so the catch error is added to the incoming error's stack . Whether StackTrace.fromError() fails or not, incoming errors are never dropped, the try catch is only attempts at replacing the incoming error's stack value, it does not return anything.

src/orchestration/Orchestration.ts
williazz: Nit / backlog item (blocking): Before merging, we need to make it clear that this is a _client side_ implementation maps: that is, update configuration names, pr title / description, and update documentation outlining the performance tradeoffs. 

RUM service team also needs to make sure the names of your new feature do not clash with any server-side implementation in the future.

williazz: No action needed, we can clean it up

williazz:
Closing because client-side source maps aren't in scope. If there is enough popularity, then we will re-open this issue and rebase the changes. Thank you Andy

PR #571, opened 2024-07-16
Stage RUM 2.0 for release

williazz:
Merge #489 and #494 into mainline in order to stage RUM 2.0 for release. When approved, I will do a FF merge to preserve the above two commits. 

---


qhanam (approved):

williazz:
With this change, the `main` branch becomes the 2.0 release branch, and we will maintain support for older 1.x versions in the release branches. If there is a bug common to 1.x and 2.x , then we will have to merge patches into main and release/1.x.x, and run releases for both. As a follow up action, I will mark older 1.x versions for deprecation to reduce the operational burden of maintaining multiple branches.

PR #590, opened 2024-10-10
feat: add csp-violation-event-plugin

anthony-ism:
This change adds a new plugin, `CspViolationPlugin` that is responsible for attaching its custom eventHandler to the `securitypolicyviolation` event. https://developer.mozilla.org/en-US/docs/Web/API/Element/securitypolicyviolation_event

---

### DIFF
--- a/src/plugins/event-plugins/CspViolationPlugin.ts
+++ b/src/plugins/event-plugins/CspViolationPlugin.ts
@@ -1,0 +1,77 @@
+import { InternalPlugin } from '../InternalPlugin';
+import { CSP_VIOLATION_EVENT_TYPE } from '../utils/constant';
+
+export const CSP_VIOLATION_EVENT_PLUGIN_ID = 'csp-violation';
+
+export type CspViolationPluginConfig = {
+    ignore: (error: SecurityPolicyViolationEvent) => boolean;
+};
+
+export type PartialCspViolationPluginConfig = {
+    ignore?: (error: SecurityPolicyViolationEvent) => boolean;
+};
+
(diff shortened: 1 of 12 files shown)


src/plugins/event-plugins/CspViolationPlugin.ts
```diff
@@ -0,0 +1,77 @@
...
+        cspViolationEvent: SecurityPolicyViolationEvent
+    ) {
+        this.context?.record(CSP_VIOLATION_EVENT_TYPE, {
+            ...cspViolationEvent,
+            version: '1.0.0'
+        });
+    }
+
+    private addEventHandler(): void {
+        window.addEventListener('securitypolicyviolation', this.eventHandler);
```
williazz, line 68: Thanks for bringing attention to this new category of RUM events. This CR looks really good and I don't see any bugs at first glance. 

However, I would like this plugin to be extensible for other security issues, and to use the best reporting tool available. From some quick research, I suspect that we should be monitoring security related issues using the ReportingObserver API. https://developer.mozilla.org/en-US/docs/Web/API/ReportingObserver

1. Could you refactor from EventListeners to ReportingObserver? 
2. I suspect we should refactor this to a generic `SecurityPlugin`, but add a configuration option if rum users want to record CSP Violations (by default disabled). This way, we can easily extend your plugin to capture other security events as necessary. 
3. Last, for knowledge share, please let me know what other security telemetries you are interested in recording, so I can try to get priority. 

Thanks

williazz:
We're planning to merge this in as a community plugin to @aws-rum/web-core. Thanks!

PR #592, opened 2024-12-10
feat: add metadata field "aws:releaseId"

williazz:
Add new configuration field `aws:releaseId` added to the metadata of every request. 

---


docs/configuration.md
```diff
@@ -19,6 +19,7 @@ For example, the config object may look similar to the following:
 | Field Name | Type | Default | Description |
 | --- | --- | --- | --- |
 | allowCookies | Boolean | `false` | Enable the web client to set and read two cookies: a session cookie named `cwr_s` and a user cookie named `cwr_u`.<br/><br/>`cwr_s` stores session data including an anonymous session ID (uuid v4) created by the web client. This allows CloudWatch RUM to compute sessionized metrics like errors per session.<br/><br/>`cwr_u` stores an anonymous user ID (uuid v4) created by the web client. This allows CloudWatch RUM to count return visitors.<br/><br/>`true`: the web client will use cookies<br/>`false`: the web client will not use cookies. |
+| aws:releaseId | string | `1.0.0` | The application release id used to find a matching source map, if any. Must match regex `^[a-zA-Z0-9_\-:/\.]+$` |
```
limhjgrace, line 22: blockers: 
1. Can we also mention the max length of 200 limit in the docs as well?
2. This field should not have a default value given customers will also have to configure the directory in S3 when they add their source map files. This field should be undefined and not added to the event metadata unless specifically set by the customer.

src/orchestration/Orchestration.ts
```diff
@@ -501,4 +505,13 @@ export class Orchestration {
...
     }
+
+    private validateConfiguration(config: Config) {
+        const awsReleaseIdTest = new RegExp(/^[a-zA-Z0-9_\-:/\.]+$/);
+        if (!awsReleaseIdTest.test(config['aws:releaseId'])) {
+            throw new Error(
+                `AWS RUM Client invalid aws:releaseID: ${this.config['aws:releaseId']} must match ${awsReleaseIdTest}`
+            );
+        }
+    }
```
limhjgrace, line 516: blocker: General rule of thumb is that we don't do data validation on web client side and only do in DP so that (1) there is single source of truth and (2) keep web client bundle size as minimal as possible.

src/orchestration/Orchestration.ts
```diff
@@ -501,4 +505,13 @@ export class Orchestration {
...
 }
```
limhjgrace, line 517: blocker: We should also update `meta-data.json` to add explicitly add this new field. We can add data validation into the schema if we want.

src/orchestration/Orchestration.ts
williazz: That sounds fine. Though this sort of data validation does not impact performance nor bundle size

src/orchestration/Orchestration.ts
williazz: Updated metadata.json

limhjgrace (approved):

limhjgrace:
note: We should also update smoke tests for this change but out of scope for this PR as we can prioritize this later/closer to launch.

docs/configuration.md
```diff
@@ -19,6 +19,7 @@ For example, the config object may look similar to the following:
 | Field Name | Type | Default | Description |
 | --- | --- | --- | --- |
 | allowCookies | Boolean | `false` | Enable the web client to set and read two cookies: a session cookie named `cwr_s` and a user cookie named `cwr_u`.<br/><br/>`cwr_s` stores session data including an anonymous session ID (uuid v4) created by the web client. This allows CloudWatch RUM to compute sessionized metrics like errors per session.<br/><br/>`cwr_u` stores an anonymous user ID (uuid v4) created by the web client. This allows CloudWatch RUM to count return visitors.<br/><br/>`true`: the web client will use cookies<br/>`false`: the web client will not use cookies. |
+| aws:releaseId | string | `undefined` | The application release id used to find a matching source map, if any. Must match regex `^[a-zA-Z0-9_\-:/\.]{1,200}$`, including size limit 200 |
```
ps863, line 22: nit: string -> String to be consistent with rest of docs

ps863 (approved):

Issue #594, opened 2024-12-16
[Bug]: RUM client ignores PORT number in the endpoint config


sandeeppoonia:
### Which web client version did you detect this bug with?

v1.13.3

### What environment (build systems, module system, and framework) did you detect this bug with?

TypeScript v5.3.3, ECMAScript modules (ESM) and React v18.2.0

### Is your web application a single page application (SPA) or multi page application (MPA)?

SPA

### Please provide your web client configuration

{
  sessionSampleRate: 1,
  signing: false,
  endpoint: `${window.location.origin}/api`,
  telemetries: ["errors", "performance", "http"],
  allowCookies: true
}

### Please describe the bug/issue

If we use a proxy for run endpoint, instead of directly going to actual RUM endpoint, and tries to run the webapp locally ( e.g. on http://localhost:4000/ ) the rum client makes requests to  http://**localhost**/api/monitors/<appID> instead of http://localhost:4000/api/monitors/<appID> .

sandeeppoonia:
The root cause is :  [in DataPlaneClient](https://github.com/aws-observability/aws-rum-web/blob/032a9f5561fea8a9789ad934009d46a9d3d7c388/src/dispatch/DataPlaneClient.ts#L114C15-L114C22) where while creating the options the port is not set. So when fetch request is made, the [port](https://github.com/aws-observability/aws-rum-web/blob/032a9f5561fea8a9789ad934009d46a9d3d7c388/src/dispatch/FetchHttpHandler.ts#L71) is not set.

qhanam:
Thanks @sandeeppoonia ! I believe your PR (#595) resolves this -- please re-open if that is not the case.

PR #595, opened 2024-12-16
Use port from endpoint url to create HttpRequestOptions: fix for 594

sandeeppoonia:
---

This is fix for https://github.com/aws-observability/aws-rum-web/issues/594 . 

These options are used to create HttpRequest (https://github.com/aws-observability/aws-rum-web/blob/032a9f5561fea8a9789ad934009d46a9d3d7c388/src/dispatch/DataPlaneClient.ts#L76) , which is then used to create the rum URL ( https://github.com/aws-observability/aws-rum-web/blob/032a9f5561fea8a9789ad934009d46a9d3d7c388/src/dispatch/FetchHttpHandler.ts#L70)

### DIFF
--- a/src/dispatch/DataPlaneClient.ts
+++ b/src/dispatch/DataPlaneClient.ts
@@ -114,7 +114,11 @@ export class DataPlaneClient {
         const options = {
             method: METHOD,
             protocol: this.config.endpoint.protocol,
-            port: this.config.endpoint.port,
+            port:
+                !this.config.endpoint.port ||
+                isNaN(parseInt(this.config.endpoint.port, 10))
+                    ? undefined
+                    : parseInt(this.config.endpoint.port, 10),
             headers: {
                 'content-type': contentType,
                 host: this.config.endpoint.host
--- a/src/dispatch/DataPlaneClient.ts
+++ b/src/dispatch/DataPlaneClient.ts
@@ -114,6 +114,7 @@ export class DataPlaneClient {
         const options = {
             method: METHOD,
             protocol: this.config.endpoint.protocol,
+            port: this.config.endpoint.port,
             headers: {
                 'content-type': contentType,
                 host: this.config.endpoint.host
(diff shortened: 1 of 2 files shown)


src/dispatch/DataPlaneClient.ts
```diff
@@ -114,6 +114,11 @@ export class DataPlaneClient {
         const options = {
             method: METHOD,
             protocol: this.config.endpoint.protocol,
+            port:
+                !this.config.endpoint.port ||
```
williazz, line 118: What's the deal with `!`?

williazz: Seems like we can simplify lines 117-118. Otherwise, looks great! Thank you

williazz, line 118: Is there coverage for this line?

Not sure what the point of `!this.config.endpoint.port` is.

src/dispatch/DataPlaneClient.ts
```diff
@@ -114,6 +114,11 @@ export class DataPlaneClient {
...
+                isNaN(parseInt(this.config.endpoint.port, 10))
+                    ? undefined
```
williazz, line 120: Is there test coverage for when the port is not a number, and becomes undefined?

qhanam (approved):
lgtm providing @williazz concerns are addressed

src/dispatch/DataPlaneClient.ts
sandeeppoonia: Almost all the other tests are for the use-case where port is not set and I have added one where port is set. I cannot add a case for invalid port because URL class does not allow that.

src/dispatch/DataPlaneClient.ts
sandeeppoonia: Yup correct! I was doing a null check but of course parseInt will return NaN in that case so its not required. In fact I will simply use Number instead of parseInt because URL class already does the checks for us and we should just rely on it to do the right thing.

src/dispatch/DataPlaneClient.ts
sandeeppoonia: I added couple of more tests to specifically test when port is not set and validated it against the code without my change.

sandeeppoonia:
@limhjgrace did you get chance to review the changes?

PR #600, opened 2025-02-17
feat: Support use of an alias field

ps863:
This PR adds support for the use of alias field to send with PutRumEvents request. Docs will follow in another PR

---

### DIFF
--- a/src/dispatch/Dispatch.ts
+++ b/src/dispatch/Dispatch.ts
@@ -224,20 +224,12 @@ export class Dispatch {
     }
 
     private createRequest(): PutRumEventsRequest {
-        if (this.config.alias) {
-            return {
-                BatchId: v4(),
-                AppMonitorDetails: this.eventCache.getAppMonitorDetails(),
-                UserDetails: this.eventCache.getUserDetails(),
-                RumEvents: this.eventCache.getEventBatch(),
-                Alias: this.config.alias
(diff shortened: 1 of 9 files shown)


src/dispatch/Dispatch.ts
```diff
@@ -224,6 +224,15 @@ export class Dispatch {
     }
 
     private createRequest(): PutRumEventsRequest {
+        if (this.config.alias) {
```
williazz, line 227: Issue(small): we need to be mindful about the number of lines we add to this package. Pretty sure we can just go ahead and add `Alias` like this. If it's undefined, then it won't be included. 

```suggestion
 private createRequest(): PutRumEventsRequest {
            return {
                BatchId: v4(),
                AppMonitorDetails: this.eventCache.getAppMonitorDetails(),
                UserDetails: this.eventCache.getUserDetails(),
                RumEvents: this.eventCache.getEventBatch(),
                Alias: this.config.alias
            };
        }
```

williazz (approved):

limhjgrace (approved):

src/dispatch/Dispatch.ts
ps863: good callout. updated

williazz:
Oh this should be a feature by the way, not a chore

Issue #615, opened 2025-03-19
[Bug]: X-Amzn-Trace-Id is not added to the header for http requests in Nuxt 3


hibira:
### Which web client version did you detect this bug with?

v1.21.0

### What environment (build systems, module system, and framework) did you detect this bug with?

nodejs 23.8.0
Nuxt3
ofetch 1.4.1

### Is your web application a single page application (SPA) or multi page application (MPA)?

SPA

### Please provide your web client configuration

import { AwsRum, type AwsRumConfig } from 'aws-rum-web';

export default defineNuxtPlugin((_nuxtApp) => {
  const runtimeConfig = useRuntimeConfig();
  if (runtimeConfig.public.rumEnabled) {
    const config: AwsRumConfig = {
      sessionSampleRate: 1,
      identityPoolId: runtimeConfig.public.rumIdentityPoolId,
      endpoint: runtimeConfig.public.rumEndpoint,
      telemetries: ["performance", "errors", ['http', { addXRayTraceIdHeader: true, recordAllRequests: true, }]],
      allowCookies: true,
      enableXRay: true,
      sessionEventLimit: -1,
    };

    const APPLICATION_ID: string = runtimeConfig.public.rumApplicationId;
    const APPLICATION_VERSION: string = '1.0.0';
    const APPLICATION_REGION: string = runtimeConfig.public.rumApplicationRegion;

    const awsRum: AwsRum = new AwsRum(
      APPLICATION_ID,
      APPLICATION_VERSION,
      APPLICATION_REGION,
      config
    );

    return {
      provide: {
        awsRum: awsRum
      }
    };
  }
});

### Please describe the bug/issue

In Nuxt3, X-Amzn-Trace-Id is not added to the http request even if addXrayTraceIdHeader: true is specified.
As a result of debugging, addamznTraceIdHeaderToInit was reached and given by 
```
init.headers [X_AMZN_TRACE_ID] = getAmznTraceIdHeaderValue (traceId, segmentID);, 
```
![Image](https://github.com/user-attachments/assets/d9f6207c-5c01-4331-8f86-6c5bc831dd65)
![Image](https://github.com/user-attachments/assets/dfaefc81-f4c8-4d08-b2f9-e6df257201a4)
![Image](https://github.com/user-attachments/assets/07b34563-4783-4eb6-bbfe-8529df111753)

[...]

PR #616, opened 2025-03-19
[bug-fix] Make it possible to grant xray traces with an http request in Nuxt3

hibira:
…
issue #615 

Resolves an issue where X_AMZN_TRACE_ID is not assigned to the http request header in nuxt3 (ofetch).
The cause is that in AddamznTraceIdHeaderToInit, a trace has been added as a property like init.headers [X_AMZN_TRACE_ID], but it is actually a Header object, so it doesn't work properly.
In contrast to this, in the case of an object that has a set method, set is used to assign a trace ID.

Before fix
Image

After fix
image

---

### DIFF
--- a/src/plugins/utils/http-utils.ts
+++ b/src/plugins/utils/http-utils.ts
@@ -168,10 +168,17 @@ export const addAmznTraceIdHeaderToInit = (
     if (!init.headers) {
         init.headers = {};
     }
-    (init.headers as any)[X_AMZN_TRACE_ID] = getAmznTraceIdHeaderValue(
-        traceId,
-        segmentId
-    );
+    if ((init.headers as any).set) {
+        (init.headers as any).set(
+            X_AMZN_TRACE_ID,
+            getAmznTraceIdHeaderValue(traceId, segmentId)
+        );
+    } else {
+        (init.headers as any)[X_AMZN_TRACE_ID] = getAmznTraceIdHeaderValue(
+            traceId,
+            segmentId
+        );
+    }
 };
 
 export const addAmznTraceIdHeaderToHeaders = (


limhjgrace (changes requested):
Thanks for the contribution! The changes LGTM but I'd like to increase test coverage for this change since it could affect all RUM events sent by the client. Can we ensure unit, integ, smoke test cover this change?

hibira:
Hi @limhjgrace .
Thanks for confirming.
I've added tests, so check them out.

### Unit Tests
Added tests in src/plugins/utils/__tests__/http-utils.test.ts to verify the addAmznTraceIdHeaderToInit function correctly handles both types of headers objects:

1. Test for headers with set method (Nuxt3 case):
   • Verifies that when a headers object has a set method, the function uses this method to add the trace header

2. Test for headers without set method (standard case):
   • Verifies that when a headers object doesn't have a set method, the function adds the trace header as a property

### Integration Tests
Added a test in src/plugins/event-plugins/__tests__/FetchPlugin.test.ts that simulates the Nuxt3 environment:
• Tests that when fetch is called with headers that have a set method, the trace header is correctly added
• Verifies that the X-Ray trace event is properly recorded

### HTML Test Page Update
Added a new button and handler to app/http_fetch_event.html that:
• Creates a Headers object (which has a set method by default)
• Makes a fetch request using this Headers object
• Allows manual verification of the trace header addition

### Smoke Test
Added a test in src/__smoke-test__/dataplane-integ.spec.ts that mocks the Nuxt3 environment by:
• Overriding the fetch API to use Headers objects
• Triggering a fetch request
• Verifying the X-Ray trace ID header is correctly added

All tests pass successfully, confirming that the fix works correctly in all scenarios, particularly in Nuxt3 applications where headers objects have a set method. The implementation is robust and doesn't 
affect existing functionality.

PR #618, opened 2025-03-26
feat: fetch with keepalive

williazz:
AWS SDK for javascript and the fetch http handler set fetch keepalive to true by default. We are using a manual implementation of fetch http handler, and should enforce best practices where we can. 

Reading
1. https://github.com/aws/aws-sdk-js-v3/issues/1003
2. https://aws.amazon.com/blogs/developer/http-keep-alive-is-on-by-default-in-modular-aws-sdk-for-javascript/
3. https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive

---

### DIFF
--- a/src/dispatch/FetchHttpHandler.ts
+++ b/src/dispatch/FetchHttpHandler.ts
@@ -88,37 +88,38 @@ export class FetchHttpHandler implements HttpHandler {
 
         const fetchRequest = new Request(url, requestOptions);
         const raceOfPromises = [
-            this.fetchFunction!.apply(window, [fetchRequest]).then(
-                (response) => {
-                    const fetchHeaders: any = response.headers;
-                    const transformedHeaders: HeaderBag = {};
+            this.fetchFunction!.apply(window, [
+                fetchRequest,
+                { keepalive: true }
+            ]).then((response) => {
+                const fetchHeaders: any = response.headers;
+                const transformedHeaders: HeaderBag = {};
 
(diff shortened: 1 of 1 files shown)


src/dispatch/FetchHttpHandler.ts
```diff
@@ -88,37 +88,38 @@ export class FetchHttpHandler implements HttpHandler {
...
-                                statusCode: response.status,
-                                body
-                            })
-                        }));
-                    }
-                    // Return the response with streaming body
-                    return {
+            this.fetchFunction!.apply(window, [
+                fetchRequest,
+                { keepalive: true }
```
williazz, line 93: This line is the only change. Unfortunately, the linter forced me to make all of these formatting changes, and I could not exclude them.

ps863 (approved):

limhjgrace (approved):

qhanam:
The decision to not use keepalive was intentional. Because want to minimize the impact of monitoring on the application, we consider monitoring data delivery "best effort" -- we do not guarantee delivery.

Is there a concrete need for keepalive? What is the impact? Will end-users now have to wait for RUM before they can close or navigate away from the page?

williazz:
> RUM before they can close or navigate away from the page

No it doesn't impact the user experience like that. It tells the browser the complete the request even if the user closes the page. It does have some consequences for our backend though, as we have discussed offline.

PR #643, opened 2025-03-31
chore: use fetch() instead of sendBeacon() in smoke tests

williazz:
## Why

Fix the smoke test for CLS. This does not have any impact on the prod behavior. 

https://github.com/williazz/aws-rum-web/actions/runs/14180465141/job/39725211063

## Root cause

Playwright's `waitForResponse` is unaware of sendBeacon API, which means it will never find the CLS event. Instead, we must use fetch. 

Previously, it seems likely that the smoke tests were using fetch instead of sendBeacon under the hood due to an unaddressed race condition. After #631 addressed the race condition, then playwright became completely blind. 

## Implementation

Set useBeacon=false in smoke tests. 

---

### DIFF
--- a/app/smoke.html
+++ b/app/smoke.html
@@ -33,7 +33,8 @@
                 endpoint: $ENDPOINT,
                 telemetries: ['performance', 'errors', 'http', 'interaction'],
                 allowCookies: true,
-                enableXRay: true
+                enableXRay: true,
+                useBeacon: false
             });
         </script>
 
--- a/smoke/smoke-test-application-CDN/smoke.html
+++ b/smoke/smoke-test-application-CDN/smoke.html
@@ -38,7 +38,8 @@
                     Authorization: `Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c`,
                     'x-api-key': 'a1b2c3d4e5f6',
                     'content-type': 'application/json'
-                }
+                },
+                useBeacon: false
             });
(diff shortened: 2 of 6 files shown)


haotiantest (commented):
Looks good to me.

ps863:
> Playwright's waitForResponse is unaware of sendBeacon API, which means it will never find the CLS event. Instead, we must use fetch.


Is this an observation or you found this in their docs? Can we link it here so future contributors know why we did this in the tests?

ps863:
question: Does this mean we have no end to end test for when `sendBeacon = true`?

williazz:
> question: Is this an observation or you found this in their docs? Can we link it here so future contributors know why we did this in the tests?

Seems that this happens to a lot of folks using playwright

https://github.com/microsoft/playwright/issues?q=is%3Aissue%20%20sendBeacon%20

This issue in particular is our exact use case (closed last april with no resolution) 

https://github.com/microsoft/playwright/issues/23635

williazz:
> question: Does this mean we have no end to end test for when sendBeacon = true?

Yes, but it seems that we were never using sendBeacon before in the smoke test. It's unclear through because the race condition is pretty tricky. These are the possibilities: 

1. Playwright can detect SendBeacon, but it stopped suddenly for some reason in this release. 
2. Playwright cannot detect SendBeacon, but the CLS event was fired via Fetch on `visibilitychange` due to race condition which Playwright was able to see. 
3. Playwright cannot detect SendBeacon, but the CLS event was fired via Fetch **on the next regular interval** which Playwright was able to see. 

Seems that (2) or (3) are more likely, but it's a little unclear honestly.

ps863:
> This issue in particular is our exact use case (closed last april with no resolution)

Should we re-open this issue? Meanwhile I think this is not blocking

ps863 (approved):

williazz:
>Should we re-open this issue? Meanwhile I think this is not blocking

Left a comment, but I don't have permission to re-open. https://github.com/microsoft/playwright/issues/23635

Issue #614, opened 2025-03-18
[Feature Request]: URL normalization


gbzd:
### Please provide your feature request

Right now, if I'm not mistaken, there's no way to normalize two different paths that may be canonically the same path: either because they are aliases of each other, or because they contain a parameter that should be ignored.

For instance, in an SPA, with a paramaterized path of `/resource/:id` I may not want to consider all values of `:id` as separate paths: if this is something like a UUID, it's almost certain that I don't want to consider them as separate paths because it will make reporting so noisy as to become useless

From what I can see, there isn't currently a way to specify an overwrite for the path for RUM reporting purposes, but this feels sorely needed.

qhanam:
If I understand correctly, this needs to be done by the application code (client-side), since CloudWatch RUM doesn't currently have the ability to group or filter URLs server side.

A hook should be added to the router of whatever framework you are using (Angular/React/Vue/etc.). When the hook is called, record the page Id and filter out any high-cardinality or sensitive data.

The aws-rum-web docs have examples for [React](https://github.com/aws-observability/aws-rum-web/blob/main/docs/cdn_react.md#instrument-routing-to-record-page-views) and [Angular](https://github.com/aws-observability/aws-rum-web/blob/main/docs/cdn_angular.md#instrument-routing-to-record-page-views)(be sure to also set [`disableAutoPageView: false`](https://github.com/aws-observability/aws-rum-web/blob/main/docs/configuration.md#configuration-options). You may also find the blog ["Using Amazon CloudWatch RUM with a React web application in five steps"](https://aws.amazon.com/blogs/mt/using-amazon-cloudwatch-rum-with-a-react-web-application-in-five-steps/) helpful. It discusses integrating CloudWatch RUM into a React application and includes a section titled "Step 4: Omit user ID from page views".

gerardaz:
So does recording them manually affect RUM data? This wasn't very clear from the documentation (or at least my understanding of it). 

I kind of assumed it will only affect the page view events, not other events. But the example given with the user ID I guess suggests that's not the case, and it affects all data?

limhjgrace:
Hi @gerardaz, yes manually recording page views (I'm assuming you're reference this doc: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-RUM-modify-snippet.html) does affect the pageId that's added to all RUM data. I think this should be enough to get you unblocked. 

[...]

PR #650, opened 2025-04-17
feat: add configuration option to normalize URLs on HTTP events (Fetch and XHR plugins)

sergioflores-j:
# Feature proposal

This PR adds a new configuration option to the HTTP telemetry.

```ts
eventURLNormalizer?: (url: string) => string;
```

> PS: Nomenclature is up to debate

This function will be invoked before recording the HTTP event in RUM. And is useful for scenarios like the following:

1. HTTP requests done for a single endpoint that contains Path Parameters:
```
/users/1234
/users/5678

Can be normalized to, e.g.:
/users/{userId}
```

This helps avoiding noise in the RUM monitor by aggregating data (request count, sessions, etc) about same endpoints in a single "row".

2. HTTP requests that contain possibly sensitive information in the URL 

```
/users/<some-sensitive-info-here>
```

3. For normalizing known services, to help finding information in the code faster

```
/users/1234

Can be normalized to, e.g.:
"My Service Name - GetUserById"
```

## Considerations

- I have been running this in a patched version in my project and did not experience any regression issues.

- I have considered not exposing this configuration as a function! 
Instead it would be receiving URL patterns and what to replace it with. (a similar configuration to `urlsToInclude`) to avoid shifting complexity to consumers. 
But that would be more complexity to be maintained here, and could not be predicting all scenarios. So I followed the idea of the `ignore` function (from JSError plugin)

- I have NOT changed all files (documentation, integration tests, etc) because I want to collect feedback for the proposal first.

---

### DIFF
--- a/src/plugins/event-plugins/FetchPlugin.ts
+++ b/src/plugins/event-plugins/FetchPlugin.ts
@@ -224,10 +224,16 @@ export class FetchPlugin extends MonkeyPatched<Window, 'fetch'> {
         init?: RequestInit
     ): HttpEvent => {
         const request = input as Request;
+        const url = resourceToUrlString(input);
+        const normalizedUrl =
+            typeof this.config.eventURLNormalizer === 'function'
+                ? this.config.eventURLNormalizer(url)
+                : url;
+
         return {
             version: '1.0.0',
             request: {
-                url: resourceToUrlString(input),
+                url: normalizedUrl,
                 method: init?.method
                     ? init.method
                     : request.method
--- a/src/plugins/event-plugins/XhrPlugin.ts
+++ b/src/plugins/event-plugins/XhrPlugin.ts
@@ -250,14 +250,27 @@ export class XhrPlugin extends MonkeyPatched<XMLHttpRequest, 'send' | 'open'> {
         return status >= 200 && status < 300;
     }
 
+    private createHttpEvent(xhrDetails: XhrDetails): HttpEvent {
+        const url = xhrDetails.url;
+
+        const normalizedUrl =
+            typeof this.config.eventURLNormalizer === 'function'
+                ? this.config.eventURLNormalizer(url)
+                : url;
+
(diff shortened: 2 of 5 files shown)


williazz:
This is a very good PR, and I've seen this pattern used a lot. It also addresses https://github.com/aws-observability/aws-rum-web/issues/545

If our purpose is to hide PID, then could keep in extensibility in mind? It might make sense to consolidate the pid scrubbing/transforming logic

sergioflores-j:
I created another branch to start proposing something reusable by all plugins (very WIP to exercise the idea of the interface, basic implementation, etc):
[Branch](https://github.com/sergioflores-j/aws-rum-web/blob/scrubbing/src/plugins/utils/http-utils.ts)
[Comparison](https://github.com/aws-observability/aws-rum-web/compare/main...sergioflores-j:aws-rum-web:scrubbing?expand=1)
> PS: Notable changes are in `src/plugins/utils/http-utils.ts` to introduce a  "dataScrubbing" configuration 😄 

---

Though I must say that while working on it, "Request URL normalization" was the only real use case I could really think of for HTTP Events. 
As there is very little data being recorded in them (i.e. no request/response headers & body, etc) scrubbing to prevent PII info being sent is often unnecessary. 
Maybe other plugins (like the error plugin from the linked issue) could benefit more from scrubbing mechanisms?

williazz:
Hi @sergioflores-j , I am rebasing this PR to get it merged. Lmk if you are still available for collab

williazz:
hi @sergioflores-j , i've rebased your pr here - https://github.com/williazz/aws-rum-web/tree/sergio/url-norm


since this is from your fork, i will have to close this pr and migrate your commits if you want it merged. lmk in the next few days if you want to rebase it yourself. thanks for the contrib

williazz:
Resolves https://github.com/aws-observability/aws-rum-web/issues/614

PR #651, opened 2025-04-18
Revert "feat: fetch with keepalive (#618)"

williazz:
Chromium based browsers enforce a 65 kb limit when keepalive is enabled. This should not be enabled by default. 

---


limhjgrace (approved):

williazz:
https://github.com/getsentry/sentry-javascript/issues/1464

PR #659, opened 2025-05-07
chore(release): 1.23.0

williazz:
Release v1.23.0

---


ps863:
comment: I think you need to update [here](https://github.com/limhjgrace/aws-rum-web/blob/main/src/event-cache/__tests__/EventCache.integ.test.ts#L9) as well

ps863 (approved):

williazz:
> comment: I think you need to update [here](https://github.com/limhjgrace/aws-rum-web/blob/main/src/event-cache/__tests__/EventCache.integ.test.ts#L9) as well

That line was removed in a previous commit

PR #662, opened 2025-05-19
fix: add prerender offset to tti events

ishajos:
We are adding an offset for prerendered pages to the tti recorded by the web client
This offset is activationStart which is the time from when the page is prerendered to when its activated (actually visited)

When a page has prerender rules, the browser makes actual requests to the specified pages. We do not want to record these requests as tti events. So we load the tti plugin for
- a normal page load  OR
- when a prerendered page is activated
We do not load tti plugin if 
- the page is in the middle of prerendering OR
- the prerendered page is not yet activated

References:
https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API

---

### DIFF
--- a/src/plugins/event-plugins/TTIPlugin.ts
+++ b/src/plugins/event-plugins/TTIPlugin.ts
@@ -7,10 +7,15 @@ export const TTI_EVENT_PLUGIN_ID = 'time-to-interactive';
 
 export class TTIPlugin extends InternalPlugin {
     protected fpsEnabled;
+    private prerenderedPageLoad;
+    private normalPageLoad;
 
     constructor(fpsMeasurementEnabled = false) {
         super(TTI_EVENT_PLUGIN_ID);
         this.fpsEnabled = fpsMeasurementEnabled;
+        this.prerenderedPageLoad = false;
+        this.normalPageLoad = true;
+        this.checkPrerenderingActivity();
     }
 
     enable(): void {
@@ -26,8 +31,50 @@ export class TTIPlugin extends InternalPlugin {
     }
 
     onload(): void {
-        onTTI(this.handleTTI, { fpsEnabled: this.fpsEnabled });
+        if (this.normalPageLoad || this.prerenderedPageLoad) {
+            this.recordTTIEvent();
+        }
     }
+
+    private recordTTIEvent = (): void => {
+        onTTI(this.handleTTI, { fpsEnabled: this.fpsEnabled });
+    };
+
+    private checkPrerenderingActivity = (): void => {
+        if (
+            typeof document !== 'undefined' &&
+            typeof document.prerendering === 'boolean' &&
+            document.prerendering
+        ) {
+            this.normalPageLoad = false;
+            document.addEventListener('prerenderingchange', () => {
+                this.prerenderedPageLoad = true;
+                this.recordTTIEvent();
+            });
+        }
+
+        if (
+            typeof performance !== 'undefined' &&
+            typeof performance.getEntriesByType === 'function'
+        ) {
+            try {
+                const entries = performance.getEntriesByType('navigation');
+                if (entries && entries.length > 0) {
+                    const navigation = entries[0];
+                    if (
+                        navigation &&
+                        navigation.activationStart &&
+                        navigation.activationStart > 0
+                    ) {
+                        this.prerenderedPageLoad = true;
+                        this.normalPageLoad = false;
+                    }
+                }
+            } catch (e) {
+                console.debug('Error accessing Performance API:', e);
+            }
+        }
+    };
 
     private handleTTI = (metric: TTIMetric): void => {
         const ttiEvent: TimeToInteractiveEvent = {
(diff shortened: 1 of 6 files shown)


src/time-to-interactive/QuietWindowSearch.ts
```diff
@@ -127,7 +129,41 @@ export class QuietWindowSearch {
...
+
+    private getPrerenderedOffset(tti: number): number {
+        const activationStart = this.getActivationStart();
+
+        if (activationStart === undefined) {
+            return tti;
+        } else {
+            const newTti = Math.floor(tti) - activationStart;
+
+            // return the offset (at least 1ms)
```
ps863, line 145: question: can you explain this?

src/plugins/event-plugins/TTIPlugin.ts
```diff
@@ -26,9 +31,51 @@ export class TTIPlugin extends InternalPlugin {
     }
 
     onload(): void {
-        onTTI(this.handleTTI, { fpsEnabled: this.fpsEnabled });
+        if (this.normalPageLoad || this.prerenderedPageLoad) {
```
ps863, line 34: question: is there a usecase here wherein we should do this prerender offset as a configuration option? with the current change, i believe any user using the latest version of RUM web client will have these adjustments made to TTI automatically.

src/time-to-interactive/QuietWindowSearch.ts
ishajos: we are subtracting the activation start from the original tti if it not undefined. if this delta is negative then we just return the original tti. if it is less than 1 then we return 1 because thats too small. otherwise we return the difference

activation start can be undefined if a browser does not support prerendering or it can be 0 if the page is not prerendered in which case we just return the original tti.

src/plugins/event-plugins/TTIPlugin.ts
ishajos: i think if a page is prerendered then we should record the offset value and not a very large value like we are doing now. so this should not be a configuration option

in the case of non-prerendered pages, activationStart would be 0 and it will record the original tti so there should not be any change in behavior

src/plugins/event-plugins/__tests__/TTIPlugin.test.ts
```diff
@@ -101,3 +115,108 @@ describe('Time to Interactive - Plugin Tests', () => {
...
+            }
+        ]);
+
+        const plugin: TTIPlugin = new TTIPlugin();
+
+        plugin.load(context);
+
+        await new Promise((resolve) => process.nextTick(resolve));
+        expect(record).toHaveBeenCalledWith(TIME_TO_INTERACTIVE_EVENT_TYPE, {
+            value: 201,
```
limhjgrace, line 218: question: why is the value both `201` here and above? shouldn't they be different (e.g. this one be 101=201-100)

src/time-to-interactive/__tests__/QuietWindowSearch.test.ts
```diff
@@ -92,4 +106,199 @@ describe('Quiet window search tests', () => {
...
+                2, 2, 2, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0
+            ]
+        };
+
+        // Advance time ahead
+        jest.advanceTimersByTime(7000);
+
+        // TTI should not be adjusted since prerendering is not supported
+        expect(ttiVal).toBe(312);
```
limhjgrace, line 139: question: where's the additional 100 coming from?

src/plugins/event-plugins/TTIPlugin.ts
limhjgrace: @ps863 I don't think these are config options that customers need to manage but more like flags used for internal logic handling. But with that being said, maybe the question is more about why these flags are needed to begin with @ishajos?

ps863: @limhjgrace yes not exactly these. more so asking on a high level of if the adjustments being made to the TTI should have an enabled or disabled flag?

does boomerang apply this adjustment by default for anyone using new version of their library? or is there some mechanism of opting in?

ishajos: we need these flags (`normalPageLoad` and `prerenderedPageLoad`) because we want a tti event to be recorded only if a user actually visits the page. When a page is prerendering, an **actual request** is made to that page and we don't want to record an event in this case.
The flags are used to check if the page is activated i.e. if the user has visited the page. In the prerendering condition check we set `normalPageLoad` to false but the event is only recorded when `prerenderingchange` is observed (page is activated)

limhjgrace: @ps863 Oh I see. In that case, unless Boomerange has a very good reason for making this opt-in, I'd rather we not do that to reduce the work that customers have to do for this feature. I thought we created this for one specific customer and they are the ones asking for this update?

limhjgrace:
question: do we need smoke tests for these changes?

ps863:
comment: As discussed offline, the present setup does not include the TTIPlugin in the cwr.js bundle. When releasing this, we need to decide if we want to add it to the bundle and the size implications of doing so. The most ideal path forward would be having different builds but the work may be out of scope for now. 

Either way, out of scope of this PR.

ishajos:
> question: do we need smoke tests for these changes?

we don't need it for these changes but would be good to add it for tti events

PR #674, opened 2025-06-18
Feature/Session Replay: 'rrweb' integrated into as SessionReplayManager

parkjoj:
SessionReplayPlugin created with corresponding unit tests.

### DIFF
--- a/src/plugins/event-plugins/SessionReplayPlugin.ts
+++ b/src/plugins/event-plugins/SessionReplayPlugin.ts
@@ -9,12 +9,7 @@ export interface eventWithTime {
     [key: string]: any;
 }
 import { InternalPlugin } from '../InternalPlugin';
-import {
-    Session,
-    RUM_SESSION_START,
-    RUM_SESSION_EXPIRE
-} from '../../sessions/SessionManager';
-import { Topic } from '../../event-bus/EventBus';
+import { Session } from '../../sessions/SessionManager';
 
 export const SESSION_REPLAY_EVENT_TYPE = 'com.amazon.rum.session_replay_event';
 
@@ -55,7 +50,7 @@ export class SessionReplayPlugin extends InternalPlugin {
      * Override getPluginId to return the expected ID format in tests
      */
     public getPluginId(): string {
-        return 'aws:rum.rrweb';
+        return SESSION_REPLAY_EVENT_TYPE;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/plugins/event-plugins/SessionReplayPlugin.ts
```diff
@@ -0,0 +1,217 @@
+import * as rrweb from 'rrweb';
+// Define the eventWithTime interface based on how it's used in the code
+export interface eventWithTime {
+    type: number;
+    data: {
```
williazz, line 5: Is it possible to be more specific with the schema? What schema does rrweb use?

src/plugins/event-plugins/SessionReplayPlugin.ts
```diff
@@ -0,0 +1,217 @@
...
+            return;
+        }
+
+        try {
+            const recordConfig = {
+                emit: (event: eventWithTime) => {
+                    this.events.push(event);
+
+                    if (this.events.length >= this.BATCH_SIZE) {
+                        this.flushEvents();
```
williazz, line 98: Reaching the batch limit seems to be the primary way we are flushing the session replay events. 

Is it possible to improve how frequently we are reporting this, such as on an interval? 

What happens to the remaining replay events when the window is closed?

src/plugins/event-plugins/SessionReplayPlugin.ts
```diff
@@ -0,0 +1,217 @@
+import * as rrweb from 'rrweb';
+// Define the eventWithTime interface based on how it's used in the code
+export interface eventWithTime {
+    type: number;
```
williazz, line 4: What is type? Is it an index?

src/plugins/event-plugins/SessionReplayPlugin.ts
```diff
@@ -0,0 +1,217 @@
...
+                void this.sendToS3(
+                    eventsToSend,
+                    this.session?.sessionId,
+                    forced
+                );
+            } else {
+                // Default behavior - send to RUM service
+                this.context.record(SESSION_REPLAY_EVENT_TYPE, {
+                    events: eventsToSend,
+                    sessionId: this.session?.sessionId
```
williazz, line 141: sessionId is already in the metadata

williazz:
Approach looks good overall, you'll need smoke and integ tests though

Also try running webpack bundle analyzer to see how many bytes this adds

PR #698, opened 2025-10-03
recreate dependent objects in orchestration when cookie config changes

anthony-nhs:
When cookie preferences are changed, not all dependent objects in orchestrator are updated, leading to incorrect data being sent to RUM, so for instance, user journey page shows no data


williazz:
Closing in favor of #740

See this comment for rationale 

https://github.com/aws-observability/aws-rum-web/issues/699#issuecomment-3684607824

PR #712, opened 2025-11-21
fix: ignore PutRumEvents for PerformanceResource events

williazz:
## Summary

PutRumEvents is not getting properly ignored when recorded as PerformanceResourceTiming under rare circumstances. In 99% of cases, PutRumEvent requests are made with fetch unless the window is closed or hidden, after which the dispatcher attempts to flush the remaining queue via `sendBeacon` API. 

1. SendBeacon requests are by default sampled, so it's rarely captured unless `other` is added to `recordAllTypes`. 

2. By default, the default resource ignore function drops all http requests if the initiatorType is fetch or xhr - (https://github.com/aws-observability/aws-rum-web/blob/main/src/plugins/utils/performance-utils.ts#L3). This is how most PutRumEvent requests are ignored. 

3. In #532, we attempted to add PutRumEvents pattern to the built-in logic that cannot be opted out of, but we copied the incorrect pattern from the test data (https://github.com/aws-observability/aws-rum-web/blob/main/src/test-utils/mock-data.ts#L168-L173), so the problem persisted. 

When PutRumEvent requests are made with SendBeacon and is sampled, then it'll get recorded. 

## Implementation

In production, we are using `/.*\/application\/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\/events/`, when we should instead be using `/.*\/appmonitors\/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/`. 

## Testing

I reproduced the issue with this configuration. After the patch, I could not see the issue anymore. 

```typescript
    sessionSampleRate: 1,
    sessionEventLimit: 0,
    telemetries: [
      "errors",
      [
        "performance",
        {
          ignore: () => false,
          recordAllTypes: ["other"],
        },
      ],
    ],
  };
```

Screenshot 2025-11-20 at 5 30 24 PM

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,7 +61,7 @@
                 "npm-run-all": "^4.1.5",
                 "prettier": "2.8.4",
                 "standard-version": "^9.3.2",
-                "testcafe": "^3.3.0",
+                "testcafe": "^3.7.2",
                 "testcafe-reporter-html": "^1.4.6",
                 "testcafe-reporter-json": "^1.0.0",
                 "ts-jest": "^29.0.3",
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
         "npm-run-all": "^4.1.5",
         "prettier": "2.8.4",
         "standard-version": "^9.3.2",
-        "testcafe": "^3.3.0",
+        "testcafe": "^3.7.2",
         "testcafe-reporter-html": "^1.4.6",
         "testcafe-reporter-json": "^1.0.0",
         "ts-jest": "^29.0.3",


williazz:
Looks like firefox integ tests are not setting up correctly anymore, though the functionality is fine

Screenshot 2025-11-21 at 2 47 49 PM

limhjgrace:
Given that we manually tested this in Firefox and this is a high sev issue, I've approved PR but let's prioritize fixing the tests.

PR #719, opened 2025-12-04
test: migrate from testcafe to playwright

williazz:
## Summary

The firefox integration test infra has stopped working, seemingly something to do with hammerhead used in testcafe 3.x. 

Screenshot 2025-12-04 at 3 31 43 PM

## Implementation

After trying to fix this for a while, I found it would be easier to migrate to playwright. 

This PR has a lot of lines but it is quite simple, since it is mostly migrating test files from `*.test.ts` over to `*.spec.ts`. And the `package.lock` diff was quite large as well. 

## Other

1. I've also re-enabled webkit (safari) integ tests, which only runs on macos. Webkit will also work on windows because playwright helps us with that, but there is no need for us to test that. 
2. I've also enabled firefox/chrome on ubuntu, windows, and macos for more complete coverage. 

## Tests

See workflow

---

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,8 +59,6 @@ jobs:
                       browser: webkit
                     - os: windows-latest
                       browser: webkit
-                    - os: ubuntu-latest
-                      browser: edge
 
         steps:
             - name: Checkout AWS RUM Web Client Repository


src/plugins/event-plugins/__integ__/DomEventPlugin.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+        const eventType = events[0].type;
+        const eventDetails = JSON.parse(events[0].details);
+
+        expect(eventType).toBe(DOM_EVENT_TYPE);
+        expect(eventDetails).toMatchObject({
+            event: 'click',
+            cssLocator: '[label="label1"]'
+        });
+    });
+});
```
limhjgrace: question: Seems like we're missing some tests, was this intentional?

1. when element id click event configured then button click is recorded
2. when client is disabled and enabled then button click is recorded
3. when two elements identified by a CSS selector are clicked then CSS selector is recorded
4. when element not identified by a CSS selector is clicked then CSS selector field is not recorded
5. when element ID and CSS selector are specified then only event for element identified by CSS selector is recorded
6. when new DOM events are registered and then a button is clicked, the event is recorded
7. when enableMutationObserver is false by default and listening for a click on a dynamically added element given an element id, the event is not recorded
8. when enableMutationObserver is false by default and listening for a click on a dynamically added element given a CSS locator, the event is not recorded
9. when enableMutationObserver is false by default and listening for a click given a CSS selector on an existing element and a dynamically added element, only one event is recorded
10. when client is disabled then click on dynamically added element is not recorded
11. when client is disabled then clicks on existing or dynamically added element are not recorded

src/plugins/event-plugins/__integ__/JsErrorPlugin.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        const eventType = events[0].type;
+        const eventDetails = JSON.parse(events[0].details);
+
+        expect(events.length).toBe(1);
+        expect(eventType).toBe(JS_ERROR_EVENT_TYPE);
+        expect(eventDetails.type).toContain('Error');
+        expect(eventDetails.message).toContain('My error message');
+    });
+});
```
limhjgrace: question: Seems like we're missing some tests, was this intentional?

1. stack trace is recorded by default
2. when a string is thrown then name and message are recorded
3. when the application records a caught error then the plugin records the error
4. when ignore function matches error then the plugin does not record the error
5. when error invoked with record method then the plugin records the error

PR #727, opened 2025-12-12
feat: add debugging mode to expose data loss scenarios

williazz:
## Summary

Misconfiguration can lead to lots of data loss scenarios, and there's not really an easy way to make sure it doesn't happen. We've seen reports for all of the following scenarios:

1. Plugins not getting loaded
2. Failure to get credentials
3. Dropped events due to insufficient sessionEventLimit or sessionSampleRate
4. Dropped events due to insufficient eventCacheSize

## Implementation

Add `debug` config setting which logs failure scenarios for all of the above use cases.

## Tests

Set up a heavy application and enabled debug mode

```typescript
    sessionSampleRate: 1,
    sessionLengthSeconds: 30,
    endpoint: "https://dataplane.rum.us-east-1.amazonaws.com",
    telemetries: [
      [
        "performance",
        {
          eventLimit: Number.MAX_SAFE_INTEGER,
          recordAllTypes: [
            "other",
            "document",
            "stylesheet",
            "image",
            "font",
            "script",
          ],
          sampleTypes: [],
          ignore: () => false,
        },
      ],
      "errors",
      "http",
    ],
    allowCookies: true,
    enableXRay: true,
    signing: false,
    sessionEventLimit: 200,
    debug: true,
    eventCacheSize: 200,
  };
```

### Insufficient in-memory queue

Screenshot 2025-12-11 at 4 33 52 PM

### Insufficient session sampling 

Screenshot 2025-12-11 at 4 35 28 PM

---

### DIFF
--- a/src/event-cache/EventCache.ts
+++ b/src/event-cache/EventCache.ts
@@ -172,22 +172,38 @@ export class EventCache {
      * Returns true if there are one or more events in the cache.
      */
     public hasEvents(): boolean {
+        // For debug mode
         if (this.config.debug && this.sessionLimitExceeded > 0) {
             const total = this.events.length + this.sessionLimitExceeded;
-            InternalLogger.warn(
-                `Dropped ${
-                    this.sessionLimitExceeded
-                } of ${total} total events (${(
-                    (this.sessionLimitExceeded / total) *
-                    100
-                ).toFixed(
-                    2
-                )}%) for the current session. Consider increasing sessionEventLimit (currently ${
-                    this.config.sessionEventLimit
-                }) or sessionSampleRate (currently ${
-                    this.config.sessionSampleRate
-                }) to avoid data loss.`
-            );
+            if (this.sessionManager.isSampled()) {
+                InternalLogger.warn(
+                    `Dropped ${
+                        this.sessionLimitExceeded
+                    } of ${total} recently observed events (${(
+                        (this.sessionLimitExceeded / total) *
+                        100
+                    ).toFixed(
+                        2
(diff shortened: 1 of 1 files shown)


src/dispatch/Dispatch.ts
```diff
@@ -251,6 +277,12 @@ export class Dispatch {
             // RUM disables only when dispatch fails and we are certain
             // that subsequent attempts will not succeed, such as when
             // credentials are invalid or the app monitor does not exist.
+            if (this.config.debug) {
+                InternalLogger.error(
+                    'Dispatch failed with status code:',
+                    e.message
```
limhjgrace: question: Is `e.message` the correct value we should be printing here?

src/event-cache/EventCache.ts
```diff
@@ -179,6 +204,24 @@ export class EventCache {
...
+                `Dropped ${
+                    this.droppedEvent
+                } of ${total} recently observed events (${(
+                    (this.droppedEvent / total) *
+                    100
+                ).toFixed(
+                    2
+                )}%) due to insufficient in-memory queue. Increase eventCacheSize to ${total} (currently ${
+                    this.config.eventCacheSize
+                }) to avoid data loss.`
```
limhjgrace: question: Is there a max eventCacheSize we should recommend in the docs? If we have an estimated size for an event e.g. 1 KB we could probably do the math for customers to let them know how much more they can increase it without impacting their web application's performance due to the browser's in-memory queue being too occupied by the RUM data.

PR #728, opened 2025-12-12
feat: increase default eventCacheSize to 1000

williazz:
## Summary

The default eventCacheSize is too small and at times is incorrectly used as a sampling enforcer. We should instead be using sessionSampleRate and sessionEventLimit for sampling. In the worst cases, this is a data loss scenario that users may not be very aware of. 

## Implementation

I chose 1000 as the new default limit, given that events are typically 1 kb in size, so 1000 x 1 kb = 1 MB. This is an acceptable temporary limit given that browsers have GB worth of memory, and this is only a temporary limit under rare scenarios that'll get evicted through the dispatch interval. 

## Testing

Updated unit tests, and ran against new `debug` option, and saw significant improvement against heavy demo app. 

Screenshot 2025-12-11 at 4 45 59 PM

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -44,7 +44,7 @@ For example, the config object may look similar to the following:
 | telemetries | [Telemetry Config Array](#telemetry-config-array) | `[]` | See [Telemetry Config Array](#telemetry-config-array) |
 | batchLimit | Number | `100` | The maximum number of events that will be sent in one batch of RUM events. |
 | dispatchInterval | Number | `5000` | The frequency (in milliseconds) in which the webclient will dispatch a batch of RUM events. RUM events are first cached and then automatically dispatched at this set interval. |
-| eventCacheSize | Number | `1000` | The maximum number of events the cache can contain before dropping events. |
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -43,7 +43,7 @@ For example, the config object may look similar to the following:
 | telemetries | [Telemetry Config Array](#telemetry-config-array) | `[]` | See [Telemetry Config Array](#telemetry-config-array) |
 | batchLimit | Number | `100` | The maximum number of events that will be sent in one batch of RUM events. |
 | dispatchInterval | Number | `5000` | The frequency (in milliseconds) in which the webclient will dispatch a batch of RUM events. RUM events are first cached and then automatically dispatched at this set interval. |
-| eventCacheSize | Number | `200` | The maximum number of events the cache can contain before dropping events. |
+| eventCacheSize | Number | `1000` | The maximum number of events the cache can contain before dropping events. Each event is typically 1KB in size, so we recommend a default limit of 1000 events -> 1 MB to balance performance against capturing all observed events. If necessary, feel free to enable debug mode to get detailed logs on how to optimize cache size depending on your application behavior. |
```
ps863: question: should we clarify the 1KB thing to be for non custom events?

williazz: it's typically 1kb due to metadata, so honestly it'll probably be the same for custom events

Issue #507, opened 2024-02-09
[Feature Request]: Slim Build


madmed88:
### Please provide your feature request

When using RUM client with a proxy endpoint (no identityPoolId, no guestRoleArn, signing: false), the client size can be optimized by getting rid of the Auth code and dependencies.
I did some testing here https://github.com/aws-observability/aws-rum-web/commit/f5a8437b63a8c871ac4deca28bc5262394638648 and found that this reduces the bundle size by more than half. This can have a significant impact on load times.

williazz:
closes https://github.com/aws-observability/aws-rum-web/issues/305#issuecomment-1335540896

williazz:
Merged #799, pending release

Issue #294, opened 2022-11-15
feat: consider accepting the navigator.userAgent string in the dataplane API


JoshBarr:
Hello 👋 

I'm implementing RUM monitoring for a library that's deployed on a number of websites. For this use-case, the final bundle size is of critical importance. I noticed that RUM is bundling `ua-parser-js` (approx 6kb gzipped) to obtain information about the navigator. Unfortunately this adds too much weight for me to include within the performance budget for my library.

Could the RUM dataplane API accept the raw user agent string as a metadata property, instead of parsing this information into separate fields on the client-side?

qhanam:
Hi Josh, agree this is a good ask. Out of curiosity, what is your budget for RUM in your library (compressed/uncompressed)?

JoshBarr:
We've written a very minimal custom RUM client that fits in about ~4kb gzipped. We only really need this RUM instance for error tracking on sessions, it doesn't need x-ray or web vitals. This library is running on pages that may have their own RUM solution already (either AWS RUM or another vendor), so we want to avoid loading multiple RUMs on the page. 

We plan on improving on this total size by shifting some functionality into a lambda that forwards requests to the monitor instead. I think we can shift sigV4 signing, authentication, and UA parsing over to the lambda. The client-side JS would then only be responsible for emitting simple event payloads to an unauthenticated endpoint. 

In general, for libraries with lots of optional functionality, I really dig it when it's possible to build your own. This gives bundlers the best possible chance of shaking out everything you're not using, and the option to skip polyfills if you don't need them:

```ts
import { baseClient } from "rum/base"
import { xray } from "rum/xray"
import { vitals } from "rum/vitals"
 
const client = baseClient.configure({
  plugins: [
    xray({ /* xray options */ }),
    vitals({ /* vitals options */}),
  ],
  // inject native implementations for things that are typically polyfilled
  browserApis: {
    uuid: crypto.randomUUID,
    sha256: (message) => crypto.subtle.digest('SHA-256', new TextEncoder().encode(message)),
  }
})

/* leanest, meanest client that fits my exact needs 🤙  */
```

williazz:
ua-parser-js removed in 3.0 #799

Issue #546, opened 2024-04-22
[Bug]:  Use of moved dependancies


de87684333:
### Which web client version did you detect this bug with?

v1.17.2

### What environment (build systems, module system, and framework) did you detect this bug with?

TypeScript v5.3.3, Webpack v5.30.0, ECMAScript modules (ESM) and Angular 17.2.3

### Is your web application a single page application (SPA) or multi page application (MPA)?

SPA

### Please provide your web client configuration

{
  allowCookies: true,
  endpoint: "https://dataplane.rum.us-west-2.amazonaws.com",
  guestRoleArn: "arn:aws:iam::000000000000:role/RUM-Monitor-us-west-2-000000000000-00xx-Unauth",
  identityPoolId: "us-west-2:00000000-0000-0000-0000-000000000000",
  sessionSampleRate: 1,
  telemetries: ['errors', 'performance', 'http']
}

### Please describe the bug/issue

When the application build, I got npm warning telling me that 4 dependancies are deprecated because the packages have moved:

npm WARN deprecated @aws-sdk/signature-v4@3.374.0: This package has moved to @smithy/signature-v4
npm WARN deprecated @aws-sdk/fetch-http-handler@3.374.0: This package has moved to @smithy/fetch-http-handler
npm WARN deprecated @aws-sdk/util-hex-encoding@3.374.0: This package has moved to @smithy/util-hex-encoding
npm WARN deprecated @aws-sdk/querystring-builder@3.374.0: This package has moved to @smithy/querystring-builder
npm WARN deprecated @aws-sdk/protocol-http@3.374.0: This package has moved to @smithy/protocol-http

When I visit the associated npm pages, it's effectively written than the packages have moved.

After checking, the aws-rum-web package.json refers deprecated packages:
        "@aws-sdk/fetch-http-handler": "^3.36.0",
        "@aws-sdk/protocol-http": "^3.36.0",
        "@aws-sdk/signature-v4": "^3.36.0",
        "@aws-sdk/util-hex-encoding": "^3.36.0",

Could you please update dependancies?

de87684333:
Any update on this one? I still get these warnings with the latest version 1.19.0

Frontrider:
I am getting dependency issues where the requested package simply does not exist. eg: string-width-cjs@4.2.3,  5.1.1 and 1.0.0 are the only versions on NPM.

williazz:
[...]

Issue #305, opened 2022-11-23
architecture: don't require authentication to post metrics


JoshBarr:
When a new user visits my site, CW RUM kicks off a series of steps before sending metrics: the browser obtains an anonymous identity from the Identity Pool, fetches a token for the identity, assumes role via STS, and stashes the credentials in localstorage. 

Once an identity is obtained, the client then performs cryptography (sigv4) on every metrics payload before sending it over the wire.

Initially I thought that the credentials obtained from Cognito were used to establish a tracking session, but looking at the code, the generated user ID in the `cwr_u` cookie, and the session ID in the session cookie are locally generated UUIDs, separate to the Cognito Identity.

Typically RUM solutions are fire-and-forget, not requiring coordination between the collector endpoint and the client. Instead of credentials, the origin URL, device fingerprint (user agent etc) and the monitor ID are sent to the endpoint. 

Could you consider an architecture where the collector endpoints are unauthenticated, and take care of any required AWS service authorisation internally? 

Client side libraries should aim to do the least amount of work possible, in the fewest bytes.

qhanam:
Hi @JoshBarr thanks for this feedback/feature request! I agree with your points and the short answer is yes, the CloudWatch RUM team does consider (and appreciate!) such requests.

However, let me also shed some light on why we do this, and offer a workaround.

### Why CloudWatch RUM uses authentication and authorization

Like most AWS services, the CW RUM endpoint uses AWS authentication and authorization. This gives customers full control over who sends data to their CW RUM app monitor. This is especially useful [if your application has authenticated users](https://aws.amazon.com/blogs/mt/how-to-isolate-signed-in-users-from-guest-users-within-amazon-cloudwatch-rum/) and you want to gain assurance that only those users can send RUM data to your app monitor.

For example, if your organization is using CW RUM to monitor the front-end of a critical internal system which is only accessible to employees, it might be important to authenticate the RUM data.

However, if you want to collect data from anonymous users, authentication is a barrier. We work around this by using Cognito's unauthenticated identity, which effectively drops authentication but retains authorization.

### Workaround using a proxy

To avoid doing the signing step on the end-user device, you can use a proxy. There are CW RUM customers who already do this. The rough steps for this are:
1. Create a proxy. We recommend using API Gateway + Lambda.
1. The proxy holds AWS credentials which can send the app monitor data via the PutRumEvents operation.
[...]

PR #799, opened 2026-04-08
AWS RUM Web v3.0

williazz:
## Summary

Restructure `aws-rum-web` into a Lerna monorepo with three packages and introduce a slim distribution that is **77% smaller** (10 KB gzipped). With session replay, the old 2.x module would have been 155+70=225 KB parsed at least. 

Improvements

1. aws-rum-web is much leaner because we removed ua-parser, unused auth modules, and bumped es target to es17. This would have cut the total package from 150KB -> 100KB; however, we also added session replay plugin (with RRWeb dependency) which adds 60KB, so our new total package size is 160KB. 
2. We are introducing two new packages `@aws-rum/web-slim` which is an imperative rewrite, and starts out at just 33 KB (10KB gzipped). Common logic is migrated over to `@aws-rum/web-core`
3. SessionReplay is also going out with this package, and natively shipped with the legacy declarative aws-rum-web distro. We will continue supporting the old format but will recommend that advanced users migrate over to the slim distribution. 
4. Metadata is compressed by 90% because we are consolidating repeated fields in optional top-level field `SessionMetadata`. 
5. User agent parsing is also migrated to server-side and the native [userAgentData](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData) API if available
6. Compression by gzip is also enabled by default. This cuts network size down to <1KB typically. 

Doc updates will not be released until day of deployment so as not to confuse readers

## Breaking Changes

All of these are fully backwards compatible, but are still significant changes in expected behavior. 

- **Monorepo restructure** ([#507](https://github.com/aws-observability/aws-rum-web/issues/507), [#305](https://github.com/aws-observability/aws-rum-web/issues/305)) — single package → `@aws-rum/web-core`, `aws-rum-web`, `@aws-rum/web-slim`. Slim drops mandatory Cognito auth for proxy users. `npm install aws-rum-web` still works as drop-in replacement.
- **UA parsing removed** ([#294](https://github.com/aws-observability/aws-rum-web/issues/294)) — `ua-parser-js` dropped (−18 KB). Browser/OS from `navigator.userAgentData` (Chromium ~70%) or `undefined` (Firefox/Safari). Raw UA string sent as `aws:userAgent` for server-side parsing.
- **Migrated to `@smithy/*`** ([#546](https://github.com/aws-observability/aws-rum-web/issues/546)) — deprecated `@aws-sdk/*` dependencies replaced with `@smithy/*`. Slim doesn't ship them at all.
- **ES2017 target** ([#776](https://github.com/aws-observability/aws-rum-web/pull/776)) — Babel removed. Minimum: Chrome 55, Firefox 52, Safari 10.1, Edge 15. This should not make any meaningful loss of support, however, because we already depend on modules (e.g. sendbecaon) that are not supported in older version anyways. 
- **Session metadata deduplication** — session-level fields (`browserName`, `osName`, `domain`, etc.) sent once per request via `SessionMetadata` instead of repeated per event. **87% payload reduction** on 100-event batch (~44 KB → ~5.8 KB).
- **Slim: no `telemetries`, which now only exist on `aws-rum-web` for backwards compatibility** — use `eventPluginsToLoad` instead. 
- **Web** default telemetries have been changed from none to performance, error, http, and session replay
- **Slim**: Removing `cognitoPoolId` and `guestRoleArn` from configuration, but it can still be dependency injected into slim and still exists on web

## Package Architecture

```
packages/
├── core/    @aws-rum/web-core     (shared engine — internal)
├── web/     aws-rum-web           (full distribution)
└── slim/    @aws-rum/web-slim     (lightweight distribution)
```

```
                    ┌─────────────────────┐
                    │  @aws-rum/web-core  │
                    │  (shared engine)    │
                    └────────┬────────────┘
                             │
              ┌──────────────┴──────────────┐
              │                             │
   ┌──────────▼──────────┐      ┌───────────▼──────────┐
   │  @aws-rum/web-slim  │      │     aws-rum-web      │
   │  (10 KB gz)         │◄─────│     (50 KB gz)       │
   │                     │      │  extends slim +      │
   │  • unsigned dispatch│      │  • auth + signing    │
   │  • PageViewPlugin   │      │  • all plugins       │
   │  • opt-in plugins   │      │  • session replay    │
   └─────────────────────┘      └──────────────────────┘
```

**Full** (`aws-rum-web`, 163 KB / 50 KB gz) includes:
- Slim base + CognitoCredentialFactory + SigV4 signing
- 8 event plugins (RRWeb, Navigation, Xhr, Fetch, WebVitals, DomEvent, Resource, JsError)
- Dropped from 2.x full: `ua-parser-js` (−18 KB)

**Slim** (`@aws-rum/web-slim`, 32 KB / 10 KB gz) excludes via tree-shaking:
- ❌ Auth classes (11.4 KB) · ❌ `@smithy/*` (20.9 KB) · ❌ `ua-parser-js` (18 KB)
- ❌ web-vitals (11.5 KB) · ❌ Event plugins (26.6 KB) · ❌ `@aws-crypto` (1.5 KB)

## Comparisons

**Bundle sizes**
- **The old v2.x** `aws-rum-web` — 154.1 KB raw / 42.5 KB gzip
- **The new full distribution v3.0** `aws-rum-web` (+ session replay) — 163.4 KB raw / 49.8 KB gzip (+17%)
- **The new slim distribution v3.0** `@aws-rum/web-slim` — 32.1 KB raw / 9.9 KB gzip (**−77%**) - _Slim is one of the smallest among competitors at 32 KB raw, and much smaller than the [PutRumEvents Command from AWS SDK (123 KB)](https://github.com/aws/aws-sdk-js-v3/issues/7726)._ 

## Migration
- **Existing CDN users** — no changes. `cwr.js` works as before.
- **Existing NPM users** — no changes for most. Check for `undefined` instead of `"unknown"` in UA fields.
- **New slim users (proxy deployments):**

```ts
import { AwsRum } from '@aws-rum/web-slim';

new AwsRum(appId, version, region, {
    endpoint: 'https://my-proxy.example.com',
    eventPluginsToLoad: [new FetchPlugin(), new JsErrorPlugin()]
});
```

## Testing

1. Unit and integ tests are updated - https://github.com/williazz/aws-rum-web/actions/runs/24157292369
2. Smoke test workflow works, but ingestion fails because `SessionMetadata` has not been deployed to prod yet. We will re-run the smoke test wflow after deployment - https://github.com/williazz/aws-rum-web/actions/runs/24157292363
3. Deployed to personal npm account for validation
    a. https://www.npmjs.com/package/billyzh-aws-rum-web
    b. https://www.npmjs.com/package/@billyzh-aws-rum/web-core
    c. https://www.npmjs.com/package/@billyzh-aws-rum/web-slim
4. Deployed to gamma CDN
    b. https://client.rum-gamma.us-west-2.amazonaws.com/3.0.0/cwr.js
    c. https://client.rum-gamma.us-west-2.amazonaws.com/3.0.0/cwr-slim.js

## Still in progress

1. Merging a bunch of open community prs into core. We can be more lenient about merging stuff in now because just because it's in core, doesn't mean it's actually shipped into the bundle. I'm going to create a community plugins section for stuff that's not bundled, but still in the core library. 
2. Clear all security violations
3. Refactor from deep import to barrel imports

## Misc

This is going to be a fast forward merge to preserve the history. 

---


williazz:
Use bundle-stats.js to see full comparison

```

══════════════════════════════════════════════════════════════════════
  FULL BUNDLE: cwr.js
══════════════════════════════════════════════════════════════════════
  Total:  163.5 KB raw  /  50.1 KB gzip  (EXACT)
──────────────────────────────────────────────────────────────────────
  Category                              Raw % of total
──────────────────────────────────────────────────────────────────────
  rrweb                             67.3 KB      41.2%
  app: plugins                      26.9 KB      16.5%
  aws-sdk/smithy                    19.6 KB      12.0%
  web-vitals                        10.8 KB       6.6%
  app: dispatch                     10.2 KB       6.2%
  app: sessions                      5.7 KB       3.5%
  app: orchestration                 5.1 KB       3.1%
  app: auth                          4.6 KB       2.8%
  app: event-cache                   4.0 KB       2.5%
  app: utils                         2.8 KB       1.7%
  app: other                         1.7 KB       1.1%
  tslib                              1.5 KB       0.9%
  shimmer                            1.4 KB       0.8%
  uuid                               0.8 KB       0.5%
  app: event-bus                     0.4 KB       0.3%
  app: remote-config                 0.4 KB       0.2%
  webpack runtime                    0.1 KB       0.1%
  (unattributed/boilerplate)         0.1 KB       0.1%
──────────────────────────────────────────────────────────────────────


══════════════════════════════════════════════════════════════════════
  SLIM BUNDLE: cwr-slim.js
══════════════════════════════════════════════════════════════════════
  Total:  32.1 KB raw  /  9.9 KB gzip  (EXACT)
──────────────────────────────────────────────────────────────────────
  Category                              Raw % of total
──────────────────────────────────────────────────────────────────────
  app: dispatch                      8.7 KB      27.0%
  app: sessions                      5.7 KB      17.9%
  app: event-cache                   4.0 KB      12.6%
  app: orchestration                 3.5 KB      11.0%
  app: plugins                       2.7 KB       8.3%
  aws-sdk/smithy                     1.9 KB       5.8%
  app: utils                         1.4 KB       4.4%
  shimmer                            1.4 KB       4.3%
  app: other                         1.4 KB       4.3%
  uuid                               0.8 KB       2.6%
  app: event-bus                     0.4 KB       1.3%
  webpack runtime                    0.1 KB       0.4%
──────────────────────────────────────────────────────────────────────
```

.github/scripts/deploy.sh
```diff
@@ -5,8 +5,14 @@ bucket=$1
...
 echo "current-version=$version" >> $GITHUB_OUTPUT
 
-aws s3api put-object --bucket $bucket --key "content/$version/cwr.js" --body build/assets/cwr.js --cache-control max-age=604800 --content-type "text/javascript"
-aws s3api put-object --bucket $bucket --key "content/$version/cwr.js.map" --body build/assets/cwr.js.map --cache-control max-age=604800
+# Upload full distribution (cwr.js)
+aws s3api put-object --bucket $bucket --key "content/$version/cwr.js" --body packages/web/build/assets/cwr.js --cache-control max-age=604800 --content-type "text/javascript"
+aws s3api put-object --bucket $bucket --key "content/$version/cwr.js.map" --body packages/web/build/assets/cwr.js.map --cache-control max-age=604800
+
+# Upload slim distribution (cwr-slim.js)
+aws s3api put-object --bucket $bucket --key "content/$version/cwr-slim.js" --body packages/slim/build/assets/cwr-slim.js --cache-control max-age=604800 --content-type "text/javascript"
```
ps863, line 13: comment: We need to add monitoring on this if we don't have it already. I think right now we only monitor the other bundle

.github/workflows/cd.yaml
```diff
@@ -116,7 +116,10 @@ jobs:
               timeout-minutes: 20
 
             - name: Publish to NPM
-              run: npm publish --provenance
+              run: |
```
ps863: question: How does the flow here exactly work? 
What if we have a scenario wherein the publish to the first 2 succeeds but third one does not.

How would RUM operators handle such scenarios in deployments or roll backs?

packages/web/src/orchestration/config.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const defaultConfig = (cookieAttributes: CookieAttributes): Config => {
+    return {
+        ...slimDefaultConfig(cookieAttributes),
+        signing: true,
+        telemetries: [
+            TelemetryEnum.Errors,
+            TelemetryEnum.Performance,
+            TelemetryEnum.Http,
+            TelemetryEnum.Replay
```
ps863, line 52: question: What is this doing? Are we enabling replay by default or just pulling it into to distribution?

ishajos (approved):

packages/core/package.json
```diff
@@ -0,0 +1,53 @@
...
+        "test": "jest -c ../../jest.unit.config.js",
+        "build:schemas": "node ../../scripts/generate-schema-ts.js",
+        "build:dist:es": "tsc --project ./tsconfig.es.json",
+        "build:dist:cjs": "tsc --project ./tsconfig.cjs.json",
+        "build": "npm-run-all -s build:schemas build:dist:*"
+    },
+    "dependencies": {
+        "@aws-crypto/sha256-js": "^3.0.0",
+        "@aws-sdk/querystring-builder": "^3.36.0",
+        "@babel/runtime": "^7.16.0",
```
ps863: question: do you still need this?

.github/scripts/deploy.sh
williazz: yeah we can add monitoring for this. nice idea

.github/workflows/cd.yaml
williazz: npm publish is idempotent so rum operator can just retry the wflow. but, this is quite annoying. 

i can split this up so that we have separate wflows for each distribution that can be retries separately

i can follow up on this but it should work fine as is + retry

packages/web/src/orchestration/config.ts
williazz: yeah this is enabling session replay by default. i also want to look into increasing the default replay sample rate, since we have reduced the cost of session replay so significantly for rum users with our batching strategy

packages/core/package.json
williazz: no, we don't need babel anymore. good catch. i think this is already tree shaken but we should remove it from dependencies

williazz: removed

williazz: there's also a ton of other dep warnings from the bot that I'll follow up with

.github/workflows/cd.yaml
williazz: split into separate steps as well, so if core fails, then we don't attempt to publish to slim or web

williazz:
Also resolves the following open prs by deleting package-lock in examples directory

Resolves #776
Resolves #795
Resolves #798
Resolves #797
Resolves #796
Resolves #794
Resolves #793
Resolves #792
Resolves #784
Resolves #783
Resolves #780
Resolves #769
Resolves #768
Resolves #767
Resolves #763

ps863 (approved):

williazz:
Still has one remaining issue #546
Closes #294
Closes #507

PR #818, opened 2026-04-29
fix: replace uuid dependency with native crypto.randomUUID

piotrekwitkowski:
## Summary
- Removes the `uuid` package to eliminate [GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq) (moderate severity, uuid <14.0.0)
- Replaces `v4()` calls with a `generateUUID()` utility in `packages/core/src/utils/random.ts`
- Uses `crypto.randomUUID()` where available, with a fallback to `crypto.getRandomValues()` for older browsers (IE11+ via msCrypto)

## Motivation
Upgrading uuid to 14.0.0 (#815) breaks the Jest test suite because uuid@14 is ESM-only. Rather than fighting ESM compatibility, this removes the dependency entirely.

## Test plan
- All 45 test suites pass (668 tests)
- Webpack production build compiles successfully

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3258,21 +3258,6 @@
             "dev": true,
             "license": "MIT"
         },
-        "node_modules/@noble/hashes": {
-            "version": "1.8.0",
-            "resolved": "https://registry.npmjs.org/@noble/hashes/-/hashes-1.8.0.tgz",
-            "integrity": "sha512-jCs9ldd7NwzpgXDIf6P3+NrHh9/sD6CQdxHyjQI+h/6rDNo88ypBxxz45UDuZHz9r3tNz7N/VInSVoVdtXEI4A==",
-            "dev": true,
-            "license": "MIT",
-            "optional": true,
-            "peer": true,
-            "engines": {
-                "node": "^14.21.3 || >=16"
-            },
-            "funding": {
-                "url": "https://paulmillr.com/funding/"
-            }
-        },
         "node_modules/@nodelib/fs.scandir": {
             "version": "2.1.5",
             "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.5.tgz",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
         "@types/pako": "^2.0.4",
         "@types/shimmer": "^1.0.2",
         "@types/ua-parser-js": "^0.7.35",
-        "@types/uuid": "^9.0.0",
         "@typescript-eslint/eslint-plugin": "^5.38.0",
         "@typescript-eslint/parser": "^5.38.0",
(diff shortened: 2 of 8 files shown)


piotrekwitkowski:
For anyone wondering about the bundle impact: with Vite/Rollup tree-shaking, the uuid v4 path (v4.js + rng.js + stringify.js + native.js) contributes ~3 KB raw / ~1.3 KB gzip to the final bundle. Small but free savings — especially welcome alongside the rrweb tree-shaking issue (#823) which is a much larger problem for npm consumers.

williazz (approved):

williazz:
this seems to have broken the cdn build - https://github.com/aws-observability/aws-rum-web/actions/runs/25335096721/job/74278671197

developers who are concerned about package size should migrate to slim

piotrekwitkowski:
@williazz Thanks! I don't think this PR caused that failure though

The build steps pass, and the errors are all Token expired from the ingestion smoke tests (expiration is ~3.5 min before "current time" on every retry). Looks like a Cognito/STS or clock-skew issue in the smoke-test env, not something crypto.randomUUID would touch. Mind re-running?

williazz:
Could you provide some proof? The expired token is a flaky setup issue that can be improved and maybe causes some false positives, but we're also getting that exception because we can't find the events in time. If you look at the earlier failed fetches, the tokens are not expired.

piotrekwitkowski:
Locally I verified the prod-minified `cwr.js` produces valid v4 UUIDs for BatchId, event.id, sessionId, and userId, and terser preserves the `crypto.randomUUID` check.

--

On the other hand the log literally only tells us "at least one expected id wasn't in the response." It doesn't tell us whether the response had zero events, or many events that didn't include this one. We can't distinguish those from the log we have.

What we can still say:
- `getIngestedEvents` calls `rumClient.send(command)` directly. If that throws, the catch prints the AWS SDK error message instead of "Event ... not ingested." The log doesn't show that, so `GetAppMonitorData` itself returned 200.
- `GetAppMonitorData` returned 200 but its `Events` array either didn't contain the expected id, or was empty. We can't tell which.

So the fix to actually get more useful info would be to log ingestedEvents.size on failure, or even the first few ids that were returned. Then the next run distinguishes "backend returned nothing" from "backend returned events but not right ones". Do you want me to implement that and then we can rerun the pipeline?

williazz:
Please point it against the actual prod put rum events endpoint and try to fetch them. That endpoint is very opinionated and can silently invalidate events. I can try to deflake the credentials in the smoke test on my backlog.

piotrekwitkowski:
@williazz I have build what you asked for today, please see: https://github.com/piotrekwitkowski/uuid-818-repro

In my opinion the smoke test flakiness is unrelated to the crypto.randomUUID() change. Can we force merge this and include in the next release?

PR #830, opened 2026-05-15
feat: manage cross-context users and sessions 

### LINKED ISSUE #354: feat: Support setting userId and sessionId manually
I am working on adding CloudWatch RUM to a web app. Our users will authenticate with AWS Identity Center (IDC) which assigns each user a UUID already. I'd like to use that same UUID as the `userId` RUM instead of letting RUM generate a new UUID for me. Having consistent userIds will make it easier to connect RUM data with our backend events. We could pass in the IDC user ID as event metadata but it makes sifting through data more confusing and error prone.

Passing in user and session IDs could look like:
```
  userId: 'bd33b2a2-4b14-4ea6-aaeb-ee85c2062f5d',
  sessionId: '40af9b9a-6fa2-45f8-a437-67d2566380d9'
};
```

williazz:
## Summary

In order to support custom platforms and architectures, you need two things:

1. Very fine-level layers to modify metadata, especially to bypass silent and opinonated input validations during PutRumEvents (already covered by https://github.com/aws-observability/aws-rum-web/pull/829). This lets you safely spoof things like `domain` or `platform` and force feed telemetries into a web app monitor. 
2. Ability to manage single user sessions across multiple contexts, such as microfrontend architectures stitched together by webviews. 

(1) and (2) can also be used together to gain observability into unsupported platforms, such as electron or vscode extensions. 

### Improved Management APIs for Users and Sessions

Setting the basic logic for supporting web views in other embedded context where one user session spans multiple DOMs. We do this by allowing session to be manually managed while also suppressing session start events using leader and follower framework. Currently, we remain unopinionated about how followers communicate with leaders due to different event bridge contexts (webview, vscode, electron, etc.) In the future, we may add dedicated tooling. Similarly, we want to allow setting user id programatically. 

Closes #354

### Changes

Creating new public SDK commands that are only exposed to NPM in this release. Exposing to CDN commands would also need a bit of a re-architecture because CDN commands are all fire and forget, so only mutating commands were added to CDN. We can follow up with read commands in separate PR. 

| Command | NPM | CDN | Description |
| --- | --- | --- | --- |
| **getSessionId** | `awsRum.getSessionId()` | _(NPM only)_ | Returns the current session ID, minting a new one if no session exists. Use to read the leader's session ID for broadcast to follower contexts. See [Sharing a session across contexts](../concepts/sessions.md#sharing-a-session-across-contexts). |
| **pinSessionId** | `awsRum.pinSessionId('shared-uuid')` | `cwr('pinSessionId', 'shared-uuid')` | Adopt an externally-minted session ID. Subsequent dispatches use this ID; already-buffered events inherit it at dispatch time (sessionId rides on `UserDetails` per batch, not per event). Does **not** emit `session_start` and does **not** re-roll sampling. Empty values are ignored with a warning. |
| **startSession** | `awsRum.startSession()` <br> `awsRum.startSession({ sessionId, userId })` | `cwr('startSession')` <br> `cwr('startSession', { sessionId, userId })` | Begin a new session immediately at a host-known boundary (sign-in, sign-out, kiosk handoff). With no args: mints a fresh UUID, re-rolls sampling, disengages session manual mode, emits `session_start` (subject to `suppressSessionStartEvent`). Optional `sessionId` adopts a host-chosen ID and engages manual mode; optional `userId` rotates the user identity in the same call (silent — no `user_start` event exists). NPM returns the new session ID; the CDN form is fire-and-forget — pass `sessionId` if you need to know it. See [Starting a session at runtime](../concepts/sessions.md#starting-a-session-at-runtime). |
| **getUserId** | `awsRum.getUserId()` | _(NPM only)_ | Returns the current anonymous user ID. Returns the nil UUID (`00000000-0000-0000-0000-000000000000`) when cookies are disabled and no `userId` has been seeded or set. Use to read the leader's user ID for broadcast to follower contexts so a single human is not counted as N anonymous users. See [Sharing a user ID across contexts](../concepts/sessions.md#sharing-a-user-id-across-contexts). |
| **pinUserId** | `awsRum.pinUserId('shared-uuid')` | `cwr('pinUserId', 'shared-uuid')` | Adopt an externally-supplied user ID. Engages manual mode for the life of the instance, overriding `userIdRetentionDays: 0` and the `allowCookies: false` gate in `getUserId()`. No event is emitted. Empty values are ignored with a warning. |
---

## Limitation

Session sampling decision is not propagated from leader to followers. We would like to do that as a follow up. Out of scope for this PR. 


limhjgrace:
question: Does the leader need to be in manual mode for idle-leader architectures?

The documented leader example constructs in auto mode — so the SDK will auto-rotate the session if the leader is inactive for sessionLengthSeconds. I'm assuming that VS Code extensions or microfrontend shells might be scenarios where the leader is idle while followers generate telemetry and was wondering if this would silently fragment the session (leader rotates, followers stay on the old ID)?

If so, should the leader also be in manual mode for those architectures? Or is the assumption that the leader is always active enough to keep its expiry renewed?

limhjgrace (approved):

williazz:
> where the leader is idle while followers generate telemetry and was wondering if this would silently fragment the session (leader rotates, followers stay on the old ID)?

If the leader is gone, then the followers will not exist anymore. For example, the embedded webview inside your website or electron app is getting torn down if the root context is gone.

williazz:
We can improve this as we go but it definitely unblocks most of what we want to do

PR #847, opened 2026-06-11
ci: switch to default CodeQL configuration

Francydijesi:
Remove custom CodeQL workflow in favor of repository-level default setup, which provides broader coverage including GitHub Actions analysis.


williazz:
if we are removing cache as a dependency, then please restore the bundle size analyzer script which we just broke.

PR #865, opened 2026-07-20
fix(smoke): pin @types/node to ^20 for TS 4.9 compatibility

boyanjt:
The smoke-test build broke because @types/node floated to v26, which uses TypeScript 5.0 `const` type-parameter syntax that the smoke apps' TypeScript 4.9 cannot parse (parse errors in ffi.d.ts). Pin @types/node to ^20 via devDependency + npm overrides in both the NPM-ES and NPM-CJS smoke apps and regenerate their lockfiles. Unblocks the smoke tests added in this branch.

---


ps863:
bundle analyzer has permission issue which works otherwise - https://github.com/aws-observability/aws-rum-web/pull/864

ps863 (approved):

PR #888, opened 2026-08-19
fix: replace uuid dependency with native crypto.randomUUID

### LINKED ISSUE #824: Update uuid to `^14.0.0` to address GHSA-w5hq-g745-h8pq
### Which web client version did you detect this bug with?

- v2.1.0
- v3.0.0

### What environment (build systems, module system, and framework) did you detect this bug with?

web:
 - npm + vite
 - pnpm + vite

### Is your web application a single page application (SPA) or multi page application (MPA)?

SPA

### Please provide your web client configuration

_No response_

### Please describe the bug/issue

`npm audit` highlights that this package still uses a vulnerable version of [uuid](https://www.npmjs.com/package/uuid) which is

> Missing buffer bounds check in v3/v5/v6 when buf is provided 

https://github.com/advisories/GHSA-w5hq-g745-h8pq

```json
{
  "advisories": {
    "1116970": {
      "findings": [
        {
          "version": "9.0.1",
          "paths": [
            ".>aws-rum-web>@aws-rum/web-core>uuid",
            ".>aws-rum-web>@aws-rum/web-slim>@aws-rum/web-core>uuid"
          ],
          "dev": false,
          "optional": false,
          "bundled": false
        }
      ],
      "id": 1116970,
      "title": "uuid: Missing buffer bounds check in v3/v5/v6 when buf is provided",
      "module_name": "uuid",
      "vulnerable_versions": "<14.0.0",
      "patched_versions": ">=14.0.0",
      "severity": "moderate",
      "cwe": "CWE-787, CWE-1285",
      "github_advisory_id": "GHSA-w5hq-g745-h8pq",
      "url": "https://github.com/advisories/GHSA-w5hq-g745-h8pq"
    }
  },
}
```

---

### Related to:
- #818
- b154d516477e47739bfa3c84b3e731969ab83788

piotrekwitkowski:
Re-lands #818, which was merged as 493469e and then reverted in b154d51. Fixes #824.

`@aws-rum/web-core` depends on `uuid` solely for `v4()`. That pulls `uuid@9.0.1` into every consumer's production tree, where `npm audit` flags [GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq) via `.>aws-rum-web>@aws-rum/web-core>uuid`. The advisory is a missing buffer bounds check in `v3`/`v5`/`v6` when `buf` is passed, a code path this library never calls. The practical problem is therefore a permanently dirty audit for every downstream consumer, not an exploitable defect.

This replaces the dependency with `crypto.randomUUID()`, falling back to a v4 built from `crypto.getRandomValues()` where `randomUUID` is unavailable (it requires a secure context). No runtime dependency, no advisory, smaller bundle.

Per @TrevorBurnham's [comment](https://github.com/aws-observability/aws-rum-web/issues/824#issuecomment-5330779141) on #824, the alternative is bumping to `uuid@11.1.1`. The advisory has been rescoped from `<14.0.0` to `<11.1.1`, so the ESM-only 14.x problem blocking #879 no longer applies. Removing the dependency seemed preferable to keeping one for a single `v4()` call, but the bump is a reasonable call if you'd rather not own the UUID generation.

## Addressing the re-land review conditions

**Unit tests for `generateUUID`.** `random.ts` was at 28% line coverage; it is now at 100% across lines, statements, functions, and branches, via 12 new tests. Three deterministic byte vectors pin the bit manipulation rather than just regex-matching the output:

| bytes | expected |
|---|---|
| `0x00`×16 | `00000000-0000-4000-8000-000000000000` |
| `0xff`×16 | `ffffffff-ffff-4fff-bfff-ffffffffffff` |
| `0xab`×16 | `abababab-abab-4bab-abab-abababababab` |

The all-zero vector is the important one: it's the regression test for `padStart(2, '0')`, where a hand-rolled hex encoder would otherwise drop leading zeros and silently emit a short, malformed UUID. The suite also covers the native path (asserting the byte fallback is not touched), the `msCrypto` fallback, the throw when no crypto exists, and 1000-iteration format/uniqueness loops over both the native and fallback paths.

**Hardening `getRandomValues`.** The bare truthiness checks are now `typeof` guards. This mattered more than it looks: `msCrypto` is only `declare`d and never bound, so `else if (msCrypto)` threw `ReferenceError` on precisely the legacy path it exists to serve, which made the `throw new Error('No crypto library found.')` branch unreachable in practice.

**CDN smoke run.** This one needs a maintainer, since `smoke.yml` has no `pull_request` trigger (only `push` to `main`/`release/*`, `workflow_call`, `workflow_dispatch`) and it publishes to the Gamma CDN using repository secrets. Happy to iterate if a `workflow_dispatch` run surfaces anything.

## Notes on the re-land

**A plain revert of b154d51 will not compile.** #830 added a ninth `uuid.v4()` call site to `SessionManager.test.ts` after #818 was reverted, so #818's import removal no longer covers the whole file. Combining the two yields `TS2304: Cannot find name 'uuid'`, which git won't surface as a conflict because the edits are in different regions. This PR converts that call site along with the other eight, so all nine now use `crypto.randomUUID()`.

**The original #818 left the lockfile unpruned.** It removed `uuid` from `package.json` and `packages/core/package.json` but never pruned the lock, which still carried `uuid: ^9.0.0` under `packages/core`, a root `@types/uuid`, and `node_modules/uuid` 9.0.1 entries. In-repo `npm audit` therefore kept reporting a `node_modules/uuid` node that `npm ci` doesn't even install. Removed here as 22 pure deletions.

Rebased onto v3.2.1, so the lock now inherits the `3.2.0` `version` fields that #889 wrote. The remaining `3.2.0` against `3.2.1` gap is pre-existing drift, because the v3.2.1 release commit didn't touch the lock, and syncing it here would only conflict with your next release commit.

## Verification

On a clean `npm ci`:

- Unit suite: **765 passed, 46/46 suites**, jest exit 0
- Coverage thresholds pass, with branches 87.05% (≥86), functions 88.48% (≥88), lines 93.2%, statements 93.21%
- `npm run build`: exit 0
- `npm audit --omit=dev`: no `uuid` at all. The remaining `uuid@8.3.2` nodes are nested under `nightwatch` and `sockjs`, both devDependencies
- Published `@aws-rum/web-core` declares 10 dependencies, none of them `uuid`; no built bundle references a uuid module
- eslint and prettier clean

Not verified locally: the Playwright integ matrix, since no browsers are cached in my environment. Leaving that to CI, which covers all four engines.

### DIFF
--- a/packages/core/src/sessions/SessionManager.ts
+++ b/packages/core/src/sessions/SessionManager.ts
@@ -347,7 +347,7 @@ export class SessionManager {
         } else if (this.useCookies()) {
             userId = this.getUserIdCookie();
             this.userId = userId ? userId : generateUUID();
-            this.createOrRenewUserCookie(userId, this.userExpiry);
+            this.createOrRenewUserCookie(this.userId, this.userExpiry);
         } else {
             this.userId = generateUUID();
         }
--- a/packages/core/src/sessions/SessionManager.ts
+++ b/packages/core/src/sessions/SessionManager.ts
@@ -1,6 +1,5 @@
 import { storeCookie, getCookie } from '../utils/cookies-utils';
 
-import { v4 } from 'uuid';
 import { Config, userAgentDataProvider } from '../orchestration/config';
 import { Page, PageManager } from './PageManager';
 import { SESSION_COOKIE_NAME, USER_COOKIE_NAME } from '../utils/constants';
@@ -8,6 +7,7 @@ import { AppMonitorDetails } from '../dispatch/dataplane';
 import { RecordEvent } from '../plugins/types';
 import { SESSION_START_EVENT_TYPE } from '../plugins/utils/constant';
 import { InternalLogger } from '../utils/InternalLogger';
+import { generateUUID } from '../utils/random';
 
 export const NIL_UUID = '00000000-0000-0000-0000-000000000000';
 
@@ -346,10 +346,10 @@ export class SessionManager {
             this.userId = '00000000-0000-0000-0000-000000000000';
         } else if (this.useCookies()) {
             userId = this.getUserIdCookie();
-            this.userId = userId ? userId : v4();
+            this.userId = userId ? userId : generateUUID();
             this.createOrRenewUserCookie(userId, this.userExpiry);
         } else {
-            this.userId = v4();
+            this.userId = generateUUID();
         }
     }
 
@@ -414,7 +414,7 @@ export class SessionManager {
         // We ensure the nil session and new session created right after initialization have the same sampling decision.
         // Otherwise, we will always reevaluate the sample decision.
         this.session = {
-            sessionId: v4(),
+            sessionId: generateUUID(),
             record:
                 this.session.sessionId === NIL_UUID
                     ? this.session.record
--- a/packages/core/src/utils/random.ts
+++ b/packages/core/src/utils/random.ts
@@ -11,3 +11,21 @@ export const getRandomValues = (holder: Uint8Array): Uint8Array => {
         throw new Error('No crypto library found.');
     }
 };
+
+export const generateUUID = (): string => {
+    if (typeof crypto !== 'undefined' && crypto.randomUUID) {
+        return crypto.randomUUID();
+    }
+    const bytes = getRandomValues(new Uint8Array(16));
+    // eslint-disable-next-line no-bitwise
+    bytes[6] = (bytes[6] & 0x0f) | 0x40;
+    // eslint-disable-next-line no-bitwise
+    bytes[8] = (bytes[8] & 0x3f) | 0x80;
+    const hex = Array.from(bytes, (b) => b.toString(16).padStart(2, '0')).join(
+        ''
+    );
+    return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(
+        12,
+        16
+    )}-${hex.slice(16, 20)}-${hex.slice(20)}`;
+};
--- a/packages/core/src/utils/random.ts
+++ b/packages/core/src/utils/random.ts
@@ -24,7 +24,10 @@ export const getRandomValues = (holder: Uint8Array): void => {
 };
 
 export const generateUUID = (): string => {
-    if (typeof crypto !== 'undefined' && crypto.randomUUID) {
+    if (
+        typeof crypto !== 'undefined' &&
+        typeof crypto.randomUUID === 'function'
+    ) {
         return crypto.randomUUID();
     }
     const bytes = new Uint8Array(16);
--- a/packages/core/src/utils/random.ts
+++ b/packages/core/src/utils/random.ts
@@ -1,12 +1,23 @@
 declare let msCrypto:
     | undefined
-    | { getRandomValues: (holder: Uint8Array) => Uint8Array };
+    | { getRandomValues: (holder: Uint8Array) => void };
 
-export const getRandomValues = (holder: Uint8Array): Uint8Array => {
-    if (typeof crypto !== 'undefined') {
-        return crypto.getRandomValues(holder);
-    } else if (typeof msCrypto !== 'undefined') {
-        return msCrypto.getRandomValues(holder);
+// The holder is filled in place. Callers must read the array they passed in,
+// because legacy implementations such as IE11's msCrypto return undefined.
+export const getRandomValues = (holder: Uint8Array): void => {
+    // `typeof x !== 'undefined'` must come first. Optional chaining still
+    // evaluates the identifier, so `crypto?.foo` throws a ReferenceError when
+    // no `crypto` binding exists at all.
+    if (
+        typeof crypto !== 'undefined' &&
+        typeof crypto.getRandomValues === 'function'
+    ) {
+        crypto.getRandomValues(holder);
+    } else if (
+        typeof msCrypto !== 'undefined' &&
+        typeof msCrypto.getRandomValues === 'function'
+    ) {
+        msCrypto.getRandomValues(holder);
     } else {
         throw new Error('No crypto library found.');
     }
@@ -16,7 +27,8 @@ export const generateUUID = (): string => {
     if (typeof crypto !== 'undefined' && crypto.randomUUID) {
         return crypto.randomUUID();
     }
-    const bytes = getRandomValues(new Uint8Array(16));
+    const bytes = new Uint8Array(16);
+    getRandomValues(bytes);
     // eslint-disable-next-line no-bitwise
     bytes[6] = (bytes[6] & 0x0f) | 0x40;
     // eslint-disable-next-line no-bitwise
--- a/packages/core/src/utils/random.ts
+++ b/packages/core/src/utils/random.ts
@@ -3,9 +3,9 @@ declare let msCrypto:
     | { getRandomValues: (holder: Uint8Array) => Uint8Array };
 
 export const getRandomValues = (holder: Uint8Array): Uint8Array => {
-    if (crypto) {
+    if (typeof crypto !== 'undefined') {
         return crypto.getRandomValues(holder);
-    } else if (msCrypto) {
+    } else if (typeof msCrypto !== 'undefined') {
         return msCrypto.getRandomValues(holder);
     } else {
         throw new Error('No crypto library found.');
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,6 @@
                 "@types/pako": "^2.0.4",
                 "@types/shimmer": "^1.0.2",
                 "@types/ua-parser-js": "^0.7.35",
-                "@types/uuid": "^9.0.0",
                 "@typescript-eslint/eslint-plugin": "^5.38.0",
                 "@typescript-eslint/parser": "^5.38.0",
                 "@webpack-cli/serve": "^3.0.1",
@@ -5865,13 +5864,6 @@
             "version": "0.7.39",
             "resolved": "https://registry.npmjs.org/@types/ua-parser-js/-/ua-parser-js-0.7.39.tgz",
             "integrity": "sha512-P/oDfpofrdtF5xw433SPALpdSchtJmY7nsJItf8h3KXqOslkbySh8zq4dSWXH2oTjRvJ5PczVEoCZPow6GicLg==",
-            "dev": true,
-            "license": "MIT"
-        },
-        "node_modules/@types/uuid": {
-            "version": "9.0.8",
-            "resolved": "https://registry.npmjs.org/@types/uuid/-/uuid-9.0.8.tgz",
-            "integrity": "sha512-jg+97EGIcY9AGHJJRaaPVgetKDsrTgbRjQ5Msgjh/DQKEFl0DtyRr/VCOyD1T2R1MNeWPK/u7JoGhlDZnKBAfA==",
             "dev": true,
             "license": "MIT"
         },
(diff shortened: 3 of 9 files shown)


packages/core/src/utils/random.ts
```diff
@@ -3,11 +3,29 @@ declare let msCrypto:
...
     } else {
         throw new Error('No crypto library found.');
     }
 };
+
+export const generateUUID = (): string => {
+    if (typeof crypto !== 'undefined' && crypto.randomUUID) {
+        return crypto.randomUUID();
+    }
+    const bytes = getRandomValues(new Uint8Array(16));
```
TrevorBurnham: This is the only caller that uses the return value of `getRandomValues` rather than the buffer it passed in. The three pre-existing call sites in `plugins/utils/http-utils.ts` (294, 307, 317) all discard the return and read their own array, which is also what `uuid`'s `rng-browser` did.

That matters because an implementation that fills in place and returns `undefined` is a real shape: it's what IE11's `msCrypto.getRandomValues` does, and it's the common form of the hand-rolled `crypto` stub consumers inject into jsdom/Node test environments. In that case `bytes` is `undefined` and line 21 throws `TypeError: Cannot set properties of undefined`.

The new test can't catch it, because `fillWith` is `(holder) => holder.fill(byte)` and `Uint8Array.prototype.fill` returns the array, so the stub is more generous than the declared contract. Suggest:

```ts
const bytes = new Uint8Array(16);
getRandomValues(bytes);
```

---
*Generated by an automated code review (Claude Code), posted on behalf of @TrevorBurnham.*

packages/core/src/utils/random.ts
```diff
@@ -3,11 +3,29 @@ declare let msCrypto:
     | { getRandomValues: (holder: Uint8Array) => Uint8Array };
 
 export const getRandomValues = (holder: Uint8Array): Uint8Array => {
-    if (crypto) {
+    if (typeof crypto !== 'undefined') {
```
TrevorBurnham: The guard checks that `crypto` exists but then calls `crypto.getRandomValues`, so on an engine where the object is present and the method is not, this branch is taken and throws `TypeError: crypto.getRandomValues is not a function` instead of falling through. That legacy IE/Edge shape is exactly what the `msCrypto` branch below exists to serve, so the branch stays unreachable in its own motivating case (and any partial `crypto` stub hits the same wall).

`generateUUID` on line 16 already gets this right by method-checking `crypto.randomUUID`, so the two guards in this file currently disagree. `typeof crypto?.getRandomValues === 'function'` makes them consistent. Worth adding a test for "crypto present, method missing" too: the existing case only removes `crypto` entirely, which is why 100% branch coverage doesn't surface this.

---
*Generated by an automated code review (Claude Code), posted on behalf of @TrevorBurnham.*

packages/core/src/utils/random.ts
```diff
@@ -3,11 +3,29 @@ declare let msCrypto:
...
+export const generateUUID = (): string => {
+    if (typeof crypto !== 'undefined' && crypto.randomUUID) {
+        return crypto.randomUUID();
+    }
+    const bytes = getRandomValues(new Uint8Array(16));
+    // eslint-disable-next-line no-bitwise
+    bytes[6] = (bytes[6] & 0x0f) | 0x40;
+    // eslint-disable-next-line no-bitwise
+    bytes[8] = (bytes[8] & 0x3f) | 0x80;
+    const hex = Array.from(bytes, (b) => b.toString(16).padStart(2, '0')).join(
```
TrevorBurnham, line 39: `padStart` is ES2017 and I think this is the first hard ES2017 *library* dependency in `src/`. `package.json`'s `browserslist` declares `Chrome >= 55`, but `String.prototype.padStart` shipped in Chrome 57, and there's no Babel or core-js in the build (ts-loader only) to polyfill it. The other three entries are fine; Chrome is the only gap.

On Chrome 55-56 `crypto.randomUUID` is absent, so `generateUUID` always reaches this line and throws, which takes out session ID, user ID, and `BatchId` together. `uuid@9` had no such requirement (byte-to-hex lookup table).

To be fair, `docs/migration_v2_to_v3.md` states an ES2017 target, which legitimately includes `padStart`, so the real defect may be a stale `browserslist` field rather than this line. Either way the two disagree today. `(b + 0x100).toString(16).slice(1)` sidesteps the question for free and keeps the all-zero vector green.

---
*Generated by an automated code review (Claude Code), posted on behalf of @TrevorBurnham.*

packages/core/src/sessions/SessionManager.ts
```diff
@@ -346,10 +346,10 @@ export class SessionManager {
             this.userId = '00000000-0000-0000-0000-000000000000';
         } else if (this.useCookies()) {
             userId = this.getUserIdCookie();
-            this.userId = userId ? userId : v4();
+            this.userId = userId ? userId : generateUUID();
             this.createOrRenewUserCookie(userId, this.userExpiry);
```
TrevorBurnham: Not introduced here, but the line above is in the diff and the fix is one token: this passes the local `userId`, not `this.userId`. With no `cwr_u` cookie, `userId` is `''`, so the freshly generated ID on line 349 is discarded and an empty cookie is written.

It's usually repaired by the first `getSession()` → `createSession()` → `storeSessionAsCookie()`, but that repair is skipped when a valid unexpired session cookie was restored and `getSession()` returns early. In that case `cwr_u` stays empty and a new user ID is minted on each page load, which quietly breaks return-visitor counting.

---
*Generated by an automated code review (Claude Code), posted on behalf of @TrevorBurnham.*

TrevorBurnham (commented):
The dependency removal looks right and the bit manipulation is correct: I checked the version/variant masks and the 8-4-4-4-12 slice offsets by hand, and verified `npm ci`, `npm test`, `npm run build`, and `npm run lint:errors` all pass on this branch, with no `uuid` call sites left anywhere in `src/`.

One thing worth stating explicitly, since it's what the comments below hang on: the fallback path is not vestigial. `crypto.randomUUID` requires a secure context in addition to Chrome 92 / Firefox 95 / Safari 15.4, so any customer serving over plain `http://` takes the hand-rolled path for every UUID it generates. That makes it worth being as defensive as the code it replaces.

Only the first comment is a blocker for me. The other three are cheap.

---
*Generated by an automated code review (Claude Code), posted on behalf of @TrevorBurnham.*

packages/core/src/utils/random.ts
TrevorBurnham: I'd add that I don't think there's any good reason to support Chrome 55-56 at this point (Chrome 56 was released in 2017), so updating `browserslist` strikes me as the most sensible solution.

packages/core/src/utils/random.ts
piotrekwitkowski: Agreed, and going a step further, because nothing reads the return value once this is fixed. The three callers in `plugins/utils/http-utils.ts` (294, 307, 317) already call it as a bare statement, and `FetchPlugin.test.ts:29` and `XhrPlugin.test.ts:25` mock it as a no-op. So the return type can become `void`, which makes the dependency impossible to reintroduce.

```ts
const bytes = new Uint8Array(16);
getRandomValues(bytes);
bytes[6] = (bytes[6] & 0x0f) | 0x40; // reads our array, not the return value
```

`getRandomValues` never leaves the package, since `packages/core/src/index.ts` does not export `utils/random`, so this changes no public signature.

You are right about why the test misses it. My stub at `random.test.ts:22` does `holder.fill(byte)`, and `fill` returns the array, so it hands something back when nothing requires it to. Once the return is `void` the stub can stay as it is, because nothing can read what it gives back. Four assertions check the return value today and move to checking `holder`, which is what production reads.

packages/core/src/utils/random.ts
piotrekwitkowski: Correct. Switching to `typeof crypto?.getRandomValues === 'function'` and adding the "crypto present, method missing" test. This lands in the same edit as your line 19 comment.

```ts
declare let msCrypto:
    | undefined
    | { getRandomValues: (holder: Uint8Array) => void };

export const getRandomValues = (holder: Uint8Array): void => {
    if (typeof crypto?.getRandomValues === 'function') {
        crypto.getRandomValues(holder);
    } else if (typeof msCrypto !== 'undefined') {
        msCrypto.getRandomValues(holder);
    } else {
        throw new Error('No crypto library found.');
    }
};
```

Two things fall out of it. The `msCrypto` branch becomes reachable for the first time, because line 6 used to swallow the legacy shape before line 8 could ever see it. And the declared `=> Uint8Array` on `msCrypto` was wrong, since IE11 writes in place and returns `undefined`, so it becomes `=> void`.

packages/core/src/utils/random.ts
piotrekwitkowski: I agree that updating `browserslist` is the right fix, and I would like to do it separately, because `padStart` is not what breaks it.

`VirtualPageLoadTimer.ts:190` calls `.finally()`, which ES2018 puts at Chrome 63, Firefox 58, Safari 11.1 and Edge 18. That misses all four `browserslist` entries rather than Chrome alone, and it predates this PR. Since the real floor already sits at 63, well above `padStart` at 57, my line cannot lower anything. Chrome 55 and 56 do break as you describe, but they break today without it.

One correction on "first hard ES2017 library dependency". `Object.entries` also dates from ES2017 and already appears at `EventCache.ts:43` and `StsClient.ts:37`. It causes no trouble, having shipped in Chrome 54, whereas `.finally` exceeds every declared target.

No tool reads the field, which explains how this survived. `browserslist` appears only at `package.json:108`, the tree carries no `babel`, `core-js`, `autoprefixer` or `postcss`, and ts-loader handles the build alone as you noted. Because `tsconfig.json` pairs `target: es2017` with `lib: ["dom", "es2018"]`, TypeScript accepts ES2018 library calls and never polyfills them. A green build and full branch coverage both walk straight past `.finally`.

Deciding the policy belongs to you rather than me. Dropping Chrome below 63 officially may want a release note, and the field itself could either track `tsconfig` or go away entirely, given nothing reads it. Filed as #890 with the detail above. Happy to bump `browserslist` in this PR instead if you would rather keep it in one change, though I would not want the change to imply `padStart` opened a gap `.finally` had already opened.

packages/core/src/sessions/SessionManager.ts
piotrekwitkowski: Good catch. The early return is what makes it bite. A live unexpired session cookie means `createSession()` never runs, so nothing repairs the empty `cwr_u` and a fresh user ID gets minted on every page load. Return visitor counts drift and no error surfaces.

Fixing it here, since the line already sits in this diff. Also filed as #891, deliberately. #818 was merged and reverted once already, and I would rather this bug survive as a record if the same thing happens to this re-land. Happy to move it to its own PR if you would prefer this one stay scoped to the dependency removal.

packages/core/src/utils/random.ts
piotrekwitkowski: Correction to my reply above. `typeof crypto?.getRandomValues === 'function'` does not work here, and the existing tests caught it.

Optional chaining guards a null or undefined *value*, not an undeclared *binding*. `crypto?.getRandomValues` still evaluates the `crypto` identifier, so once `removeCrypto()` deletes it from `globalThis` the expression throws `ReferenceError: crypto is not defined` rather than falling through. Four tests failed on it, including the two that assert `No crypto library found.`, so the throw branch you wanted reachable became unreachable a second way.

The `typeof x !== 'undefined'` test has to stay in front, with the method check after it.

```ts
if (
    typeof crypto !== 'undefined' &&
    typeof crypto.getRandomValues === 'function'
) {
    crypto.getRandomValues(holder);
} else if (
    typeof msCrypto !== 'undefined' &&
    typeof msCrypto.getRandomValues === 'function'
) {
    msCrypto.getRandomValues(holder);
} else {
    throw new Error('No crypto library found.');
}
```

That reaches your intent, since an engine carrying `crypto` without `getRandomValues` now falls through to `msCrypto`, and it keeps working when no binding exists at all. I gave `msCrypto` the same treatment for symmetry, so a partial `msCrypto` reaches the throw rather than a `TypeError`.

Pushed as f1d5843. `random.ts` holds 100% on lines, statements, functions and branches, and the suite is at 769 across 46 suites.

piotrekwitkowski: Following the same thread, `generateUUID` had the mirror image of this problem, so I fixed it in e29ae13.

Line 27 guarded on the truthiness of `crypto.randomUUID` rather than its type, so after the `getRandomValues` change the two guards in the file disagreed again, just in the opposite direction. A `crypto` object carrying a non-callable `randomUUID` passed the truthiness test and then threw `TypeError: crypto.randomUUID is not a function`, rather than falling through to the byte fallback that exists for exactly that case.

```ts
if (
    typeof crypto !== 'undefined' &&
    typeof crypto.randomUUID === 'function'
) {
    return crypto.randomUUID();
}
```

The new test pins it. Against the old guard it fails with that TypeError, and it passes once the type is checked. Both guards now use the same shape, so 770 tests pass across 46 suites and `random.ts` stays at 100% on lines, statements, functions and branches.

